PR #2, opened 2020-11-23
Fix weird 'C' in CompletionResponse struct

bnwlkr:
🤔

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -49,7 +49,7 @@ type CompletionResponse struct {
 	Object  string   `json:"object"`
 	Created uint64   `json:"created"`
 	Model   string   `json:"model"`
-	Сhoices []Choice `json:"choices"`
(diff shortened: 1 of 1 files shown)


bnwlkr:
My friend has informed me that the 'C' is not weird but Cyrillic 😅

sashabaranov:
@bnwlkr Wow that's a good one! Thank you for the PR 😌

The latin C and cyrillic С are on the same [keyboard key](https://store.storeimages.cdn-apple.com/4982/as-images.apple.com/is/MLA22RS?wid=4000&hei=1800&fmt=jpeg&qlt=95&op_usm=0.5,0.5&.v=1496943898748), I guess I've started typing that word in cyrillic (it would be "Срщшсуы") and spotted error half-way through this word. And then instead of pressing option+backspace I've manually backspaced all characters except first one, because it looked fine.

PR #6, opened 2021-07-18
Files endpoint + answers endpoint bug fix + API error messages

blfletcher:
- Added a new files endpoint for:
  - list files
  - get file
- Corrected documents double array bug in answers endpoint
- Added error message if exists in API response


sashabaranov:
Merged, thank you for the PR!

PR #10, opened 2021-09-01
Add CreateCompletionWithFineTunedModel function

sauceman40:
OpenAI recently added the ability to use a fine-tuned model to generate completions with GPT-3.

This PR adds a function to implement this feature, which uses a different endpoint than CreateCompletion.

See here for the format: https://beta.openai.com/docs/guides/fine-tuning/use-a-fine-tuned-model

Lmk if any additional comments etc would be useful here.


sashabaranov:
Merged, thank you for the PR!

PR #11, opened 2021-09-30
Add omitempty to optional fields in AnswerRequest struct

markerdmann:


sashabaranov:
@markerdmann 🙌 thank you for the PR!

PR #21, opened 2022-07-01
rev: added a file to support moderations endpoint

galpt:
Added a file to support the new [Moderation endpoint](https://beta.openai.com/docs/guides/moderation/moderation).
Previously mentioned in https://github.com/sashabaranov/go-gpt3/issues/19#issue-1290828426.

### DIFF
--- a/moderation.go
+++ b/moderation.go
@@ -1,0 +1,69 @@
+package gogpt
+
+import (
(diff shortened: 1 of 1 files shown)


sashabaranov:
Thank you for the PR! 🙌🏻

PR #24, opened 2022-07-02
fix: Choice redeclared in this block

galpt:
This fixes the following errors:
- `edits.go:28:6: Choice redeclared in this block`
- `completion.go:31:6: other declaration of Choice`


galpt:
sorry, I for got something~
will fix it and make another PR asap.

Issue #30, opened 2022-08-02
Add a Mocked Server for Testing


RobotSail:
Based on a request from #27, we should introduce a mocked server which tests that values are correctly sent by `go-gpt3` and received by the server, as well as the ability for the client to receive data from the server.

This can help us to quickly iterate on tests without spending any OpenAI credits.

We need the following to complete this task
- [x] Create a mocked OpenAI API server
- [ ] Write endpoint handlers and tests for the mocked server:
  - [x] Completions
  - [x] Edits
  - [ ] Moderation
  - [ ] Search
  - [ ] Embeddings
  - [ ] Answers
  - [ ] Files
  - [ ] Classifications

vimitdhawan:
@RobotSail Do you need any help on this?

RobotSail:
Hi @vimitdhawan, we merged #31 last year, and I haven't touched the others once since. If you wanted to contribute some tests for these other points, that would be amazing.

Rascal0814:
The testing of the file is not done yet, can I contribute a bit to this?

Rascal0814:
> The testing of the file is not done yet, can I contribute a bit to this?

And I think it is unreasonable to put all the tests in the api_test file, and the functions of the same nature should be put in one file

drkennetz:
@RobotSail Hi. Client should be an interface to a `client` which implements the methods. That would solve all your testing problems. Then you could just setup a mock on your interface. Here's a simple example:

```
type Client interface {
    ListEngines(ctx context.Context) (engines EnginesList, err error)
}

type client struct {
    config ClientConfig
}

func (c *client) ListEngines(ctx context.Context) (engines EnginesList, err error) {
...
}

type MockClient struct {
    mock.Mock
}

func (c *MockClient) ListEngines(ctx context.Context) (engines EnginesList, err error) {
    args := c.Called(ctx)
    return args.Get(0).(EnginesList), args.Error(1)
}
```

This is untested but should be the general format. I'd be happy to try to take it on, it would just be a pretty large refactor but would make it easier for users (like me) to test the Client.

sashabaranov:
@drkennetz, the approach you are talking about would generally be correct if there were some business logic in `Client`. However, `Client` is the whole business logic we have here, you don't get it tested if you mock it.

RobotSail:
After some further thought, do we really even need the rest of these tests? 
[...]

PR #31, opened 2022-08-02
Add OpenAI Mock Server

RobotSail:
This PR introduces the following changes:

- `EditsUsage` and `CompletionsUsage` are removed and `Usage` is created as a separate type in `common.go` 
- Some available models are added as constants for the completions endpoint so that they can be included in the `CompletionRequest.Model` field
- A mocked server is added for testing the edits and completions endpoint 
- Makefile along with targets to downloads to install golangci-lint
- Fixes up the code so that the linter will pass
- Includes the golden [golangci-lint file](https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322)

Note: not all of the endpoints have been implemented with the mocked server yet, and we should continue to do so as more changes are iterated on in this project. 

This PR aims to close out #30, however we should aim to increase the amount of testing done and add coverage data.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,6 @@ help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
 
-
 ##@ Development
 
 .PHONY: test
@@ -26,3 +25,11 @@ test: ## Test the Go modules within this package.
 	@ echo ▶️ go test $(TEST_ARGS) $(TEST_TARGETS)
 	go test $(TEST_ARGS) $(TEST_TARGETS)
 	@ echo ✅ success!
+
+
+.PHONY: lint
+LINT_TARGETS ?= ./...
+lint: ## Lint Go code with the installed golangci-lint
+	@ echo "▶️ golangci-lint run"
+	golangci-lint run $(LINT_TARGETS)
+	@ echo "✅ golangci-lint run"
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,28 @@
+##@ General
+
+# The help target prints out all targets with their descriptions organized
+# beneath their categories. The categories are represented by '##@' and the
+# target descriptions by '##'. The awk commands is responsible for reading the
+# entire set of makefiles included in this invocation, looking for lines of the
+# file as xyz: ## something, and then pretty-format the target and help. Then,
+# if there's a line with ##@ something, that gets pretty-printed as a category.
+# More info on the usage of ANSI control characters for terminal formatting:
+# https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters
+# More info on the awk command:
+# http://linuxcommand.org/lc3_adv_awk.php
+
+.PHONY: help
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+
+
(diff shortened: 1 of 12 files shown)


sashabaranov:
This is so cool. Thank you for the PR! Got

Makefile
```diff
@@ -0,0 +1,61 @@
...
+OS := $(shell go env GOOS)
+ARCH := $(shell go env GOARCH)
+
+.PHONY: golangci-lint
+GOLANGCILINT := $(LOCALBIN)/golangci-lint
+GOLANGCI_URL := https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh
+golangci-lint: $(GOLANGCILINT) ## Download golangci-lint
+$(GOLANGCILINT): $(LOCALBIN)
+	@ echo "▶️ Downloading golangci-lint"
+	curl -sSfL $(GOLANGCI_URL) | sh -s -- -b $(LOCALBIN) $(GOLANGCI_VERSION)
```
sashabaranov, line 60: There are different opinions on the `curl | sh` pattern, I mostly think of it as an antipattern. Can we remove downloading here and just error out if golangci-lint is not present?

sashabaranov: Note: we are also using the golangci-lint [action](https://github.com/golangci/golangci-lint-action),

answers.go
```diff
@@ -33,19 +33,23 @@ type AnswerResponse struct {
 }
 
 // Search — perform a semantic search api call over a list of documents.
-func (c *Client) Answers(ctx context.Context, request AnswerRequest) (response AnswerResponse, err error) {
+func (c *Client) Answers(ctx context.Context, request AnswerRequest) (AnswerResponse, error) {
```
sashabaranov, line 36: I'm not sure this helps code readability, maybe we should just relax linter settings?

Curious why Github's golangci-lint did not report this.

api_test.go
```diff
@@ -1,11 +1,24 @@
...
+	"time"
+
+	. "github.com/sashabaranov/go-gpt3"
+)
+
+const (
+	testAPIToken = "this-is-my-secure-token-do-not-steal!!"
 )
 
 func TestAPI(t *testing.T) {
```
sashabaranov, line 24: Currently, running `make test` fails due to this test. Maybe we should mark it as a [long test](https://stackoverflow.com/questions/55180613/how-do-i-write-test-in-go-that-make-use-of-the-short-flag-and-can-it-be-combine), and only run short tests in the `make test`?

api_test.go
```diff
@@ -70,6 +73,60 @@ func TestAPI(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("CreateCompletion error: %v", err)
+	}
+}
+
+// TestEdits Tests the edits endpoint of the API using the mocked server.
+func TestEdits(t *testing.T) {
+	// create the test server
+	var err error
+	ts := OpenAITestServer()
```
sashabaranov, line 103: ❤️

sashabaranov (changes requested):
Awesome work! My concerns here are mostly about
- `curl | sh` for golang-ci setup
- golang-ci config strictness and related changes to make linter happy

answers.go
RobotSail: I added a `.golangci.yaml` file which has a bunch of settings enabled, one of which is no-named imports. We can just disable it

api_test.go
RobotSail: @sashabaranov What issues are you running into with this test? It's been running fine on mine

sashabaranov: @RobotSail this test requires `.openai-token` to be tested against real API. I suggest we disable this test to run by default and only run your new awesome one with a mock server!

RobotSail: @sashabaranov I set this test to be skipped in short mode. In the long-term, I think we should migrate the other tests in short mode to also have mocked methods (I don't have experience with the other APIs, so I didn't include them in this PR).
My only concern with replacing all of the current tests with our mocked ones is the accuracy of the client.

One workaround to this could be having an OpenAI token set in the repo, and then running tests against the API after the runs are approved in order to keep costs down.

sashabaranov: Awesome work!

>  One workaround to this could be having an OpenAI token set in the repo, and then running tests against the API after the runs are approved in order to keep costs down.

I've contacted @openai about the test token for the CI, but they didn't reply back.

sashabaranov (approved):

sashabaranov:
Great work, thank you for the PR! 🔥

api_test.go
RobotSail: @sashabaranov We've also asked them about having test endpoints to be able to test without burning tokens, they said that it was something that they were considering, but I've yet to see it.

PR #44, opened 2022-12-15
New Ada Similarity Model

mtrefilek:
Just earlier today OpenAI released "text-similarity-ada-002" Their first v2 embeddings model. 
https://openai.com/blog/new-and-improved-embedding-model/
https://beta.openai.com/docs/guides/embeddings/embedding-models

This replaces the need for the 5 separate text models. It outperforms all except for davinci-001 on one specific test set. It is also more cost efficient.

(I should have waited one more hour before my last PR. Who would have known!)

### DIFF
--- a/embeddings.go
+++ b/embeddings.go
@@ -52,6 +52,7 @@ const (
 	AdaCodeSearchText
 	BabbageCodeSearchCode
 	BabbageCodeSearchText
+	AdaSimilarityV2
 )
 
 var enumToString = map[EmbeddingModel]string{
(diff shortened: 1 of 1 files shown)


embeddings.go
```diff
@@ -52,6 +52,7 @@ const (
 	AdaCodeSearchText
 	BabbageCodeSearchCode
 	BabbageCodeSearchText
+	AdaSimilarityV2
```
mtrefilek, line 55: I wasn't sure how to name this as it breaks the earlier convention of having multiple embedding models for different use cases. Please let me know if there are any alternate suggestions!

sashabaranov:
That was swift, thank you! @openai should step up their game and auto-generate APIs already.

sauceman40:
@sashabaranov @mtrefilek let me know if I'm misreading something, but on their blog page, the model seems to be named:

`text-embedding-ada-002`

When I try using AdaSimilarityV2 right now, I get the error `error, status code: 404, message: That model does not exist`

Can one of you update this?

Issue #49, opened 2023-01-05
Does this library support streaming?


nullbio:
I can see that the CompletionRequest takes a Stream bool, but I don't see anything in the code that would indicate streaming support or functionality. I could very easily be missing something though, so wanted to check with you. Thanks.

sashabaranov:
@nullbio Thank you for the question! Currently, the library does not support streaming, and contributions are very welcome 😄

nullbio:
No worries. Did you have some ideas about how the implementation would look? I can have a go at it, but I'd rather seek some guidance from you first in terms of how it should be done. And if not for me, for whoever else sees this issue and wants to do it.

sashabaranov:
@nullbio So it seems OpenAI [uses](https://beta.openai.com/docs/api-reference/completions/create#completions/create-stream) server-sent events, here's a great [intro](https://germano.dev/sse-websockets/) about them. 

Here are a couple of ideas about implementation:

• High-level API might look like something like that:
```go
stream := client.CreateCompletionStream(...)
for {
	dataPiece, err := stream.Recv()
	if err == io.EOF {
		// stream ended
	}

	// process dataPiece
}
// inspired by gRPC streaming https://grpc.io/docs/languages/go/basics/#client-side-streaming-rpc
```

• Here's an example implementation of an SSE client for Go: https://grciuta.medium.com/server-sent-events-with-go-server-and-client-sides-6812dca45c7

• It would be awesome if we could end up without using 3rd-party libraries here. Here's an example of such a library: https://github.com/r3labs/sse

• There's an `io.Reader` pattern in Go, it might be useful here https://medium.com/learning-the-go-programming-language/streaming-io-in-go-d93507931185

ealvar3z:
I just sent a PR with a possible implementation. Please note that I have not consumed this new feature as a client yet. This implementation simply passes all linting and test checks.

sashabaranov:
Wow, thank you for the PR, it's so cool! 🙌🏻

I've got a couple of comments below, I also would like to test as a client before we merge.

priuatus:
Hey @ealvar3z and @sashabaranov I was just looking for this feature. Glad you're already on it. 

I hope the API you come up with supports `io.Reader` as that would make it easy to use with all the other Reader/Writers in the standard library without extra boilerplate.

Just modifying the example usage to use `io.Copy` would look like that:
```go
package main

import (
    "context"
    "fmt"
    "io"
    "os"

[...]

PR #54, opened 2023-01-22
Add streaming support feature

ealvar3z:
Removes golangci linting deprecation warnings 
See: [Issue #49](https://github.com/sashabaranov/go-gpt3/issues/49)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -149,16 +149,13 @@ linters:
   disable-all: true
   enable:
     ## enabled by default
-    - unused # Finds unused code
     - errcheck # Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases
     - gosimple # Linter for Go source code that specializes in simplifying a code
     - govet # Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
     - ineffassign # Detects when assignments to existing variables are not used
     - staticcheck # Staticcheck is a go vet on steroids, applying a ton of static analysis checks
-    - unused # Finds unused struct fields
     - typecheck # Like the front-end of a Go compiler, parses and type-checks Go code
     - unused # Checks Go code for unused constants, variables, functions and types
-    - unused # Finds unused global variables and constants
     ## disabled by default
     # - asasalint # Check for pass []any as any in variadic func(...any)
     - asciicheck # Simple linter to check that your code does not contain non-ASCII identifiers
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -149,16 +149,16 @@ linters:
   disable-all: true
   enable:
     ## enabled by default
-    - deadcode # Finds unused code
+    - unused # Finds unused code
     - errcheck # Errcheck is a program for checking for unchecked errors in go programs. These unchecked errors can be critical bugs in some cases
     - gosimple # Linter for Go source code that specializes in simplifying a code
     - govet # Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
     - ineffassign # Detects when assignments to existing variables are not used
     - staticcheck # Staticcheck is a go vet on steroids, applying a ton of static analysis checks
-    - structcheck # Finds unused struct fields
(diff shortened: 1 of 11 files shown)


completion.go
```diff
@@ -106,3 +110,68 @@ func (c *Client) CreateCompletion(
...
+		return
+	}
+	defer res.Body.Close()
+
+	reader := bufio.NewReader(req.Body)
+	var line []byte
+
+	ch = make(chan string)
+
+	go func() {
```
sashabaranov, line 150: There's a [recommendation](https://github.com/golang/go/wiki/CodeReviewComments#synchronous-functions) to prefer synchronous functions (also see Concurrency section [here](https://medium.com/@cep21/aspects-of-a-good-go-library-7082beabb403)).

Could we remove the goroutine from here? We can try something similar to the goroutine-free API proposal [here](https://github.com/sashabaranov/go-gpt3/issues/49#issuecomment-1373423407).

sashabaranov: It might also make sense to move all streaming-related logic into separate file

ealvar3z: Yes, that makes complete sense (just learned something new  lol). I will refactor per your guidelines. Thank you for your assistance.

stream.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	reader := bufio.NewReader(res.Body)
+	var line []byte
+	var headerData = []byte("data: ")
+
+	var responses []CompletionResponse
+	for {
```
sashabaranov, line 50: This loop effectively makes it a non-streaming API. I suggest this method returns a struct like:

```go

type CompletionStream struct {
	reader *bufio.Reader
	response *http.Response
}

func (stream *CompletionStream) Recv() {
	// do the loop body logic with reader.ReadBytes, etc.
}

func (stream *CompletionStream) Close() {
	// close response body, reader, etc. (basically what you've done in defers)
}

func (c *Client) CreateCompletionStream(
	ctx context.Context,
	request CompletionRequest,
) (*CompletionStream, error) {
	// make request and construct stream, return before the loop
}
```

.golangci.yml
```diff
@@ -149,16 +149,16 @@ linters:
   disable-all: true
   enable:
     ## enabled by default
-    - deadcode # Finds unused code
```
sashabaranov, line 152: Note sure it's okay to disable those

sashabaranov (changes requested):

stream.go
ealvar3z: Very well. I will iterate and re-submit. Thank you for the guidance.

.golangci.yml
ealvar3z: I disabled those based on the WARN messages. They give out deprecation warnings and instructs the user to remove them. I can switch them back if you prefer.

cmd/main.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	gogpt "github.com/sashabaranov/go-gpt3"
+)
+
+func main() {
```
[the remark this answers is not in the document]
ealvar3z: I will remove this folder also. It was built to test the library, and I never removed it.

.golangci.yml
sashabaranov: You are correct, however, `unused` was already enabled in the linter, so we don't need to duplicate it. Fixed in the master branch. 🙌🏻

kkdai:
looking forward to this feature. :)

sashabaranov:
@ealvar3z do you mind I finish this up?

ealvar3z:
No Sir.

Thank you for your patience on this, as I haven’t been able to spend any of my off time on it.

Sent from my iPhone

On Feb 7, 2023, at 3:46 AM, sashabaranov ***@***.***> wrote:

﻿
CAUTION: This email originated from outside of the organization. Do not click links or open attachments unless you recognize the sender and know the content is safe.


@ealvar3z<https://github.com/ealvar3z> do you mind I finish this up?

PR #63, opened 2023-02-09
Modify the test module, add the file upload test, and add the image edit api

Rascal0814:
- Because the test files are all written in one file, I think it is unreasonable to make changes
- File upload test
- Add image edit api and corresponding test file
- An adjustment has been made to OpenAITestServer. The test needs to register the handler with the server


sashabaranov:
Thank you for the PR! 🙌🏻

image_test.go
```diff
@@ -0,0 +1,158 @@
...
+	ctx := context.Background()
+	client.BaseURL = ts.URL + "/v1"
+
+	origin, err := os.Open("./static/image_edit_original.png")
+	if err != nil {
+		t.Error("open origin file error")
+		return
+	}
+
+	mask, err := os.Open("./static/image_edit_mask.png")
```
sashabaranov, line 105: It's a bit of overkill bringing static images into the repo to test with the mock server

internal/api/server.go
```diff
@@ -0,0 +1,43 @@
+package api
```
sashabaranov, line 1: May I suggest we place it into `/test` or `/testing` or `/internal/test[ing]`? The `/internal/api` might look like it's not test-related.

See also: https://github.com/golang-standards/project-layout

internal/api/server.go
```diff
@@ -0,0 +1,43 @@
...
+			return
+		}
+		handler(w, r)
+	}))
+}
+
+type Handler func(w http.ResponseWriter, r *http.Request)
+
+// serverMap.
+var serverMap = make(map[string]Handler)
```
sashabaranov, line 38: Not sure that global map is a good idea here, maybe we can do something along the lines of:

```go
type TestServer struct {
	handlers map[string]Handler
}

func NewTestServer() *TestServer { ... }
func (ts *TestServer) RegisterHandler(path string, handler Handler) { ... }
```

Thanks!

internal/api/server.go
sashabaranov: Maybe we can even place this file on the root level and make the whole test server stuff non-exportable

sashabaranov (changes requested):
This is an excellent PR! Just a couple of minor comments 🚀

internal/api/server.go
Rascal0814: Should the map in RegisterHandler be determined whether it exists or directly overwritten?

internal/api/server.go
Rascal0814: Although internal cannot be exported, in order to simplify the project structure, it is better to put it in the root directory

image_test.go
Rascal0814: 👌

internal/api/server.go
Rascal0814: I think global map is feasible here, and there is no need to consider concurrency. Maybe I think it is simple, don't you think?

internal/api/server.go
Rascal0814: Another problem is that the package of the test file is different from that of the normal file. Therefore, if the method is set to not exportable, it cannot be accessed

internal/api/server.go
sashabaranov: We also have `GetTestToken` here, I think it makes sense to keep test-server-related logic encapsulated in a single tidy struct, in that way, it would be obvious how to extend the test server in the future.

Rascal0814: Well, I'll make it simple

sashabaranov (approved):

sashabaranov:
Thank you for all the changes!

PR #68, opened 2023-02-13
Improve error reporting

DarkDust:
The errors returned by `sendRequest` did not allow for easy error inspection. This PR aims to improve this.

To be able to test what type of error was returned by the server, an `APIError` struct is introduced which is basically the previous `ErrorResponse.Error` struct, but with three changes:

* The `Code` field's type has been fixed, it's a `string`.
* An additional field `StatusCode int` field was added to transport the HTTP status code as well.
* Implements the `Error` interface.

There's also a new `RequestError` for errors which do not provide an API error object.

The errors returned by `sendRequest` now make use of error wrapping via `%w`. This way the returned error strings retain their previous format but `errors.Is` and `errors.As` now work to get access to the additional error information.

### DIFF
--- a/api.go
+++ b/api.go
@@ -66,7 +66,11 @@ func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 		var errRes ErrorResponse
 		err = json.NewDecoder(res.Body).Decode(&errRes)
 		if err != nil || errRes.Error == nil {
-			return fmt.Errorf("error, status code: %d", res.StatusCode)
+			reqErr := RequestError {
+				StatusCode: res.StatusCode,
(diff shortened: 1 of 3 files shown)


sashabaranov:
Thank you for the PR, it's really thoughtful! 🔥 

I would love to merge once linter is happy 😄

sashabaranov (approved):

DarkDust:
That was fast, thank you very much!

sashabaranov:
@DarkDust sure, thank you so much! Future note: I think there might be some room for error-reporting improvements in `stream.go` if you are up for it :D

PR #75, opened 2023-02-15
Configurable Transport

mwillfox:
Non-breaking changes which allow for a configurable "Transport" (language which was already used in the project, really a `*http.Client`).

This allows for setting a custom `RoundTripper` or using `github.com/hashicorp/go-retryablehttp`:

```go
transport := retryablehttp.NewClient().StandardClient()
c := gogpt.NewClientWithTransport(openAIAPIKey, transport)
```

With the recent stability issues and load of the OpenAI API this is necessary to use the module in a production environment.

### DIFF
--- a/api.go
+++ b/api.go
@@ -30,6 +30,16 @@ func NewClient(authToken string) *Client {
 	}
 }
 
+// NewClientWithTransport creates new OpenAI API client with the provided http.Client.
+func NewClientWithTransport(authToken string, transport *http.Client) *Client {
+	return &Client{
+		BaseURL:    apiURLv1,
+		HTTPClient: transport,
+		authToken:  authToken,
+		idOrg:      "",
+	}
+}
+
 // NewOrgClient creates new OpenAI API client for specified Organization ID.
 func NewOrgClient(authToken, org string) *Client {
 	return &Client{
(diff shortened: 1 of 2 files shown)


sashabaranov:
Hey, thank you for that PR! It made me think about how we want to approach client initialization — there was a great article about different ways of doing that, but I could not find it. The closest is this one: https://asankov.dev/blog/2022/01/29/different-ways-to-initialize-go-structs/

I don't think the usage of `.WithA().WithB().WithC()` chaining is a way forward for us here. I would rather prefer an API akin to:

```go
client := gogpt.NewClient()

// or

config := gogpt.DefaultConfig()
config.Transport = retryablehttp.NewClient().StandardClient()
client := gogpt.NewClientWithConfig(config)
```

@mwillfox would you have the bandwidth to do such a refactoring? It's totally cool if not — I can do it in a follow-up PR.

mwillfox:
I should have time to refactor in the next few days and I'll update the PR. Thanks for the feedback!

sashabaranov:
Will follow up this work in a `better-config` branch

Eriddle96 (commented):
This is a great idea

PR #77, opened 2023-02-18
Fix endless loop while network error

yangyanqing:
An endless loop occurs if ReadBytes() returns non-IO error, such as network interrupt.

```go
waitForData:
	line, err := stream.reader.ReadBytes('\n')
	if err != nil {
		if errors.Is(err, io.EOF) {
			return
		}
	}

	var headerData = []byte("data: ")
	line = bytes.TrimSpace(line)
	if !bytes.HasPrefix(line, headerData) {
		emptyMessagesCount++
		if emptyMessagesCount > emptyMessagesLimit {
			err = ErrTooManyEmptyStreamMessages
			return
		}

		goto waitForData
	}
```
 
image

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 )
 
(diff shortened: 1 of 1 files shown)


sashabaranov:
Such a good catch, thank you! 🙌🏻

sashabaranov (approved):

PR #79, opened 2023-02-19
Better configuration

### LINKED ISSUE #70: support custom  stream `emptyMessagesLimit`

sashabaranov:
This PR aims to make the client completely configurable, following up #75 and #4. It also fixes #70.

The idea is to pass-by-value config struct, which would be encapsulated inside the client.

Here we preserve the standard way of creating clients as `client := gogpt.NewClient("auth token")` and add a new way of configuring clients:

```go
config := DefaultConfig("auth token")
config.BaseURL = server.URL + "/v1"
config.OrgID = "my org id"
config.HTTPClient.Transport = &tokenRoundTripper{
	test.GetTestToken(),
	http.DefaultTransport,
}
config.EmptyMessagesLimit = 10000

client := NewClientWithConfig(config)
ctx := context.Background()
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -4,7 +4,10 @@ import (
 	"net/http"
 )
 
-const apiURLv1 = "https://api.openai.com/v1"
+const (
+	apiURLv1                       = "https://api.openai.com/v1"
+	defaultEmptyMessagesLimit uint = 300
+)
 
 // ClientConfig is a configuration of a client.
 type ClientConfig struct {
@@ -25,6 +28,6 @@ func DefaultConfig(authToken string) ClientConfig {
 		OrgID:      "",
 		authToken:  authToken,
 
-		EmptyMessagesLimit: 300,
+		EmptyMessagesLimit: defaultEmptyMessagesLimit,
 	}
 }
--- a/config.go
+++ b/config.go
@@ -1,0 +1,24 @@
+package gogpt
+
+import (
+	"net/http"
+)
+
+const apiURLv1 = "https://api.openai.com/v1"
+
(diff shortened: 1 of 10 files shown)


sashabaranov:
@mwillfox @yangyanqing @DarkDust @ealvar3z @Rascal0814 @RobotSail  would love your review!

config.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const (
+	apiURLv1                       = "https://api.openai.com/v1"
+	defaultEmptyMessagesLimit uint = 300
+)
+
+// ClientConfig is a configuration of a client.
+type ClientConfig struct {
+	authToken string
```
DarkDust, line 14: Why is `authToken` private, but the other fields are public?

RobotSail:
LGTM

config.go
sashabaranov: Great question! I made it public at first but then figured that I couldn't imagine a case where it needs to be modified or explicitly accessed. 

I also thought it would prevent accidental logging, but it seems that it is not the case.

Should I make it public?
🤔 🤔 🤔

mwillfox:
This satisfies my original use case of plugging in `retryablehttp` as the transport and I agree is a cleaner way of doing config than my original PR. Looks good!

config.go
sashabaranov: @mwillfox btw, it leaks if you do `%v` printing of the struct

sashabaranov:
Merged and published in https://github.com/sashabaranov/go-gpt3/releases/tag/v1.2.0

PR #97, opened 2023-03-02
refactor: use http.NewRequestWithContext instead of http.NewRequest

bestgopher:


sashabaranov:
Thank you for the PR! The `http.MethodPost` looks cleaner 🙌🏻 Are there any other non-obvious benefits from such an approach to building requests?

sashabaranov (approved):

bestgopher:
> Thank you for the PR! The `http.MethodPost` looks cleaner 🙌🏻 Are there any other non-obvious benefits from such an approach to building requests?

https://github.com/golang/go/blob/02411bcd7c8eda9c694a5755aff0a516d4983952/src/net/http/request.go#L354-L355.
I think it's a better choice.

PR #106, opened 2023-03-03
chat completion streaming

itxiaolin:
https://github.com/sashabaranov/go-gpt3/issues/102


sashabaranov:
@itxiaolin Thank you for the PR! We've just merged the chat streaming call in https://github.com/sashabaranov/go-gpt3/pull/101, but more tests are welcome 🙌🏻

Issue #80, opened 2023-02-20
Code examples


sashabaranov:
I think we can do a better job with documentation by providing testable examples https://go.dev/blog/examples

LordThevenin:
@sashabaranov I would love to contribute to this project, can I pick this issue?

sashabaranov:
@LordThevenin yes, please! It would be cool to move out all tests with real API [token](https://github.com/sashabaranov/go-openai/blob/master/api_test.go#L15) into testable examples:.

PR #112, opened 2023-03-03
Add Http Client Transport Support

xusenlin:
Set up a request proxy for the client, I think many people need this feature.


sashabaranov:
Hey, thank you for the PR! This method is too specific for this library's API and is inconsistent with our configuration approach (see #79). Having said that — I'm open to having a proxy configuration example in docs or as a testable example (see #80).

PR #116, opened 2023-03-03
fix json tag

bestgopher:
image

Maybe there are some error.
Should we change the field's name from CreatedAt to Created?


sashabaranov:
Thank you for fixing this!

sashabaranov (approved):

Issue #118, opened 2023-03-04
The authToken cannot be passed in when setting the http proxy  


yxk756:
authToken cannot be passed in when setting http proxy. When setting custom config using NewClientWithConfig(), authToken cannot be set externally because authToken is in lower case. You want to provide a way to set authToken.

eryajf:
You can take a look at my example here.

https://github.com/eryajf/chatgpt-dingtalk/blob/6d4d1df8087e6c661ee4f65b1dc442a7eef6de05/pkg/chatgpt/chatgpt.go#L40-L62

yxk756:
> You can take a look at my example here.
> 
> https://github.com/eryajf/chatgpt-dingtalk/blob/6d4d1df8087e6c661ee4f65b1dc442a7eef6de05/pkg/chatgpt/chatgpt.go#L40-L62

谢谢！

vvatanabe:
I'm closing this because it appears to have already been resolved.

PR #122, opened 2023-03-04
Add a setAuthToken method

wuqinqiang:
#118


sashabaranov:
@wuqinqiang Thank you for the PR! Could you please explain what the use case for this method is? See also similar PR recently https://github.com/sashabaranov/go-openai/pull/112

wuqinqiang:
I didn't notice the DefaultConfig at first. so  I thought I couldn't set a token, emm....

PR #123, opened 2023-03-04
add "name" property for ChatCompletionMessage

fussraider:
I found an undocumented possibility - for messages of the gpt-3.5 model, you can additionally specify the "name" property. This additional feature is useful for chats with multiple participants.

Found in the examples of the official library for python (Item No. 6 - Counting tokens for chat API calls): [https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken .ipynb](https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb)

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -59,7 +59,7 @@ func TestAPI(t *testing.T) {
 			Model: GPT3Dot5Turbo,
 			Messages: []ChatCompletionMessage{
 				{
-					Role:    "user",
+					Role:    ChatMessageRoleUser,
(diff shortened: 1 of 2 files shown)


sashabaranov:
Nice find! Thank you for the PR 🙌

Could you please add a comment in code about that field being non-documented in the official docs?

fussraider:
Yes, no problem. Added 👌

sashabaranov (approved):

PR #124, opened 2023-03-05
Add speech to text example in docs

hoani:
Adds an example for audio speech to text in README documentation.


sashabaranov:
Thank you for the PR! 🙌🏻

sashabaranov (approved):

Issue #132, opened 2023-03-08
Issue on embedding type


IANTHEREAL:
The vector embedding type definition used by common vector engines are all []float32, such as [weaviate](https://github.com/weaviate/weaviate-go-client/blob/166fe5a522098db1ff0f2738605f72142995db65/weaviate/graphql/hybridbuilder.go#L23). Would go-openai consider changing `Embedding []float64` to `Embedding []float32`?

sashabaranov:
Just added this in v1.5.0 https://github.com/sashabaranov/go-openai/releases/tag/v1.5.0

thank you for the suggestion!

Issue #66, opened 2023-02-12
Embeddings upgrade


sashabaranov:
Current embeddings API might use an improvement:

- Use `float32` instead of `float64` (I highly doubt that OpenAI uses 64-bit floats in their LLM)
- Add some basic operations between embeddings, like cosine similiarity (which is equal to dot product due to embeddings being normalized to length 1)

ealvar3z:
Are you envisioning something along the lines of [OpenAI's embedding utils](https://github.com/openai/openai-python/blob/main/openai/embeddings_utils.py)?

sashabaranov:
@ealvar3z `embeddings_utils.py` is mainly focused on plotting, which is out of scope for this library as well as many other ML-specific use cases. However, having cosine similarity (= dot product) and embedding addition/subtraction should be helpful.

PR #133, opened 2023-03-08
use float32 in embeddings

sashabaranov:
#66 #132

### DIFF
--- a/embeddings.go
+++ b/embeddings.go
@@ -103,7 +103,7 @@ var stringToEnum = map[string]EmbeddingModel{
 // then their vector representations should also be similar.
 type Embedding struct {
 	Object    string    `json:"object"`
-	Embedding []float64 `json:"embedding"`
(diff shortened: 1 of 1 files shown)


embeddings.go
```diff
@@ -103,7 +103,7 @@ var stringToEnum = map[string]EmbeddingModel{
 // then their vector representations should also be similar.
 type Embedding struct {
 	Object    string    `json:"object"`
-	Embedding []float64 `json:"embedding"`
+	Embedding []float32 `json:"embedding"`
```
robrohan, line 106: We are trying to track down a bug, and I think this change causes problems when you use Python to store embeddings and golang to try to do similarity scores against them. Python seems to be using float64 (or something with even more precision), and when we try to use those values with the embeddings from this library we get wildly different results then when we do the same code using Python only. I am not sure if the endpoint returns 64, but it would seem down casting to 32 bit make more sense then losing precision you can't get back.

PR #153, opened 2023-03-14
Add image variation implementation and fix #149

### LINKED ISSUE #149: The CreateEditImage API should have an optional Mask parameter.
In image.go func CreateEditImage,
Mask is required in current implementation:
       // mask
	mask, err := writer.CreateFormFile("mask", request.Mask.Name())
	if err != nil {
		return
	}
But it is optional in official docs:
https://platform.openai.com/docs/api-reference/images/create-edit
/*
mask
string
Optional
An additional image whose fully transparent areas (e.g. where alpha is zero) indicate where image should be edited. Must be a valid PNG file, less than 4MB, and have the same dimensions as image.
*/

itegel:
1. Fix issue #149 
2. Add image variation implementation according to 
    https://platform.openai.com/docs/api-reference/images/create-variation

### DIFF
--- a/image.go
+++ b/image.go
@@ -122,15 +122,16 @@ func (c *Client) CreateEditImage(ctx context.Context, request ImageEditRequest) 
 	return
 }
 
-// ImageVariationRequest represents the request structure for the image API.
-type ImageVariationRequest struct {
+// ImageVariRequest represents the request structure for the image API.
+type ImageVariRequest struct {
 	Image *os.File `json:"image,omitempty"`
 	N     int      `json:"n,omitempty"`
 	Size  string   `json:"size,omitempty"`
 }
 
-// ImageVariationRequest - API call to create an image variation. This is the main endpoint of the DALL-E API.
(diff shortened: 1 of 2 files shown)


sashabaranov:
@itegel there's some issue with CI workflow, could you please re-base on the `master` branch?

itegel:
> @itegel there's some issue with CI workflow, could you please re-base on the `master` branch?
Hi, 
1. I`ve fixed the lint problem. 
2. The Codeball problem seems to be related to repo permissions. Here is the suggestion from the codeball owner, you may need to install "Codeball AI Writer" on the repo. I do not know to fix it, please help me 

/*
Forks and public repositories
GitHub does not offer (and reasonably so) a way for Pull Requests from a fork to a public repository to run with "write" permissions to the parent repository.
If you want to use Codeball on a public repository, install the ["Codeball AI Writer"](https://github.com/apps/codeball-ai-writer) app on the parent repository. This allows the Codeball Action to use the permissions from the app as a fallback if the action is running without write permissions.
*/

sashabaranov (approved):

sashabaranov:
looks good, thank you!

Issue #9, opened 2021-08-07
`omitempty` field of struct will generate incorrect request when parameter is 0.


ArthurAmes:
The 'omitempty' field of the struct should be removed. This is because it generates an incorrect request when a parameter is 0. For example, a request with the temperate field set to '0' will actually return a response as if the field had been set to one. This is because the go json parser does not differentiate between a 0 value and no value for float32.

sashabaranov:
Thank you for the issue! So, with `omitempty` we have:

* "Empty" (zero-valued) arguments are removed
* OpenAI's backend uses its default for omitted arguments
* Zeroing floats requires something like `1e-45 ≈ math.SmallestNonzeroFloat32`

Without `omitempty`:

* Non-specified arguments are zero-valued and present in request json
* OpenAPI's backend does not apply its defaults (which may or may not be equal to zero-values)
* In case OpenAPI defaults ≠ zero-values, we may have a default request value which can be tuned

I would check whether OpenAPI defaults match Go's zero-values. In case they do I think it is best to remove `omitempty`.

sashabaranov:
Looking at `CompletionRequest` in [completion.go](https://github.com/sashabaranov/go-gpt3/blob/master/completion.go) it seems that OpenAPI's default parameter values do not match Go's zero-values:

```go
type CompletionRequest struct {
	Prompt    string `json:"prompt,omitempty"`
	MaxTokens int    `json:"max_tokens,omitempty"` // OpenAPI default to 16

	Temperature float32 `json:"temperature,omitempty"` // OpenAPI default to 1
	TopP        float32 `json:"top_p,omitempty"` // OpenAPI default to 1

	N int `json:"n,omitempty"` // OpenAPI default to 1

	LogProbs int `json:"logprobs,omitempty"`

	Model *string `json:"model,omitempty"`

	Echo bool     `json:"echo,omitempty"`
	Stop []string `json:"stop,omitempty"`

	PresencePenalty  float32 `json:"presence_penalty,omitempty"` // OpenAPI default to 0
	FrequencyPenalty float32 `json:"frequency_penalty,omitempty"` // OpenAPI default to 0
	BestOf           int     `json:"best_of,omitempty"` // OpenAPI default to 1
}
```

ArthurAmes:
[...]

PR #163, opened 2023-03-15
fix temperature value zero bug

icyxp:


sashabaranov:
Hey @icyxp, thank you for the PR! The removal of `omitempty` would not work for us, please refer to this discussion: https://github.com/sashabaranov/go-openai/issues/9

PR #170, opened 2023-03-16
修改上传图片提示："error, status code: 400, message: Uploaded image must be a …

a289459798:
write.CreateFormFile default content-type is 'application/octet-stream'，but openai response error '"error, status code: 400, message: Uploaded image must be a PNG'


sashabaranov:
@a289459798 thank you for the PR! Please fix the linter issue and let's merge 🙌🏻

Issue #164, opened 2023-03-15
streaming data


imsprojo2fan:
hi , does the latest version still not support streaming data

sashabaranov:
it does support streaming, please see the README

imsprojo2fan:
> it does support streaming, please see the README

I mean chatgpt support stram?

sashabaranov:
Yes, https://github.com/sashabaranov/go-openai/blob/a8acb5f63bcb4d5966d1a57709a32315eeff1ff9/chat_stream_test.go#L53

imsprojo2fan:
> Yes,here's an example in the test
> 
> https://github.com/sashabaranov/go-openai/blob/a8acb5f63bcb4d5966d1a57709a32315eeff1ff9/chat_stream_test.go#L53

fine,thank you

PR #177, opened 2023-03-19
Add readme example of ChatGPT streaming completion

ryuichi-maeda:
When using `ChatCompletionStream`, I implemented the code by referencing the following test.

https://github.com/sashabaranov/go-openai/blob/a8acb5f63bcb4d5966d1a57709a32315eeff1ff9/chat_stream_test.go#L53

As it seems that #164 encountered the same issue, I  added an example of usage to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,61 @@ func main() {
 Other examples:
 
 <details>
+<summary>ChatGPT streaming completion</summary>
+
+```go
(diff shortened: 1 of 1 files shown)


sashabaranov:
thank you for the PR!

sashabaranov (approved):

PR #180, opened 2023-03-20
refactor: Refactor endpoint and model compatibility check

j178:
- Add model check for `CreateCompletionStream` and `CreateChatCompletionStream`.
- Correct model check in `CreateCompletion`.

### DIFF
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -86,15 +86,14 @@ func (c *Client) CreateChatCompletionStream(
 	ctx context.Context,
 	request ChatCompletionRequest,
 ) (stream *ChatCompletionStream, err error) {
-	switch request.Model {
-	case GPT3Dot5Turbo0301, GPT3Dot5Turbo, GPT4, GPT40314, GPT432K0314, GPT432K:
-	default:
-		err = ErrChatCompletionInvalidModel
+	urlSuffix := "/chat/completions"
+	err = checkEndpointSupportsModel(urlSuffix, request.Model)
+	if err != nil {
 		return
 	}
 
 	request.Stream = true
-	req, err := c.newStreamRequest(ctx, "POST", "/chat/completions", request)
+	req, err := c.newStreamRequest(ctx, "POST", urlSuffix, request)
 	if err != nil {
 		return
 	}
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -87,7 +87,7 @@ func (c *Client) CreateChatCompletionStream(
 	request ChatCompletionRequest,
 ) (stream *ChatCompletionStream, err error) {
 	urlSuffix := "/chat/completions"
-	if checkEndpointSupportsModel(urlSuffix, request.Model) {
+	if !checkEndpointSupportsModel(urlSuffix, request.Model) {
 		err = ErrChatCompletionInvalidModel
 		return
 	}
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -87,8 +87,8 @@ func (c *Client) CreateChatCompletionStream(
 	request ChatCompletionRequest,
 ) (stream *ChatCompletionStream, err error) {
 	urlSuffix := "/chat/completions"
-	err = checkEndpointSupportsModel(urlSuffix, request.Model)
-	if err != nil {
+	if checkEndpointSupportsModel(urlSuffix, request.Model) {
+		err = ErrChatCompletionInvalidModel
 		return
 	}
 
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -86,6 +86,13 @@ func (c *Client) CreateChatCompletionStream(
 	ctx context.Context,
 	request ChatCompletionRequest,
 ) (stream *ChatCompletionStream, err error) {
+	switch request.Model {
+	case GPT3Dot5Turbo0301, GPT3Dot5Turbo, GPT4, GPT40314, GPT432K0314, GPT432K:
+	default:
+		err = ErrChatCompletionInvalidModel
+		return
+	}
+
 	request.Stream = true
 	req, err := c.newStreamRequest(ctx, "POST", "/chat/completions", request)
 	if err != nil {
--- a/completion.go
+++ b/completion.go
@@ -7,8 +7,8 @@ import (
 )
 
 var (
-	ErrCompletionUnsupportedModel   = errors.New("this model is not supported with this method, please use CreateChatCompletion client method instead") //nolint:lll
-	ErrCompletionStreamNotSupported = errors.New("streaming is not supported with this method, please use CreateCompletionStream")                      //nolint:lll
+	ErrUnsupportedModel             = errors.New("this model is not supported with this method")                                   //nolint:lll
+	ErrCompletionStreamNotSupported = errors.New("streaming is not supported with this method, please use CreateCompletionStream") //nolint:lll
 )
 
 // GPT3 Defines the models provided by OpenAI to use when generating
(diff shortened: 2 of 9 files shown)


sashabaranov:
Hey, thank you for the PR!

stream.go
```diff
@@ -71,6 +71,13 @@ func (c *Client) CreateCompletionStream(
 	ctx context.Context,
 	request CompletionRequest,
 ) (stream *CompletionStream, err error) {
+	switch request.Model {
+	default:
+	case GPT3Dot5Turbo0301, GPT3Dot5Turbo, GPT4, GPT40314, GPT432K0314, GPT432K:
```
sashabaranov, line 76: Are we sure this endpoint does not support GPT4-based models?

completion.go
```diff
@@ -105,7 +105,9 @@ func (c *Client) CreateCompletion(
 		return
 	}
 
-	if request.Model == GPT3Dot5Turbo0301 || request.Model == GPT3Dot5Turbo {
+	switch request.Model {
+	default:
+	case GPT3Dot5Turbo0301, GPT3Dot5Turbo, GPT4, GPT40314, GPT432K0314, GPT432K:
```
sashabaranov, line 110: Are we sure this endpoint does not support GPT4-based models?

chat_stream.go
```diff
@@ -86,6 +86,13 @@ func (c *Client) CreateChatCompletionStream(
 	ctx context.Context,
 	request ChatCompletionRequest,
 ) (stream *ChatCompletionStream, err error) {
+	switch request.Model {
+	case GPT3Dot5Turbo0301, GPT3Dot5Turbo, GPT4, GPT40314, GPT432K0314, GPT432K:
```
sashabaranov, line 90: Are we sure this endpoint supports the `GPT4` model?

j178: Yes, as stated in the OpenAI documentation available at https://platform.openai.com/docs/models/model-endpoint-
  compatibility

completion.go
j178: Please check https://platform.openai.com/docs/models/model-endpoint-compatibility

chat_stream.go
```diff
@@ -86,6 +86,13 @@ func (c *Client) CreateChatCompletionStream(
...
+	default:
+		err = ErrChatCompletionInvalidModel
```
j178, line 92: This error message is not accurate anymore:
https://github.com/sashabaranov/go-openai/blob/aa149c1bf8ccd77403e07fdb7b0122f27d4c9b6b/chat.go#L16-L17

chat_stream.go
sashabaranov: Thank you for the link, it's really helpful! I wonder if we can have this model-checking in a single place, so each method can just have something like:

```go
endpoint := "/v1/completions"
err = checkEndpointSupportsModel(endpoint, request.Model)
if err != nil {
    return
}
```

sashabaranov: This way we'll be able to generate proper error messages, as per your comment below

j178:
Hi, I am wondering if it is necessary to have compatibility checks between the endpoint and model, or if it can be removed and let it fail. Here are some points to consider:

1. OpenAI will regularly release new models, which requires us to update the checks here in a timely manner, or else it will hinder users from using the new models.
2. If an incompatible model is used in the endpoint, OpenAI API will return a clear error message, which should be enough to inform users of the error.

sashabaranov:
@j178 What if we check for "incompability" — basically return error only when we are sure that the model is not supported?

j178:
> basically return error only when we are sure that the model is not supported?

While technically possible, this approach would be very long and difficult to maintain. We would need to explicitly list all unsupported models for each endpoint check, like this:

```go
func checkEndpointSupportsModel(endpoint, model string) error {
	switch endpoint {
	case "/completions":
		switch model {
		case GPT3Dot5Turbo0301, GPT3Dot5Turbo, GPT4, GPT40314, GPT432K, GPT432K0314:
			return ErrCompletionUnsupportedModel
		}
	case "/chat/completions":
		switch model {
		case GPT3TextDavinci003, GPT3TextDavinci002, GPT3TextCurie001, GPT3TextBabbage001, GPT3TextAda001, GPT3TextDavinci001, GPT3DavinciInstructBeta, GPT3Davinci, GPT3CurieInstructBeta, GPT3Curie, GPT3Ada, GPT3Babbage, CodexCodeDavinci002, CodexCodeCushman001, CodexCodeDavinci001:
			return ErrChatCompletionInvalidModel
		}
	}
	return nil
}
```

sashabaranov:
@j178 here's a simpler approach:

```go
type endpointModelPair struct {
	endpoint string
	model    string
}

var disabledPairs = map[endpointModelPair]bool{
	{"/completions", "gpt-3.5-turbo"}: true,
}

func main() {
	pairIsDisabled := disabledPairs[endpointModelPair{"/completions", "gpt-3.5-turbo"}]
	fmt.Println("Pair is disabled:", pairIsDisabled)
}
```

We can also use `map[string]map[string]bool` to have endpoint->model->disabled flag mapping.

j178:
I applied your suggestion, but I also had to introduce a breaking change: I deleted `ErrChatCompletionInvalidModel` and `ErrCompletionUnsupportedModel`, and replaced them with a unified `ErrUnsupportedModel`. Do you have any better suggestions?

completion.go
```diff
@@ -45,6 +45,42 @@ const (
...
 )
 
+type endpointModelPair struct {
+	endpoint string
+	model    string
+}
+
+var disabledPairs = map[endpointModelPair]bool{
+	{"/completions", GPT3Dot5Turbo}:                true,
+	{"/completions", GPT3Dot5Turbo0301}:            true,
```
sashabaranov, line 55: endpoint repetition looks a bit weird, what if we do something like this:

```go
var disabledModelsForEndpoints = map[string]map[string]bool{
	"/completions": map[string]bool{
		GPT3Dot5Turbo:     true,
		GPT3Dot5Turbo0301: true,
		GPT4:              true,
		GPT40314:          true,
		GPT432K:           true,
		GPT432K0314:       true,
	},
	"/chat/completions": map[string]bool{
		CodexCodeDavinci002:     true,
		CodexCodeCushman001:     true,
		CodexCodeDavinci001:     true,
		GPT3TextDavinci003:      true,
		GPT3TextDavinci002:      true,
		GPT3TextCurie001:        true,
		GPT3TextBabbage001:      true,
		GPT3TextAda001:          true,
		GPT3TextDavinci001:      true,
		GPT3DavinciInstructBeta: true,
		GPT3Davinci:             true,
		GPT3CurieInstructBeta:   true,
		GPT3Curie:               true,
		GPT3Ada:                 true,
		GPT3Babbage:             true,
	},
}
``

File: completion.go [hunk 6]
```diff
@@ -105,12 +141,12 @@ func (c *Client) CreateCompletion(
 		return
 	}
 
-	if request.Model == GPT3Dot5Turbo0301 || request.Model == GPT3Dot5Turbo {
-		err = ErrCompletionUnsupportedModel
+	urlSuffix := "/completions"
+	err = checkEndpointSupportsModel(urlSuffix, request.Model)
```
sashabaranov, line 145: we can make `checkEndpointSupportsModel` return a boolean flag, and use the previously defined errors e.g. `ErrChatCompletionInvalidModel`

j178:
The current CI failure is quite peculiar(I was able to pass the test locally). The panic stack trace appears at `stream_test.go:167` and `stream_test.go:169`, even though the `stream_test.go` file only has 165 lines in total.

```
=== RUN   TestCreateCompletionStreamError
    stream_test.go:167: CreateCompletionStream returned error: this model is not supported with this method, please use CreateChatCompletion client method instead
--- FAIL: TestCreateCompletionStreamError (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x88aac1]

goroutine 171 [running]:
testing.tRunner.func1.2({0x8c6f80, 0xbd9fc0})
	/opt/hostedtoolcache/go/1.19.7/x64/src/testing/testing.go:1396 +0x372
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.7/x64/src/testing/testing.go:1399 +0x5f0
panic({0x8c6f80, 0xbd9fc0})
	/opt/hostedtoolcache/go/1.19.7/x64/src/runtime/panic.go:890 +0x262
github.com/sashabaranov/go-openai_test.TestCreateCompletionStreamError(0xc00027fd40)
	/home/runner/work/go-openai/go-openai/stream_test.go:169 +0x581
testing.tRunner(0xc00027fd40, 0x954fa0)
	/opt/hostedtoolcache/go/1.19.7/x64/src/testing/testing.go:1446 +0x217
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.7/x64/src/testing/testing.go:1493 +0x75e
FAIL	github.com/sashabaranov/go-openai	0.196s
FAIL
Error: Process completed with exit code 1.
```

completion.go
```diff
@@ -45,6 +45,38 @@ const (
...
+		GPT3Davinci:             true,
+		GPT3CurieInstructBeta:   true,
+		GPT3Curie:               true,
+		GPT3Ada:                 true,
+		GPT3Babbage:             true,
+	},
+}
+
+func checkEndpointSupportsModel(endpoint, model string) bool {
+	return !disabledModelsForEndpoints[endpoint][model]
```
sashabaranov, line 77: The cause of the panics is that you don't check for nil here

sashabaranov: @j178 my bad, you are correct!

sashabaranov:
In terms of panic: I don't even see the `TestCreateCompletionStreamError` in your fork, could you please sync your form with the master branch?

PR #183, opened 2023-03-20
add optional params for audio api, e.g. prompt

itegel:
I found the prompt argument is useful in audio api, so added following optional args from official doc
https://platform.openai.com/docs/api-reference/audio/create :
	Prompt      string // For translation, it should be in English
	Temperature float32
	Language    string // For translation, just do not use it. It seems "en" works, not confirmed...

btw.  I found there already exist a similar pull request from someone else.  I added this change in my local env last week,  today I've done some test works(for real env), so It's just a coincidence....

It is ok, that just close this pr, I just want the changes merged ASAP~


sashabaranov:
thank you for the PR! 🙌🏻

sashabaranov (approved):

PR #184, opened 2023-03-20
fix: stream return EOF when openai return error

### LINKED ISSUE #173: ChatCompletionRequest with stream: true return EOF
When using ChatCompletionRequest and setting `stream: true` to send multiple conversation content, EOF is returned and no error information can be obtained.

liushuangls:
fix #173

### DIFF
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -39,7 +39,7 @@ type ChatCompletionStream struct {
 	errAccumulator *errorAccumulator
 }
 
-//nolint:duplicate
+//nolint:dupl // stream repetition code needs to be optimized
 func (stream *ChatCompletionStream) Recv() (response ChatCompletionStreamResponse, err error) {
 	if stream.isFinished {
 		err = io.EOF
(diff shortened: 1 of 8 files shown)


sashabaranov (changes requested):
Thank you for the PR! 🔥 Looks good overall, but there are a couple of comments. Would you be able to fix those?

chat_stream.go
```diff
@@ -48,12 +51,19 @@ func (stream *ChatCompletionStream) Recv() (response ChatCompletionStreamRespons
...
+		if errRes, _ := stream.errAccumulator.unmarshalError(); errRes != nil {
+			err = fmt.Errorf("error, %w", errRes.Error)
+		}
 		return
 	}
 
 	var headerData = []byte("data: ")
 	line = bytes.TrimSpace(line)
 	if !bytes.HasPrefix(line, headerData) {
+		if writeErr := stream.errAccumulator.write(line); writeErr != nil {
```
sashabaranov, line 63: We can cover this with tests if we make `errorAccumulator` an interface and provide failing one like here https://github.com/sashabaranov/go-openai/blob/aa149c1bf8ccd77403e07fdb7b0122f27d4c9b6b/request_builder_test.go#L30

unmarshaler.go
```diff
@@ -0,0 +1,15 @@
+package openai
+
+import (
+	"encoding/json"
+)
+
+type unmarshaler interface {
```
sashabaranov, line 7: Should we use that instead of json.Unmrashal?

sashabaranov (approved):

juggli:
nice. I meet the same EOF error in v1.5.2.

Issue #186, opened 2023-03-20
Use Assert package in the test files


rexposadas:
# Rationale

I suggest using the assert package in the tests.  See: https://github.com/stretchr/testify

It'll make the code cleaner. For example in api_test.go we have:

```go
	if err == nil {
		t.Fatal("ListEngines did not fail")
	}
```	

That can be changed to: 

```go	
assert.NoErrorf(t, err, "ListEngines did not fail")
```

# Implementation

I've implemented this feature for production level code and it makes the test code look a lot cleaner. 

If you agree with the suggestion, I am willing to make the PR for it.

sashabaranov:
@rexposadas Thank you for the suggestion! Our library aims to have no dependencies for as long as possible. While the testify library has some helpful features, they are not enough to make it a dependency. Having said that — we can add the single-line error checking in tests to the library. Your contribution would be greatly appreciated, along with any code coverage improvements 🙌🏻

rexposadas:
I see. No problem.

>  we can add the single-line error checking in tests to the library

Which library are you referring to? Would it be this one? https://github.com/sashabaranov/go-fastapi

rexposadas:
@sashabaranov But, to keep the code consistent would mean making other functions such as: 
 
checkHasError (..), checkErrorIs(..)

You're OK with that?

sashabaranov:
@rexposadas sounds good! It might also make sense to put those into [internal/test](https://github.com/sashabaranov/go-openai/tree/master/internal/test) package, but I trust your judgment here.

rexposadas:
Great. I'll work on the PR.

rexposadas:
PR for this has been merged.

PR #194, opened 2023-03-23
Refactor/internal testing

rexposadas:
See issue: https://github.com/sashabaranov/go-openai/issues/186

Add helpers in order to reduce boilerplate code in the tests.  

To test this PR run the tests: `go test ./...`.

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -136,9 +136,7 @@ func TestRequestError(t *testing.T) {
 	c := NewClientWithConfig(config)
 	ctx := context.Background()
 	_, err = c.ListEngines(ctx)
-	if err == nil {
-		t.Fatal("ListEngines did not fail")
-	}
+	checks.HasError(t, err, "ListEngines did not fail")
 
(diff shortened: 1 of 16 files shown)


sashabaranov:
Excellent work, thank you so much!

QQ: do we even need to factor `checks` into a subpackage? Single import of `test` and calls like `test.NoError(t, err, "ListEngines error")` seems like a nice API, what do you think?

rexposadas:
> Excellent work, thank you so much!
> 
> QQ: do we even need to factor `checks` into a subpackage? Single import of `test` and calls like `test.NoError(t, err, "ListEngines error")` seems like a nice API, what do you think?

Good question.  I didn't want the name to be too close with the std `testing` package. 

But if that worry is unwarranted, I can refactor it. I'm fine either way.

sashabaranov:
Got it, makes sense!

sashabaranov (approved):

Issue #175, opened 2023-03-19
The audio file cannot be deleted after calling Speech-To-Text API


MichaelWangGitHub2:
After calling the Audio Speech-To-Text API, the audio file cannot be deleted and a message appears indicating that it is being used.
![delvoice](https://user-images.githubusercontent.com/30011094/226167897-5b9ff5d7-e9d0-43d3-ac5a-2972f2ad4619.png)

elisongomes:
The fix was merged into master.

PR #195, opened 2023-03-23
fix: added defer to close audio file

elisongomes:
This PR solve the above issue:
- #175 

In some cases, the audio file opened for reading is not closed, making it impossible to remove it.

The `f.Close` call was added at the end of the execution of the `audioMultipartForm` function.


sashabaranov (approved):
Great catch, thank you! 

The audio.go is also ripe for unit testing if you'd be interested.

Issue #142, opened 2023-03-10
feature request: Add support for Azure OpenAI Service


tao12345666333:
As a user of the project, I would like to request the addition of support for Azure OpenAI. This would allow me to leverage the power of the Azure OpenAI platform, which offers cutting-edge AI technologies and services, including natural language processing, speech recognition, and computer vision.

I believe that adding support for Azure OpenAI would greatly enhance the functionality and usefulness of the project, and would be a valuable addition for users who are looking to integrate AI capabilities into their applications.

I look forward to hearing your thoughts on this proposal, and I am happy to provide additional information or assistance as needed. Thank you for considering this request.

https://azure.microsoft.com/en-us/blog/chatgpt-is-now-available-in-azure-openai-service/

Most of the APIs are the same, with only a few differences. refer to: https://learn.microsoft.com/en-us/azure/cognitive-services/openai/reference

Something like this needs to be adjusted or made configurable.
https://github.com/sashabaranov/go-openai/blob/9428f6cc3df02488c3ada31cfa7005c4c3d6c027/chat.go#L82

AtiqGauri:
+1 for this, we want to access azure openAI service, this would be a great enhancement to the library.

ziyixi:
Same here. We can also leverage the free credits provided by Azure, which is much more than OpenAI's $18.

meixiu:
When will the Azure open ai API be supported?

tao12345666333:
If no one handles it, I will pick it up. I saw some people gave it a +1, which shows that this feature is practical.

ttys3:
just filed a PR https://github.com/sashabaranov/go-openai/pull/214

PR #214, opened 2023-03-31
feat: add azure openai support

ttys3:
feat: add azure openai support

~~refactor config:~~
~~1. add `WithXXX` options, allow construct a config like  `openai.NewConfig(...Option)`~~
~~2. make the config field naming more like the `openai` python pkg's, this will make us follow py's changes more easy~~
~~3. add new `NewConfig` method, which will check the config just like the py' pkg (the error message almost the same)~~

resolve https://github.com/sashabaranov/go-openai/issues/142

### DIFF
--- a/api.go
+++ b/api.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 )
 
 // Client is OpenAI GPT-3 API client.
@@ -41,7 +42,7 @@ func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 	req.Header.Set("Accept", "application/json; charset=utf-8")
 	// Azure API Key authentication
 	if c.config.APIType == APITypeAzure {
-		req.Header.Set("api-key", c.config.authToken)
+		req.Header.Set(AzureAPIKeyHeader, c.config.authToken)
 	} else {
 		// OpenAI or Azure AD authentication
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.authToken))
@@ -91,8 +92,10 @@ func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 func (c *Client) fullURL(suffix string) string {
 	// /openai/deployments/{engine}/chat/completions?api-version={api_version}
 	if c.config.APIType == APITypeAzure || c.config.APIType == APITypeAzureAD {
-		return fmt.Sprintf("%s%s/%s/%s%s?api-version=%s",
-			c.config.BaseURL, azureAPIPrefix, azureDeploymentsPrefix, c.config.Engine, suffix, c.config.APIVersion)
+		baseURL := c.config.BaseURL
+		baseURL = strings.TrimRight(baseURL, "/")
+		return fmt.Sprintf("%s/%s/%s/%s%s?api-version=%s",
+			baseURL, azureAPIPrefix, azureDeploymentsPrefix, c.config.Engine, suffix, c.config.APIVersion)
 	}
 
 	// c.config.APIType == APITypeOpenAI || c.config.APIType == ""
--- a/api.go
+++ b/api.go
@@ -15,9 +15,12 @@ type Client struct {
 }
 
 // NewClient creates new OpenAI API client.
-func NewClient(authToken string) *Client {
-	config := DefaultConfig(authToken)
-	return NewClientWithConfig(config)
+func NewClient(authToken string) (*Client, error) {
+	config, err := DefaultConfig(authToken)
+	if err != nil {
+		return nil, err
+	}
+	return NewClientWithConfig(config), nil
 }
 
 // NewClientWithConfig creates new OpenAI API client for specified config.
@@ -31,10 +34,13 @@ func NewClientWithConfig(config ClientConfig) *Client {
 // NewOrgClient creates new OpenAI API client for specified Organization ID.
 //
 // Deprecated: Please use NewClientWithConfig.
-func NewOrgClient(authToken, org string) *Client {
-	config := DefaultConfig(authToken)
+func NewOrgClient(authToken, org string) (*Client, error) {
+	config, err := DefaultConfig(authToken)
+	if err != nil {
+		return nil, err
+	}
 	config.OrgID = org
-	return NewClientWithConfig(config)
+	return NewClientWithConfig(config), nil
 }
 
 func (c *Client) sendRequest(req *http.Request, v interface{}) error {
--- a/api.go
+++ b/api.go
@@ -45,7 +45,7 @@ func NewOrgClient(authToken, org string) (*Client, error) {
 
 func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 	req.Header.Set("Accept", "application/json; charset=utf-8")
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.ApiKey))
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.APIKey))
 
 	// Check whether Content-Type is already set, Upload Files API requires
 	// Content-Type == multipart/form-data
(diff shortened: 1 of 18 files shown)


api.go
```diff
@@ -31,15 +34,18 @@ func NewClientWithConfig(config ClientConfig) *Client {
...
+	}
 	config.OrgID = org
-	return NewClientWithConfig(config)
+	return NewClientWithConfig(config), nil
 }
 
 func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 	req.Header.Set("Accept", "application/json; charset=utf-8")
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.authToken))
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.APIKey))
```
tao12345666333, line 48: It also needs to support api-key for authentication

tao12345666333 (commented):
LGTM, there is one authorization that needs to be revised.  Also the /chat/completions API does not exist for the current stable API version `api-version`

api.go
ttys3: Oops, I forget here. I only changed the one in `newStreamRequest` .

ttys3:
> LGTM, there is one authorization that needs to be revised. Also the /chat/completions API does not exist for the current stable API version `api-version`

authorization for non stream request just fixed.

about the /chat/completions API, we are using the `preview` version. I did not aware that -_-

what can we do about the `stable` version?  add a check and just return an not supported error ?

tao12345666333:
stable API version 2022-12-01 
https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/stable/2022-12-01/inference.json

This API includes a prompt field, if we want to support it, we can just deal with this field.

Or we can just declare we only support 2023-03-15-preview api version

ttys3:
currently, I follow the `openai` python package's logic just use the api type to detect the auth method like:

```go
if c.config.APIType == APITypeAzure {
// use api-key method
} else {
// user bear token
}
```

I think maybe it is better to add a `AuthMethod` option to the config ?

like
```go
type AuthMethod interface {
	SetupAuth(req *http.Request, apiKey string)
}
```

then impl two common :

```go
type GenericAuth struct {
	KeyName string
}

func (a *GenericAuth) SetupAuth(req *http.Request, apiKey string) {
	req.Header.Set(a.KeyName, apiKey)
}

type BearTokenAuth struct{}

func (a *BearTokenAuth) SetupAuth(req *http.Request, apiKey string) {
	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", apiKey))
}
```

and we can have the default:

```
var AzureAPIKeyAuth = &GenericAuth{KeyName: "api-key"}

var OpenAITokenAuth = &BearTokenAuth{}
```

sashabaranov:
Thank you so much for the PR!

sashabaranov:
A couple of general comments:

> 1. add `WithXXX` options, allow construct a config like  `openai.NewConfig(...Option)`

We are actively avoiding this style of configuration as discussed in https://github.com/sashabaranov/go-openai/pull/79.

> 2. make the config field naming more like the `openai` python pkg's, this will make us follow py's changes more easy

While making this package close to python's, it is less of a priority than not breaking API for existing clients. I don't think renaming `Config` fields and `NewConfig` with a returned error are justified given the API non-breakage goal.

ttys3:
> 



> A couple of general comments:
> 
> > 1. add `WithXXX` options, allow construct a config like  `openai.NewConfig(...Option)`
> 
> We are actively avoiding this style of configuration as discussed in #79.
> 
> > 2. make the config field naming more like the `openai` python pkg's, this will make us follow py's changes more easy
> 
> While making this package close to python's would be a good thing, it is less of a priority than not breaking API for existing clients. I don't think renaming `Config` fields and `NewConfig` with a returned error is justified given the API non-breakage goal.

1. it's just a kind of  common practice, many project use this style. I can accept that, if you don't like it.

2. if we do not check the error in the Config constructor, then it seems inappropriate to check for configuration errors elsewhere.  we just ignore the error checking， Leave the blame to the server and have the user check for errors returned by the server ？

ttys3:
just add error check for `CreateChatCompletionStream`,

seems there's other thing need fix:

```
=== RUN   TestCompletionsStreamWrongModel
    stream_test.go:29: CreateCompletion should return ErrCompletionUnsupportedModel, but returned: Post "http://localhost/v1/completions": dial tcp [::1]:80: connect: connection refused
--- FAIL: TestCompletionsStreamWrongModel (0.00s)
```

sashabaranov:
> 1. it's just a kind of common practice, many project use this style. I can accept that, if you don't like it.

Exactly, there are [multiple](https://asankov.dev/blog/2022/01/29/different-ways-to-initialize-go-structs/#variadic-function-constructor) common ways to pass the configuration, this library prefers using a single config struct. 

> 2. if we do not check the error in the Config constructor, then it seems inappropriate to check for configuration errors elsewhere. we may need just ignore the error checking， Leave the blame to the server and have the user check for errors returned by the server ？

This PR checks for three errors:

- **Empty API key** — I think it's fine not to check this during the configuration
- **Whether API type is supported** – just type checking should be enough here if we use iota-enum for APIType
- **Empty APIVersion** — it might make sense to have a `DefaultAzureConfig(...)` that ensures all the nitty-gritty details are specified

```go
if cfg.APIKey == "" {
	return ClientConfig{}, fmt.Errorf("api key is required")
}

if _, ok := supportedAPIType[cfg.APIType]; !ok {
	return ClientConfig{}, fmt.Errorf("unsupported API type %s", cfg.APIType)
}

if cfg.APIType == APITypeAzure || cfg.APIType == APITypeAzureAD {
	if cfg.APIVersion == "" {
		return ClientConfig{}, fmt.Errorf("an API version is required for the Azure API type")
	}
}
```

ttys3:
OK.
I updated commits. and it is a mininal patch now. it is non-breaking.

try to fix the `CreateChatCompletionStream` not detect http error code issue, but which cause `ErrorAccumulatorWriteErrors` tests fail.

I think it's better deal this with other PR.

PTAL

spatecon (approved):

config.go
```diff
@@ -5,28 +5,44 @@ import (
...
+	Engine     string // required for APITypeAzure or APITypeAzureAD
 
-	BaseURL string
-	OrgID   string
+	HTTPClient *http.Client
 
 	EmptyMessagesLimit uint
 }
 
 func DefaultConfig(authToken string) ClientConfig {
```
sashabaranov, line 38: Do we want to have a `DefaultAzure` config (non-AD) so it would fill the config with the necessary fields?

config.go
```diff
@@ -5,28 +5,44 @@ import (
...
+const (
+	APITypeOpenAI  APIType = "OPEN_AI"
+	APITypeAzure   APIType = "AZURE"
+	APITypeAzureAD APIType = "AZURE_AD"
 )
 
 // ClientConfig is a configuration of a client.
 type ClientConfig struct {
 	authToken string
+	BaseURL   string
```
sashabaranov, line 26: Do we need to re-order fields here? The logic was to separate private `authToken` from others

config.go
```diff
@@ -5,28 +5,44 @@ import (
...
 	defaultEmptyMessagesLimit uint = 300
+
+	azureAPIPrefix         = "openai"
+	azureDeploymentsPrefix = "deployments"
+)
+
+type APIType string
+
+const (
+	APITypeOpenAI  APIType = "OPEN_AI"
```
sashabaranov, line 18: Just as a suggestion:

```go
type APIType uint8
const (
	APITypeOpenAI  APIType = iota
	APITypeAzure 
	APITypeAzureAD
)
```

sashabaranov (commented):
Thank you for re-working this PR, it is much more manageable now! I've added a couple of comments — I wonder what your thoughts are

config.go
```diff
@@ -5,28 +5,44 @@ import (
...
 
 // ClientConfig is a configuration of a client.
 type ClientConfig struct {
 	authToken string
+	BaseURL   string
+	OrgID     string
 
-	HTTPClient *http.Client
+	APIType    APIType
+	APIVersion string // required for APITypeAzure or APITypeAzureAD
```
sashabaranov, line 30: Maybe we can also pack Azure-related fields into a sub-structure inside the config? That way, it would be more straightforward during configuration:

```go
config.APIVersion = ...
config.Engine = ...
```
vs

```go
config.Azure.APIVersion = ...
config.Azure.Engine = ...
```

config.go
ttys3: I can add that.

config.go
ttys3: If we do this, do we have to make a special openai config for openai?

ttys3: https://github.com/openai/openai-python/blob/main/openai/__init__.py#L39

ttys3: the py default the version to `2023-03-15-preview`, we do this the same ?

```
api_version = (
    "2023-03-15-preview" if api_type in ("azure", "azure_ad", "azuread") else None
)
```

sashabaranov:
Excellent, ready to merge once codecov issues are fixed 🙌🏻

config.go
```diff
@@ -5,28 +5,61 @@ import (
...
+		OrgID:     "",
+
 		HTTPClient: &http.Client{},
-		BaseURL:    apiURLv1,
+
+		EmptyMessagesLimit: defaultEmptyMessagesLimit,
+	}
+}
+
+func DefaultAzure(apiKey, baseURL, engine string) ClientConfig {
```
sashabaranov, line 53:
```suggestion
func DefaultAzureConfig(apiKey, baseURL, engine string) ClientConfig {
```

sashabaranov (approved):
Great work, thank you!

PR #215, opened 2023-03-31
add ChatCompletionStream Usage return

nasa1024:
add ChatCompletionStream Usage return

### DIFF
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -21,6 +21,7 @@ type ChatCompletionStreamResponse struct {
 	Created int64                        `json:"created"`
 	Model   string                       `json:"model"`
 	Choices []ChatCompletionStreamChoice `json:"choices"`
+	Usage   Usage                        `json:"usage"`
 }
 
 // ChatCompletionStream


sashabaranov (approved):
Thank you for the PR!

collinvandyck:
Thank you this is what I was looking for!

collinvandyck:
I'm not actually seeing any usage data come back in the streaming response. I added a roundtripper to log request/response data and it looks like this (formatting mine):

```sh
POST https://api.openai.com/v1/chat/completions
HTTP/2.0 200 OK
data: {
  "id": "chatcmpl-70smbYYzwcwpb19BkzLb3NFlvkIsP",
  "object": "chat.completion.chunk",
  "created": 1680444941,
  "model": "gpt-3.5-turbo-0301",
  "choices": [
    {
      "delta": {
        "role": "assistant"
      },
      "index": 0,
      "finish_reason": null
    }
  ]
}
data: {
  "id": "chatcmpl-70smbYYzwcwpb19BkzLb3NFlvkIsP",
  "object": "chat.completion.chunk",
  "created": 1680444941,
  "model": "gpt-3.5-turbo-0301",
  "choices": [
    {
      "delta": {
        "content": "How"
      },
      "index": 0,
      "finish_reason": null
    }
  ]
}
```

Is there something I need to do to signal to openai that usage data should be returned?

sashabaranov:
@collinvandyck please upgrade to 1.5.8 `go get -u github.com/sashabaranov/go-openai@v1.5.8`

collinvandyck:
Hey @sashabaranov ! I'm seeing this behavior on 1.5.8, go.mod:

```go
github.com/sashabaranov/go-openai v1.5.8-0.20230401160622-b542086cbb22
```

After giving it a spin I saw that usage was always blank in the resp returned by `Recv`, so that's why I added the request/response logging to verify it wasn't actually coming back in the http response.

I'm kicking off the chat completion stream like this:

```go
req := openai.ChatCompletionRequest{
    Model:    c.model,
    Messages: messages,
    Stream:   true,
}
resp, err := c.openai.CreateChatCompletionStream(ctx, req)
if err != nil {
    return nil, fmt.Errorf("stream: %w", err)
}
```

collinvandyck:
Aha, I found this:

https://github.com/openai/openai-cookbook/blob/main/examples/How_to_stream_completions.ipynb

image

nasa1024:
too embarrassed
@collinvandyck yes, `/chat/completions` when using `stream=True` will not return Usage, i will fix it

collinvandyck:
It would be neat for go-openai to include a tokenizer to do this, but it seems like a pretty big lift. It looks like most of the folks that are doing this are calling out to a rust tokenizer since there's not an official go tokenizer. Not suggesting that you should do this, but it was surprising to me that there wasn't any official way to do this using openai's libraries in Go.

nasa1024:
> It would be neat for go-openai to include a tokenizer to do this, but it seems like a pretty big lift. It looks like most of the folks that are doing this are calling out to a rust tokenizer since there's not an official go tokenizer. Not suggesting that you should do this, but it was surprising to me that there wasn't any official way to do this using openai's libraries in Go.

sir, see [223](https://github.com/sashabaranov/go-openai/pull/223)

Issue #201, opened 2023-03-26
support batching


go-dockly:
It would be nice if the chatCompletion method could accept an array of requests optionally as described below:

[Batching requests](https://platform.openai.com/docs/guides/rate-limits/batching-requests)
The OpenAI API has separate limits for requests per minute and tokens per minute. 

If you're hitting the limit on requests per minute, but have available capacity on tokens per minute, you can increase your throughput by batching multiple tasks into each request. This will allow you to process more tokens per minute, especially with our smaller models.

Sending in a batch of prompts works exactly the same as a normal API call, except you pass in a list of strings to the prompt parameter instead of a single string.

sashabaranov:
Done in #220

PR #220, opened 2023-04-02
CompletionBatchingRequestSupport

NullpointerW:
solve iusse #201   i think this feature is useful
i closed  pr #210   because  I think there are too many unnecessary changes, here is a more concise change to implement
```go
c := openai.NewClient("your token")
	ctx := context.Background()

	
	// test CreateCompletion
	reqNor := CompletionRequest{
		Prompt:    []string{"Lorem ipsum", "Lorem ipsum", "Lorem ipsum"}, //use string or []string 
		Model:     GPT3Curie,
		MaxTokens: 5,
	}
	resp, err := c.CreateCompletion(ctx, reqNor)
	checks.NoError(t, err, "CreateCompletion returned error")
	if err == nil {
		for _, c := range resp.Choices {
			t.Logf("index =`%d`,text is `%s` \n", c.Index, c.Text)
		}
	}
```
`Prompt`   can be assigned `string` or `[]string` type
Streaming is also supported
```go
c := openai.NewClient("your token")
	ctx := context.Background()

	// test CreateCompletionStream
	streq := CompletionRequest{
		Model:     GPT3Curie,
		MaxTokens: 5,
		Prompt:   []string{ "a dog entered the alley", "a dog entered the alley", "a dog entered the alley"},
		Stream:    true,
	}
	stream, err := c.CreateCompletionStream(ctx, streq)
	if err != nil {
		t.Logf("CompletionStream error: %v\n", err)
		return
	}
	
	defer stream.Close()
	type index = int
	choicesMap := make(map[index]string)
	for {
		response, streamErr := stream.Recv()
		if errors.Is(streamErr, io.EOF) {
			t.Logf("CompletionBatch Stream finished")
			break
		}

		if streamErr != nil {
			checks.NoError(t, streamErr, "CompletionBatchStream error:Stream error")
		}
		t.Logf("Stream response: %#+v\n", response.Choices)
		choicesMap[response.Choices[0].Index] += response.Choices[0].Text
	}
	for k, v := range choicesMap {
		t.Logf("choice_index:%d,text:%s", k, v)
	}
```

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -78,6 +78,17 @@ func checkEndpointSupportsModel(endpoint, model string) bool {
 	return !disabledModelsForEndpoints[endpoint][model]
 }
 
+func checkPromptType(prompt any) (ok bool) {
+	if prompt != nil {
+		switch prompt.(type) {
+		case string, []string:
+		default:
+			return
+		}
+	}
+	return true
+}
+
 // CompletionRequest represents a request structure for completion API.
 type CompletionRequest struct {
 	Model            string         `json:"model"`
@@ -144,9 +155,7 @@ func (c *Client) CreateCompletion(
 		return
 	}
 
-	switch request.Prompt.(type) {
-	case string, []string:
-	default:
+	if !checkPromptType(request.Prompt) {
 		err = ErrCompletionRequestPromptTypeNotSupported
 		return
 	}
--- a/completion.go
+++ b/completion.go
@@ -78,15 +78,10 @@ func checkEndpointSupportsModel(endpoint, model string) bool {
 	return !disabledModelsForEndpoints[endpoint][model]
 }
 
-func checkPromptType(prompt any) (ok bool) {
-	if prompt != nil {
-		switch prompt.(type) {
-		case string, []string:
-		default:
-			return
-		}
-	}
-	return true
+func checkPromptType(prompt any) bool {
+	_, isString := prompt.(string)
+	_, isStringSlice := prompt.([]string)
+	return prompt == nil || isString || isStringSlice
 }
 
 // CompletionRequest represents a request structure for completion API.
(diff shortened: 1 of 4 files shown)


completion.go
```diff
@@ -77,10 +78,21 @@ func checkEndpointSupportsModel(endpoint, model string) bool {
 	return !disabledModelsForEndpoints[endpoint][model]
 }
 
+func checkPromptType(prompt any) (ok bool) {
```
sashabaranov, line 81: Sorry, I couldn't help myself not to do some code golfing here :D  https://go.dev/play/p/-2aR0tII8xG

```go
func checkPromptType(prompt any) bool {
	_, isString := prompt.(string)
	_, isStringSlice := prompt.([]string)
	return isString || isStringSlice
}
```

sashabaranov (approved):
Thank you so much for the PR! That was a long-standing issue 🙌🏻

completion.go
sashabaranov: @NullpointerW I've approved the PR, you can either leave your implementation or add this one — up to you!

completion.go
```diff
@@ -77,10 +78,16 @@ func checkEndpointSupportsModel(endpoint, model string) bool {
 	return !disabledModelsForEndpoints[endpoint][model]
 }
 
+func checkPromptType(prompt any) bool {
+	_, isString := prompt.(string)
+	_, isStringSlice := prompt.([]string)
+	return prompt == nil || isString || isStringSlice
```
sashabaranov, line 84: No need to check for `nil` here, see example https://go.dev/play/p/-2aR0tII8xG

sashabaranov: Also, we are trying to avoid `nil` prompts, right?

PR #221, opened 2023-04-02
function NewClient supports proxy

SSRemex:
the NewClient supports the HTTP proxy now 
the second parameter is optional
```go
c := openai.NewClient(openaiKey, proxyAddr)
```


sashabaranov:
@SSRemex thank you for the PR! This functionality is too specific to go into the `NewClient` method. Proxy configuration is documented in the README.

Issue #226, opened 2023-04-05
add set proxy in openai init


SpenserCai:
add set proxy in openai init

sashabaranov:
The option to set a proxy is available now and documented in the README, please see also https://github.com/sashabaranov/go-openai/pull/112 and https://github.com/sashabaranov/go-openai/pull/221

PR #227, opened 2023-04-05
Add set proxy feature (#226)

SpenserCai:
Add set proxy feature #226.


sashabaranov:
Thank you for the PR, please see #112 and #221

PR #230, opened 2023-04-06
Add ImageEditRequest/ImageVariRequest.ResponseFormat

aceld:


sashabaranov:
@aceld Hey, thank you for the PR! 

To fix the coverage issue, it might make sense to do something akin to [requestBuilder](https://github.com/sashabaranov/go-openai/blob/master/request_builder.go) interface. You could create something like:

```go
type formBuilder interface {
	createFormFile(fieldname, filename string) (io.Writer, error)
	writeField(fieldname, value string) error
}
```

and add it to the `Client` like this:

```go
type Client struct {
	config ClientConfig

	requestBuilder requestBuilder
	formBuilder formBuilder // <--- new
}
```

And have the default implementation for that interface with the `mime/multipart` `Writer`.  


Then, in tests, you'll be able to mock it (see failing [requestBuilder](https://github.com/sashabaranov/go-openai/blob/master/request_builder_test.go#L26) for example). That would enable you to cover all these `if err != nil { return }` lines in `image.go`:

```go
err = writer.WriteField("prompt", request.Prompt)
if err != nil {
	return
}
err = writer.WriteField("n", strconv.Itoa(request.N))
if err != nil {
	return
}
err = writer.WriteField("size", request.Size)
if err != nil {
	return
}
```

aceld:
@sashabaranov  It's just adding a request parameter, why is it so troublesome? The problem I'm encountering now is that the Base64 parameter of ImageEditRequest cannot be passed to OpenAI. If it's possible for you to add the page, thank you.

sashabaranov:
@aceId I understand the frustration! :D The thing is that we have a quality standard of code coverage, and we require it not to go down. You are working with the historically least test-covered file https://app.codecov.io/gh/sashabaranov/go-openai/blob/master/image.go

You can either cover it yourself, or we can merge into a non-master branch, and I'll take it from there.

Not sure I understand the problem with the Base64 parameter

sashabaranov:
Started working on test coverage in #235

sashabaranov:
@aceld I've added all the tests, I think this PR should be safely mergeable now!

aceld:
OK， Thanks

PR #234, opened 2023-04-08
 It is possible to customize the settings for both AuthToken and BaseURL at the same time.

shengbox:
Please allow me to customize AuthToken and BaseURL at the same time

### DIFF
--- a/api.go
+++ b/api.go
@@ -39,7 +39,7 @@ func NewOrgClient(authToken, org string) *Client {
 
 func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 	req.Header.Set("Accept", "application/json; charset=utf-8")
-	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.authToken))
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.AuthToken))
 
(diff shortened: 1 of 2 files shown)


sashabaranov:
@shengbox Hey, doesn't `DefaultAzureConfig(apiKey, baseURL, engine string)` fit your case?

https://pkg.go.dev/github.com/sashabaranov/go-openai#DefaultAzureConfig

shengbox:
Due to the "Great Firewall" in China, Chinese users typically use a proxy server deployed on a VPS to access OpenAI.

sashabaranov:
@shengbox so why doesn't `DefaultAzureConfig(apiKey, baseURL, engine string)` work for you? It sets both the `authToken` and `baseURL`

shengbox:
@sashabaranov When I use DefaultAzureConfig, the APIType is set to AZURE, but actually what I need is APIType=OPEN_AI.

sashabaranov:
@shengbox got it, what if you do something like:

```go
// set authToken and BaseURL
config := openai.DefaultConfig("your token")
config.BaseURL = "your base URL"

// create client
```

shengbox:
Yes, that should work just fine.

Issue #244, opened 2023-04-10
APIError.error.code type error!!!


crazyworld:
error, json: cannot unmarshal number into Go struct field APIError.error.code of type string

crazyworld:
The error occurred in this client.CreateChatCompletion()

sashabaranov:
cc @liushuangls

liushuangls:
> cc @liushuangls

Change the code type to any？

PR #247, opened 2023-04-13
Resolve json Unmarshal error when API returns an int instead of a string

young-steveo:
This resolves https://github.com/sashabaranov/go-openai/issues/244 .

In the case of an invalid API key, the `APIError.Code` is a string pointer, but `APIError.Code` can be an integer for other API errors, such as when the API quota has been exceeded.

By changing the `APIError.Code` type to [`json.RawMessage`](https://pkg.go.dev/encoding/json#RawMessage), we can delay parsing the value until we need it (in most cases, we probably don't need it, since the struct also contains a StatusCode and error message, in which case `APIError.Code` can be ignored).

I've updated the invalid API Key test to parse the string `"invalid_api_key"` as it did before, using a new method, `APIError.CodeAsStringPtr()`, but I was unable to think of a good test for the quota error that returns an integer. Despite that, I still provided a method to decode the `APIError.Code` as an int if anyone needs it (`APIError.CodeAsInt()`)

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -143,6 +143,51 @@ func TestAPIError(t *testing.T) {
 	}
 }
 
+func TestAPIErrorCode(t *testing.T) {
+	// test integer code
+	response := `{"code":418,"message":"I'm a teapot","param":"prompt","type":"teapot_error"}`
+	var apiErr APIError
+	err := json.Unmarshal([]byte(response), &apiErr)
+	checks.NoError(t, err, "Unexpected Unmarshal API response error")
+
+	code, err := apiErr.Code()
+	checks.NoError(t, err, "Unexpected API code error")
+
+	switch v := code.(type) {
+	case int:
+		if v != 418 {
+			t.Fatalf("Unexpected API code integer: %d; expected 418", v)
+		}
+	default:
+		t.Fatalf("Unexpected API error code type: %T", v)
+	}
+
+	// test string code
+	response = `{"code":"teapot","message":"I'm a teapot","param":"prompt","type":"teapot_error"}`
+	err = json.Unmarshal([]byte(response), &apiErr)
+	checks.NoError(t, err, "Unexpected Unmarshal API response error")
+
+	code, err = apiErr.Code()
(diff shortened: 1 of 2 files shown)


sashabaranov:
@young-steveo thank you for the PR! A couple of suggestions:

1. Should we make the `Code` field private?
2. What if we substitute `CodeAsStringPtr` and `CodeAsInt` with a single method `Code() (any, error)`? That would enable users to use idiomatic Go type [switches](https://go.dev/tour/methods/16) to figure out the type, e.g.

```go
code, err := apiErr.Code()
if err != nil {
	// ... handle the error ...
}
switch v := code.(type) {
case int:
	// ... handle integer code
case string:
	// ... handle string code
}
```

Maybe we can go even further and ignore the error here, so `apiErr.Code()` would just return nil in case Unmarshal fails.

young-steveo:
@sashabaranov Unfortunately, the`json` package requires tagged fields to be exported, so changing it from `Code` to `code` prevents the`json` package from unmarshalling the data (the `json` package uses `reflect` to inspect struct fields, and `reflect` doesn't allow access to unexported fields).

What I can do as a workaround is maybe change the field to a single letter, like `C`, and the end user can use `Code() (any, error)` as a convenience, even though they technically have access to the `json.RawMessage` via `C`.

I'll work that up for this PR and see how it looks.

sashabaranov:
@young-steveo That's right, I've completely forgotten about making fields public for json!

We may utilize custom json unmarshalling with `UnmarshalJSON` here. Here are a couple of possible implementations: https://go.dev/play/p/IOuvLovvnUV What do you think?

young-steveo:
@sashabaranov  For some reason, I never thought about implementing `UnmarshalJSON`. I like that approach! I'll open a separate PR with that solution; if you merge that one, we can close this one, or vice-versa.

young-steveo:
Opened a new PR https://github.com/sashabaranov/go-openai/pull/248. Left this one open for now until you decide which is the better implementation.

BTW, love this project.

PR #248, opened 2023-04-14
Implement Unmarshaller interface. Resolves #244

### LINKED ISSUE #244: APIError.error.code type error!!!

young-steveo:
Based on [this comment](https://github.com/sashabaranov/go-openai/pull/247#issuecomment-1507658375), this PR implements the custom JSON Unmarshaller interface to handle numbers, strings, or any other type that might be returned by the API.

Also fixed a bug where the skipped production test should EXPECT an error, rather than the opposite.

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,10 +1,13 @@
 package openai
 
-import "fmt"
+import (
+	"encoding/json"
+	"fmt"
+)
 
 // APIError provides error information returned by the OpenAI API.
 type APIError struct {
-	Code       *string `json:"code,omitempty"`
+	Code       any     `json:"code,omitempty"`
 	Message    string  `json:"message"`
 	Param      *string `json:"param,omitempty"`
 	Type       string  `json:"type"`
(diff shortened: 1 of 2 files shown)


sashabaranov:
Great stuff, thank you so much! Seems like we can fix all the problems with minimal API changes!

error.go
```diff
@@ -25,6 +28,47 @@ func (e *APIError) Error() string {
 	return e.Message
 }
 
+func (e *APIError) UnmarshalJSON(data []byte) (err error) {
+	var rawMap map[string]json.RawMessage
+	err = json.Unmarshal(data, &rawMap)
+	if err != nil {
+		return
```
sashabaranov, line 35: Maybe the simplest way to cover this would be to pass some invalid data. We also have testable jsonMarshaller [here](https://github.com/sashabaranov/go-openai/blob/master/request_builder_test.go#L22), not sure if it would be possible to accommodate interface-based jsonUnmarshaller here.

young-steveo: I added a test for this block, but I could not pass the invalid data through `json.Unmarshal` because the standard library [filters out malformed data](https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/encoding/json/decode.go;l=102) before sending it to the unmarshaller. I ended up just calling the method directly in the test.

sashabaranov (approved):

Imebeez (commented):
Done

PR #255, opened 2023-04-17
Add GetHttpResponse method for streamReader, fix #250

### LINKED ISSUE #250: Bug:  The `CreateChatCompletionStream` do not  return ` Rate limit ` and  ` Quota exceed `  error correctly.
The `CreateChatCompletionStream` do not return ` Rate limit` and `Quota exceed` error correctly, which is important for user to retry their requests.
```
stream, err := client.CreateChatCompletionStream(
		context.Background(),
		openai.ChatCompletionRequest{
			Model:    model,
			Messages: messages,
			Stream:   true,
		},
	)
```

I always get the `err` with `nil` value when actually meet a ` Rate limit` and `Quota exceed` issue, ` client.CreateChatCompletion` works well.

guoqiangqi:
Add GetHttpResponse method for streamReader to catch http status ouside.

### DIFF
--- a/stream_reader.go
+++ b/stream_reader.go
@@ -22,6 +22,10 @@ type streamReader[T streamable] struct {
 	unmarshaler    unmarshaler
 }
 
+func (stream *streamReader[T]) GetHttpResponse() (*http.Response, error) {
+	return stream.response, nil
+}
+
 func (stream *streamReader[T]) Recv() (response T, err error) {
 	if stream.isFinished {
 		err = io.EOF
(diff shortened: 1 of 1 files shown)


guoqiangqi:
HI @sashabaranov ,  Since I'm not very clear about testing for this project and this PR simply adds a method so that I can check the response status when getting HTTP stream instead of directly using `Recv` . Could you help merge this ?

sashabaranov:
@guoqiangqi Hey, thank you for the PR! Why exactly do you need this method?

sashabaranov:
Got it! It seems that a better approach here would be to improve error reporting on stream creation or `.Recv()`. I don't think we should merge the `GetHttpResponse` method as it leaks stream abstraction.

PR #256, opened 2023-04-17
Feat: support dynamically set client config 

shengzhou1216:
support update client config. e.g. auth token, org id ...

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"sync"
 )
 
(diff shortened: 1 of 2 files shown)


sashabaranov:
Hey, thank you for the PR! Could you please elaborate on the motivation for this dynamic config? Why doesn't re-creating client objects with new configs doesn't work for you?

shengzhou1216:
Yup, my bad for not carefully reviewing the source code. Re-create client object with new config is  fine.

PR #259, opened 2023-04-17
fix: chat stream resp error

### LINKED ISSUE #250: Bug:  The `CreateChatCompletionStream` do not  return ` Rate limit ` and  ` Quota exceed `  error correctly.

liushuangls:
fix #250

### DIFF
--- a/client.go
+++ b/client.go
@@ -72,17 +72,7 @@ func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 	defer res.Body.Close()
 
 	if res.StatusCode < http.StatusOK || res.StatusCode >= http.StatusBadRequest {
-		var errRes ErrorResponse
-		err = json.NewDecoder(res.Body).Decode(&errRes)
-		if err != nil || errRes.Error == nil {
-			reqErr := RequestError{
-				StatusCode: res.StatusCode,
-				Err:        err,
-			}
-			return fmt.Errorf("error, %w", &reqErr)
-		}
-		errRes.Error.StatusCode = res.StatusCode
-		return fmt.Errorf("error, status code: %d, message: %w", res.StatusCode, errRes.Error)
+		return c.handleErrorResp(res)
 	}
 
 	if v != nil {
@@ -131,4 +121,19 @@ func (c *Client) newStreamRequest(
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.authToken))
 	}
 	return req, nil
+}
+
+func (c *Client) handleErrorResp(resp *http.Response) error {
+	var errRes ErrorResponse
+	err := json.NewDecoder(resp.Body).Decode(&errRes)
+	fmt.Println(err)
+	if err != nil || errRes.Error == nil {
+		reqErr := RequestError{
+			StatusCode: resp.StatusCode,
+			Err:        err,
+		}
+		return fmt.Errorf("error, %w", &reqErr)
+	}
+	errRes.Error.StatusCode = resp.StatusCode
+	return fmt.Errorf("error, status code: %d, message: %w", resp.StatusCode, errRes.Error)
 }
--- a/client.go
+++ b/client.go
@@ -128,11 +128,11 @@ func (c *Client) handleErrorResp(resp *http.Response) error {
 	err := json.NewDecoder(resp.Body).Decode(&errRes)
 	if err != nil || errRes.Error == nil {
 		reqErr := RequestError{
-			StatusCode: resp.StatusCode,
-			Err:        err,
+			HTTPStatusCode: resp.StatusCode,
+			Err:            err,
 		}
 		return fmt.Errorf("error, %w", &reqErr)
 	}
-	errRes.Error.StatusCode = resp.StatusCode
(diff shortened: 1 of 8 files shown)


client.go
```diff
@@ -132,3 +122,17 @@ func (c *Client) newStreamRequest(
...
+	var errRes ErrorResponse
+	err := json.NewDecoder(resp.Body).Decode(&errRes)
+	if err != nil || errRes.Error == nil {
+		reqErr := RequestError{
+			StatusCode: resp.StatusCode,
+			Err:        err,
+		}
+		return fmt.Errorf("error, %w", &reqErr)
+	}
+	errRes.Error.StatusCode = resp.StatusCode
```
sashabaranov, line 136: Doesn't the response body JSON contain the status code information? (see your example with `dataBytes` in the `stream_test.go`). So maybe let's keep the original JSON value of `errRes.Error.StatusCode` and pass `resp.StatusCode` in the `fmt.Errorf` as it is done now?

sashabaranov:
Great work, thank you so much!

client.go
liushuangls: > Doesn't the response body JSON contain the status code information? (see your example with `dataBytes` in the `stream_test.go`). So maybe let's keep the original JSON value of `errRes.Error.StatusCode` and pass `resp.StatusCode` in the `fmt.Errorf` as it is done now?

I see response body JSON contain code but not status code?

guoqiangqi:
@liushuangls So excited to see the patch, thanks for fixing.

client.go
liushuangls: First we are unmarshalling into the ErrorReponse struct, when it fails we return RequestError

liushuangls: So when unmarshalling into the ErrorReponse success, need set StatusCode

sashabaranov: In case of success, we can use `errRes.StatusCode` to store the HTTP response status code. And the `errRes.Error.StatusCode` will be used for whatever OpenAI API returns. Maybe I misunderstand something — please correct me if I'm wrong

liushuangls: I used the error handling of the current code：
https://github.com/sashabaranov/go-openai/blob/3b10c032b677b468217a50f42152d84c03ca942a/client.go#L75

I guess `errRes.Error.StatusCode` was designed to store response status code

sashabaranov: 🤔🤔🤔 there was a PR to support `any` type for StatusCode, so it seems there is more to it than the HTTP response code see https://github.com/sashabaranov/go-openai/pull/248

CC @young-steveo

liushuangls: I see #248 support `any` type for Code rather than StatusCode?

liushuangls: `APIError.StatusCode` json tag is `-`, this explains it not read response body?

young-steveo: @sashabaranov It's a little confusing because three fields are currently in the code:

* `APIError.Code`,
* `APIError.StatusCode`, and
* `RequestError.StatusCode`

As @liushuangls mentioned, the `APIError.StatusCode` field tag is [`json:"-"`](https://github.com/sashabaranov/go-openai/blob/3b10c032b677b468217a50f42152d84c03ca942a/error.go#L14) which omits it from the Unmarshalling step, which makes sense because the body of the API response does not have a "status code" key. It looks like:

```
{
  "error": {
    "message": "...",
    "type": "...",
    "param":"...",
    "code": "..."
  }
}
```
...where `code` can contain either a string or a number, depending on the error returned by the API.  e.g. sometimes it will be a string like `"invalid_api_key"`, but in the case of an account with no available quota, `code` was a number (I think it was `429`, mimicking the response header status code).

`APIError` is a struct that represents the JSON "error" structure of the response body, so it perhaps shouldn't have the `StatusCode` field since the `RequestError.StatusCode` represents the same data.  Perhaps `StatusCode` was added to the `APIError` struct as a convenience?

sashabaranov: @young-steveo thank you for breaking this down!

@liushuangls would you be able to remove the `APIError.StatusCode` field in this PR? (it would break a test in `api_test.go`). I can also add a follow-up PR with this fix

liushuangls: One problem is that when response error and unmarshalling into the ErrorReponse success, how does the user get the 
 response status code?

sashabaranov: @liushuangls ah, I see your point! I propose we rename `APIError.StatusCode` to `APIError.HTTPStatusCode` to make this field's purpose more obvious

liushuangls: > @liushuangls ah, I see your point! I propose we rename `APIError.StatusCode` to `APIError.HTTPStatusCode` to make this field's purpose more obvious

good idea

sashabaranov (approved):

PR #260, opened 2023-04-18
add ImageVariRequest/ImageEditRequest.ResponseFormat

aceld:


sashabaranov:
@aceId Thank you for the PR! Could you please add tests for that here https://github.com/sashabaranov/go-openai/blob/master/image_test.go#L335

Issue #143, opened 2023-03-10
Whisper Prompt and Response Format


jkelly-multivac:
Hey there,
Was looking at the CreateTranscription and noticed there isn't an option to set prompt or response_format. I'll attempt to add locally but wanted to check to see if I'm just missing where that option might already be able to be set. I don't see it conditionally in the form part creation but thought I would check.

jkelly-multivac:
I have this working locally but had to add an alternative sendRequest to api.go also to allow for raw (non JSON) response. Probably not in a good PR state. I feel a little hacky using a different request chain if the response format isn't JSON.

hoani:
This is being addressed here https://github.com/sashabaranov/go-openai/pull/267 - probably quite similar to what you have implemented... but I think this is probably ok given that the API provides a response that is intended to be written directly to file.

PR #267, opened 2023-04-19
Adds support for audio captioning with Whisper

hoani:
Closes https://github.com/sashabaranov/go-openai/issues/143

Allows users to add a `Format: ` argument to the request to get captions from the audio apis.

Demonstration (using the same code added for README)

https://user-images.githubusercontent.com/4222666/232965446-78a47bc5-d7eb-4439-a161-3ff13f2600ff.mov

Broke `decodeResponse` out of client, because cyclomatic complexity  got above 20 with the string response changes.

Added tests for `decodeResponse` to not decrease code coverage

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,47 @@ func main() {
 </details>
 
 <details>
+<summary>Audio Captions</summary>
+
+```go
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	openai "github.com/sashabaranov/go-openai"
+)
+
+func main() {
+	c := openai.NewClient(os.Getenv("OPENAI_KEY"))
+
+	req := openai.AudioRequest{
+		Model:    openai.Whisper1,
+		FilePath: os.Args[1],
+		Format:   openai.AudioResponseFormatSRT,
+	}
+	resp, err := c.CreateTranscription(context.Background(), req)
+	if err != nil {
+		fmt.Printf("Transcription error: %v\n", err)
+		return
+	}
+	f, err := os.Create(os.Args[1] + ".srt")
+	if err != nil {
+		fmt.Printf("Could not open file: %v\n", err)
+		return
+	}
+	defer f.Close()
+	if _, err := f.WriteString(resp.Text); err != nil {
+		fmt.Printf("Error writing to file: %v\n", err)
+		return
+	}
+}
+```
+</details>
+
+<details>
 <summary>DALL-E 2 image generation</summary>
 
 ```go
--- a/README.md
+++ b/README.md
@@ -134,3 +134,34 @@ func main() {
 }
 ```
 </details>
+
+<details>
+<summary>Audio Speech-To-Text</summary>
+
+```go
+package main
+
+import (
+	"context"
+	"fmt"
+
+	openai "github.com/sashabaranov/go-openai"
+)
+
+func main() {
+	c := openai.NewClient("your token")
+	ctx := context.Background()
+
+	req := openai.AudioRequest{
+		Model:    openai.Whisper1,
+		FilePath: "recording.mp3",
+	}
+	resp, err := c.CreateTranscription(ctx, req)
+	if err != nil {
+		fmt.Printf("Transcription error: %v\n", err)
+		return
+	}
+	fmt.Println(resp.Text)
+}
+```
+</details>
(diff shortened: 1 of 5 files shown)


sashabaranov:
Thank you so much for the PR! 🙌🏻

audio.go
```diff
@@ -21,6 +28,7 @@ type AudioRequest struct {
 	Prompt      string // For translation, it should be in English
 	Temperature float32
 	Language    string // For translation, just do not use it. It seems "en" works, not confirmed...
+	Format      string
```
sashabaranov, line 31: Let's maybe make typechecker work for us a little bit here?

```go
type AudioResponseFormat string
const (
	AudioResponseFormatJSON AudioResponseFormat = "json"
	AudioResponseFormatSRT  AudioResponseFormat = "srt"
	AudioResponseFormatVTT  AudioResponseFormat = "vtt"
)

type AudioRequest struct {
...
	Format AudioResponseFormat
}
```

client.go
```diff
@@ -85,13 +85,23 @@ func (c *Client) sendRequest(req *http.Request, v interface{}) error {
 		return fmt.Errorf("error, status code: %d, message: %w", res.StatusCode, errRes.Error)
 	}
 
-	if v != nil {
-		if err = json.NewDecoder(res.Body).Decode(v); err != nil {
+	return decodeResponse(v, *res)
+}
+
+func decodeResponse(v interface{}, res http.Response) error {
```
sashabaranov, line 91:
```suggestion
func decodeResponse(v any, res http.Response) error {
```

sashabaranov: Could we also rename `v` to something sensible here, maybe `response`?

client.go
```diff
@@ -85,13 +85,23 @@ func (c *Client) sendRequest(req *http.Request, v interface{}) error {
...
+	if v == nil {
+		return nil
+	}
+
+	if result, ok := v.(*string); ok {
```
sashabaranov, line 96: The `ok { ... }` block is rather large here, while the json decoding is a one-liner. Maybe invert the logic like:

```go
stringResponsePtr, isStringPtr := v.(*string)
if !isStringPtr {
	return json.NewDecoder(res.Body).Decode(v)
}

// ... rest of the logic 
```

README.md
```diff
@@ -223,6 +223,47 @@ func main() {
 ```
 </details>
 
+<details>
+<summary>Audio Captions</summary>
+
+```go
+package main
```
sashabaranov, line 230: It would be great to add a testable example for this, see https://go.dev/blog/examples and #80

audio.go
hoani: Good idea, done

client.go
hoani: I have:
* Used `any`
* Updated `decodeResponse` arguments to `body io.Reader, v any`
  * This ordering is a little nicer (input/output)
  * I struggled to come up with a good name for `v` - this is what is used in the `json` library to mean any value that we will decode into... so I feel like it is a better name than response which is used by `http` (at least in this context)

client.go
hoani: I agree, but I broke out the string decoding into its own function instead. The main reason being that the string handling is the exception, so should be handled immediately before we drop to the default handling.

This should mean if we have other values in the future which require special handling (e.g. an `os.File` or something similar), we can handle that before using our default `json` handling.

README.md
hoani: I agree, but this isn't trivial for this library where we need to set up a test server to handle these examples.

I've done a few of these before https://hoani.net/guides/software/golang/godocs, and they're pretty cool, but I feel like that should be taken care of in a new PR.

I was also thinking when putting these docs together that it might be time for an `examples` folder, similar to other repos like https://github.com/charmbracelet/bubbletea/tree/master/examples - I'd be happy to put this together as a separate PR since it isn't too much work. What do you think?

sashabaranov: We can do something akin to `api_test.go`:
```go
apiToken := os.Getenv("OPENAI_TOKEN")
if apiToken == "" {
	t.Skip("Skipping testing against production OpenAI API. Set OPENAI_TOKEN environment variable to enable it.")
}
```

But let's separate this to another PR!

sashabaranov (approved):

PR #270, opened 2023-04-20
fix:model param type, add moderation Model Name const.

aceld:


sashabaranov (approved):
Thank you for the PR! I think it makes sense to add separate types for models in the future.

PR #272, opened 2023-04-20
Add role attribute to ChatCompletionStreamChoiceDelta

autopaca:
The first response of the stream api contains the role attribute:
image

I believe this should be in the type definition.

Thank you for the great work sashabaranov!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module github.com/sashabaranov/go-openai
+module github.com/autopaca/go-openai
 
 go 1.18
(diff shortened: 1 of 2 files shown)


sashabaranov:
Thank you for the PR! 🙌🏻

go.mod
```diff
@@ -1,3 +1,3 @@
-module github.com/sashabaranov/go-openai
+module github.com/autopaca/go-openai
```
sashabaranov, line 1:
```suggestion
module github.com/sashabaranov/go-openai
```

Issue #280, opened 2023-04-25
Not as expected err: unexpected end of JSON input


zjy282:
When OpenAI or Azure returns an undefined error, since the JSON parsing meets the expectations of ErrorResponse, the "error, unexpected end of JSON input" error will be returned 
**client.go : line 134 
func:handleErrorResp**
```golang
err := json.NewDecoder(resp.Body).Decode(&errRes) 
```
The real error message cannot be captured , but errRes.Error can get the real error message.
Example Azure origin response body:
```json
{
    "error":{
        "code":"AccessDenied",
        "message":"Access denied due to Virtual Network/Firewall rules."
    }
}
```
Calling “**response, err := client.CreateChatCompletion()**” gets “**error, unexpected end of JSON input**”,but unable to get correct error:Access denied due to Virtual Network/Firewall rules.

I think the correct error message can be obtained by modifying it in the following way:
```golang
func (c *Client) handleErrorResp(resp *http.Response) error {
	var errRes ErrorResponse
	err := json.NewDecoder(resp.Body).Decode(&errRes)
	if err != nil || errRes.Error == nil {
		reqErr := RequestError{
			HTTPStatusCode: resp.StatusCode,
			Err:            err,
		}
		if errRes.Error != nil{
			reqErr.Err = errRes.Error
		}
		return fmt.Errorf("error, %w", &reqErr)
	}
	errRes.Error.HTTPStatusCode = resp.StatusCode
	return fmt.Errorf("error, status code: %d, message: %w", resp.StatusCode, errRes.Error)
}
```

Go Version:1.19.7
go-openai Version:1.9.0

vvatanabe:
Fixed #283

PR #283, opened 2023-04-26
 Fix OpenAI or Azure returns an undefined error message return

zjy282:
Fix: OpenAI or Azure returns an undefined error message return
#280

### DIFF
--- a/client.go
+++ b/client.go
@@ -148,6 +148,9 @@ func (c *Client) handleErrorResp(resp *http.Response) error {
 			HTTPStatusCode: resp.StatusCode,
 			Err:            err,
 		}
+		if errRes.Error != nil {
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -148,6 +148,9 @@ func (c *Client) handleErrorResp(resp *http.Response) error {
 			HTTPStatusCode: resp.StatusCode,
 			Err:            err,
 		}
+		if errRes.Error != nil {
```
sashabaranov, line 151: The if condition above checks for `err != nil || errRes.Error == nil`, so it's possible that `errRes` is nil here and you'll dereference a nil pointer

sashabaranov (approved):

sashabaranov:
Thank you for the PR!

Issue #287, opened 2023-04-30
Azure OpenAI list models


jsburckhardt:
## What 

When using the api to list models `client.ListModels` the end point doesn't use the AzureConfigDefault but the Default one, making the fullURL non existent.

## Expected behaviour

- if the APIType is Azure I would expect to get a list of models available.

## Current behaviour

- receiving an 404 (wrong endpoint and wrong api)

## Support links

- the end point for listing models in Azure OpenAI is [link](https://learn.microsoft.com/en-us/rest/api/cognitiveservices/azureopenaistable/models/list?tabs=HTTP)

vvatanabe:
Fixed #290

PR #290, opened 2023-04-30
Jsburckhardt/287 azure openai list models

jsburckhardt:
## what

- feat(models): include flow for listing models with azure openai endpoint 

## why

- working on issue #287 

## how to test

- `go test`
  image

- or run test `TestAzureListModels`
  image

### DIFF
--- a/internal/test/server.go
+++ b/internal/test/server.go
@@ -31,7 +31,7 @@ func (ts *ServerTest) OpenAITestServer() *httptest.Server {
 		log.Printf("received request at path %q\n", r.URL.Path)
 
 		// check auth
-		if r.Header.Get("Authorization") != "Bearer "+GetTestToken() {
+		if r.Header.Get("Authorization") != "Bearer "+GetTestToken() && r.Header.Get("api-key") != GetTestToken() {
 			w.WriteHeader(http.StatusUnauthorized)
 			return
 		}
--- a/models.go
+++ b/models.go
@@ -40,11 +40,28 @@ type ModelsList struct {
 // ListModels Lists the currently available models,
 // and provides basic information about each model such as the model id and parent.
 func (c *Client) ListModels(ctx context.Context) (models ModelsList, err error) {
-	req, err := c.requestBuilder.build(ctx, http.MethodGet, c.fullURL("/models"), nil)
-	if err != nil {
-		return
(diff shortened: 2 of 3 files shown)


sashabaranov:
@jsburckhardt thank you for the PR! I wonder if this is covered by the logic in the `fullURL`? If not, maybe we should improve the `fullURL`

https://github.com/sashabaranov/go-openai/blob/af9ff510779fb9e64110f1034df742bde751ba1b/client.go#L106

jsburckhardt:
I'll give it a shot updating fullURL, the thing is the URL/API changes between /models and /completions and fullURL isn't aware of the service that is getting called. Leave it with me I'll check what I can do.
completions-> {endpoint}/openai/deployments/{engine}/chat/completions?api-version={api_version}
models        -> {endpoint}/openai/models?api-version=2022-12-01

sashabaranov:
@jsburckhardt got it! The`fullURL` knows about what endpoint is passed to it, so we can make a special case there

jsburckhardt:
@sashabaranov updated by adding some logic into the fullURL, should be ok now. Let me know ;)

client.go
```diff
@@ -103,6 +103,11 @@ func (c *Client) fullURL(suffix string) string {
 	if c.config.APIType == APITypeAzure || c.config.APIType == APITypeAzureAD {
 		baseURL := c.config.BaseURL
 		baseURL = strings.TrimRight(baseURL, "/")
+		// if suffix is /models change to {endpoint}/openai/models?api-version=2022-12-01
+		// https://learn.microsoft.com/en-us/rest/api/cognitiveservices/azureopenaistable/models/list?tabs=HTTP
+		if strings.Contains(suffix, "/models") {
+			return fmt.Sprintf("%s/%s%s?api-version=%s", baseURL, azureAPIPrefix, suffix, "2022-12-01")
```
sashabaranov, line 109:
```suggestion
			return fmt.Sprintf("%s/%s%s?api-version=%s", baseURL, azureAPIPrefix, suffix, c.config.APIVersion)
```

jsburckhardt: sorry mate, for some reason I thought I had to use an specific API version. Tested and works well. Cheers

sashabaranov (approved):
Thank you!

Issue #292, opened 2023-05-01
malformed Error handling contradicts official documentation instruction


qhenkart:
The official Open-AI documentation gives clear instruction on handling API errors [here](https://platform.openai.com/docs/guides/error-codes/api-errors)

Particularly they quote

> [Handling errors](https://platform.openai.com/docs/guides/error-codes/handling-errors)
> We advise you to programmatically handle errors returned by the API. To do so, you may want to use a code snippet like below:

Unfortunately the status code is embedded into an error string [here](https://github.com/sashabaranov/go-openai/blob/master/client.go#L157)
forcing consumers to string match when looking for status codes from the server.

Ideally the library would utilize its existing APIError struct that currently satisfies the error interface so that it can maintain its current structure with the possibility of type conversion to pull out status codes.

The library almost allows for this but overwrites the struct with an anonymous error string in the last moment

 To match the OpenAI documentation, ideally consumer code would look something like this

```
          respErr, ok := err.(*openai.APIError)
		if !ok {
                        // handle unknown error
		}

		switch respErr.StatusCode {
		case 401:
		        // invalid auth or key
		case 429:
			// rate limiting or engine overload (probably an additional string check required here to determine which 429)
		case 500:
			// openai server error
		default:
			// unhandled
		}
```

sashabaranov:
@qhenkart, the code example is great! 🙌🏻  Would love for something like this to be added to README (maybe with the proper `errors.Is` and `errors.As` usage).

Issue #288, opened 2023-04-30
How to get openai api result status code ?


jiriyao:
(How to get openai api result status code )
ChatGPT error 401: {
"error": {
"message": "Incorrect API key provided: sk-bwInV***************************************pk3i. You can find your API key at https://platform.openai.com/account/api-keys.",
"type": "invalid_request_error",
"param": null,
"code": "invalid_api_key"
}
}

how to get  resp.StatusCode for logic ? thanks.

jiriyao:
not surpport it ?

vvatanabe:
@jiriyao Instructions on how to check status codes are provided in the README.

```go
e := &openai.APIError{}
if errors.As(err, &e) {
  switch e.HTTPStatusCode {
    case 401:
      // invalid auth or key (do not retry)
    case 429:
      // rate limiting or engine overload (wait and retry) 
    case 500:
      // openai server error (retry)
    default:
      // unhandled
  }
}
```
Refs: https://github.com/sashabaranov/go-openai#other-examples

PR #293, opened 2023-05-01
maintain underlying error structs to allow for type conversion

qhenkart:
Resolves https://github.com/sashabaranov/go-openai/issues/292, https://github.com/sashabaranov/go-openai/issues/288

Maintains underlying error structs to allow consumers to utilize type conversion and create a defensive structure as defined by the Open AI error handling documentation.

It also slightly modifies the Error() functions to that they are responsible for converting the struct into an informational string versus just a fragment of its parent

Modifies only a single test that was missing a colon when all of the other tests have colons (including the RequestError Error function that was missing the colon that the other Error function included

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,13 +10,13 @@ This library provides Go clients for [OpenAI API](https://platform.openai.com/).
 * DALL·E 2
 * Whisper
 
-Installation:
+### Installation:
 ```
 go get github.com/sashabaranov/go-openai
 ```
 
 
-ChatGPT example usage:
+### ChatGPT example usage:
 
 ```go
 package main
@@ -52,9 +52,7 @@ func main() {
 
 ```
 
-
-
-Other examples:
+### Other examples:
 
 <details>
 <summary>ChatGPT streaming completion</summary>
--- a/README.md
+++ b/README.md
@@ -468,7 +468,7 @@ Open-AI maintains clear documentation on how to [handle API errors](https://plat
 
 example:
 ```
-e := &APIError{}
+e := &openai.APIError{}
 if errors.As(err, &e) {
   switch e.HTTPStatusCode {
     case 401:
--- a/README.md
+++ b/README.md
@@ -65,12 +65,12 @@ if errors.As(err, &e) {
     case 401:
       // invalid auth or key (do not retry)
     case 429:
-	  // rate limiting or engine overload (wait and retry) 
-	case 500:
-	  // openai server error (retry)
-	default:
-	  // unhandled
-	}
+      // rate limiting or engine overload (wait and retry) 
+    case 500:
+      // openai server error (retry)
+    default:
+      // unhandled
+  }
 }
 
 ```
--- a/README.md
+++ b/README.md
@@ -52,30 +52,6 @@ func main() {
 
 ```
 
-
-### Error handling:
-
-Open-AI maintains clear documentation on how to [handle API errors](https://platform.openai.com/docs/guides/error-codes/api-errors)
-
-example:
-```
-e := &APIError{}
-if errors.As(err, &e) {
-  switch e.HTTPStatusCode {
-    case 401:
-      // invalid auth or key (do not retry)
-    case 429:
-      // rate limiting or engine overload (wait and retry) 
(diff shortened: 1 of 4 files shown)


sashabaranov:
@qhenkart thank you for the PR! It's a bit hard for me to follow all the error cases here — could you please either add a test or code sample clearly demonstrating the benefits of this approach?

client.go
```diff
@@ -144,15 +144,16 @@ func (c *Client) handleErrorResp(resp *http.Response) error {
...
 	err := json.NewDecoder(resp.Body).Decode(&errRes)
 	if err != nil || errRes.Error == nil {
-		reqErr := RequestError{
+		reqErr := &RequestError{
 			HTTPStatusCode: resp.StatusCode,
 			Err:            err,
 		}
 		if errRes.Error != nil {
-			reqErr.Err = errRes.Error
+			reqErr.Err = fmt.Errorf(errRes.Error.Message)
```
sashabaranov, line 152: Wouldn't we lose the ability to use `errors.Is` and `errors.As` here? https://pkg.go.dev/errors

qhenkart: you are right. I have updated the code and the error message to cover this case and still match the expected error result illustrated in the existing tests

qhenkart:
@sashabaranov I have added a test to ensure type conversion as well as updated the README as requested in the linked issue to provide guidance on how to handle errors utilizing proper error package syntax.

I have another proposal that would further improve the error handling experience for consumers of this library. But I would like to keep this PR as the simple fix that it is so I will open up a separate issue so we can discuss

README.md
```diff
@@ -53,8 +53,30 @@ func main() {
 ```
 
 
+### Error handling:
 
-Other examples:
+Open-AI maintains clear documentation on how to [handle API errors](https://platform.openai.com/docs/guides/error-codes/api-errors)
```
sashabaranov, line 58: Could we please wrap this into `<details>` tag as it is done below?

client.go
```diff
@@ -144,15 +144,16 @@ func (c *Client) handleErrorResp(resp *http.Response) error {
...
 		}
 		if errRes.Error != nil {
 			reqErr.Err = errRes.Error
 		}
-		return fmt.Errorf("error, %w", &reqErr)
+		return reqErr
 	}
+
 	errRes.Error.HTTPStatusCode = resp.StatusCode
-	return fmt.Errorf("error, status code: %d, message: %w", resp.StatusCode, errRes.Error)
```
sashabaranov, line 157: I wonder if the previous version was working with the `errors.As` due to `%w` usage

README.md
```diff
@@ -462,3 +460,29 @@ func main() {
...
 </details>
+
+<details>
+<summary>Error handling</summary>
+
+Open-AI maintains clear documentation on how to [handle API errors](https://platform.openai.com/docs/guides/error-codes/api-errors)
+
+example:
+```
+e := &APIError{}
```
sashabaranov, line 471: I guess it should be `openai.APIError`

client_test.go
```diff
@@ -135,6 +136,12 @@ func TestHandleErrorResp(t *testing.T) {
 				t.Errorf("Unexpected error: %v , expected: %s", err, tc.expected)
 				t.Fail()
 			}
+
+			e := &APIError{}
+			if !errors.As(err, &e) {
```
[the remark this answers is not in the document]
qhenkart: according to the docs. It is supposed to be a double pointer [https://go.dev/blog/go1.13-errors](https://go.dev/blog/go1.13-errors)

> Note: although it may feel odd to take a pointer to a pointer, in this case it is correct. Think of it instead as taking a pointer to a value of the error type; it so happens in this case that the returned error is a pointer type.

client.go
qhenkart: yes you are right. I ran into the issue because I was using type conversion instead of errors.As. I suppose the extra information in the documentation would have negated the need for this PR. However I still think it is an improvement for the error type to handle its own messaging and formatting. Also it doesn't hurt to also allow type conversion even if it is no longer the idiomatic methodology

Issue #296, opened 2023-05-03
Remove dependency on httbin


sashabaranov:
https://github.com/sashabaranov/go-openai/blob/104c0c0b63e4915f4ff44faf07a153949e16ab9f/api_test.go#L230

PR #297, opened 2023-05-03
Resolve issue #296: Remove httpbin dependency from TestRequestError

YukiBobier:
This pull request addresses and resolves issue #296 by removing the dependency on httpbin from the TestRequestError function.


sashabaranov (approved):
Perfect, thank you!

PR #298, opened 2023-05-03
Add Readme example to example_test.go

hoani:
Moves the `README` examples to `example_test.go` which improves the go docs:

(from a local run)
Screenshot 2023-05-04 at 11 01 47 AM

I also added executable examples under an `example` folder. This works pretty well and I suspect is what most people will end up referring to when they first start using the `go-openai` package:

Screenshot 2023-05-04 at 11 02 55 AM

I've removed the examples from the `README.md` - but happy to put them back... 

Kind of relates to https://github.com/sashabaranov/go-openai/issues/80 - but really these examples aren't really testable. The purpose of this package is to interact with external AI APIs... which to be fair is never going to have deterministic results 😅


.golangci.yml
```diff
@@ -9,6 +9,9 @@ run:
   # Default: 1m
   timeout: 3m
 
+  skip-files:
+  - example_test.go
```
sashabaranov, line 13: what exactly do we break in this file? Maybe `// nolint` would be a better fit?

README.md
```diff
@@ -49,440 +49,6 @@ func main() {
 
 	fmt.Println(resp.Choices[0].Message.Content)
 }
-
-```
-
-### Other examples:
```
sashabaranov, line 55: Let's not remove examples from the README, as it's easy to reach

sashabaranov:
Great stuff, thank you for the PR!

.golangci.yml
hoani: Good point, I fixed a bunch of issues in `example_test` - only needed `// nolint` in one place 😉

README.md
hoani: Done!

common.go
```diff
@@ -1,6 +1,7 @@
-// common.go defines common types used throughout the OpenAI API.
 package openai
 
+// common.go defines common types used throughout the OpenAI API.
```
sashabaranov, line 3: why this change?

hoani: Ah, if that comment is above the package name, it shows up in the go docs package description. It didnt look like it was meant to be there

sashabaranov (approved):

sashabaranov (approved):

PR #306, opened 2023-05-09
change azure engine config to modelMapper

GargantuaX:
"azure engine" refers to the deployment name, and each deployment name can only deploy one model. 
This means that if I use multiple models, I need to configure multiple engines.
Since a client can only configure one engine, I have to maintain many clients, which is not easy.

I submitted a pull request that uses a modelMapper to map the model to the deployment name. By default, the Azure deployment name uses the model name, it can work without any engine configuration.

### DIFF
--- a/client.go
+++ b/client.go
@@ -98,7 +98,10 @@ func decodeString(body io.Reader, output *string) error {
 	return nil
 }
 
-func (c *Client) fullURL(suffix string) string {
+// fullURL returns full URL for request.
+// If API type is Azure, model is required to get deployment name.
+// If API type is OpenAI, model will be ignored.
+func (c *Client) fullURL(suffix, model string) string {
 	// /openai/deployments/{engine}/chat/completions?api-version={api_version}
 	if c.config.APIType == APITypeAzure || c.config.APIType == APITypeAzureAD {
 		baseURL := c.config.BaseURL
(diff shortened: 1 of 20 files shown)


sashabaranov:
Wow, thank you so much for such a great work!

client.go
```diff
@@ -98,7 +98,10 @@ func decodeString(body io.Reader, output *string) error {
 	return nil
 }
 
-func (c *Client) fullURL(suffix string) string {
+// fullURL returns full URL for request.
+// If API type is Azure, model is required to get deployment name.
+// If API type is OpenAI, model will be ignored.
+func (c *Client) fullURL(suffix, model string) string {
```
sashabaranov, line 104: As a suggestion: do we want to use variadic options here instead of the required `model` argument? That would require fewer changes to other parts of the library. https://gobyexample.com/variadic-functions

We didn't adopt this pattern for the whole library in #278, but it seems useful in the context of the `fullURL` method.

What do you think?

config.go
```diff
@@ -50,14 +50,21 @@ func DefaultConfig(authToken string) ClientConfig {
 	}
 }
 
-func DefaultAzureConfig(apiKey, baseURL, engine string) ClientConfig {
+func DefaultAzureConfig(apiKey, baseURL string, modelMapper map[string]string) ClientConfig {
```
sashabaranov, line 53: I suggest we remove `modelMapper` argument here as it doesn't really fit the "default" config. Let's allow modifying it afterward.

client.go
GargantuaX: I agree, I would be happy to do this job.

config.go
```diff
@@ -50,14 +50,21 @@ func DefaultConfig(authToken string) ClientConfig {
...
+	if len(modelMapper) == 0 {
+		modelMapper = map[string]string{
+			GPT3Dot5Turbo0301: "gpt-35-turbo-0301",
```
sashabaranov, line 56: What if `modelMapper` would be just a function that user can provide?

config.go
GargantuaX: OK

config.go
GargantuaX: Your suggestion is great, I tried to make those changes, the code become more concise and has better extensibility. I will submit the new code later.

GargantuaX:
Please review the code，thx

wuhuizuo:
It's not compatible, it should not release in patch version!

Ref:
- https://github.com/sashabaranov/go-openai/compare/v1.9.3...v1.9.4

Issue #304, opened 2023-05-08
Move some code into `internal` directory


sashabaranov:
The root directory of the project is pretty bloated, it would be nice if can move the following into the `internal` directory:

- [ ] `error_accumulator`
- [ ] `request_builder`
- [x] `form_builder`
- [x] `unmarshaler`
- [x] `marshaller`

Also, I don't think we need `Makefile` anymore given the CI setup.

PR #311, opened 2023-05-15
Move form_builder into internal pkg.

JiayueShi:
Fix one task of #304


sashabaranov (approved):
Thank you so much for this PR!

PR #318, opened 2023-05-19
Update README.md with Azure OpenAI Embeddings example

coggsfl:
@sashabaranov I know there's an open PR with a more complex example for embeddings, but I think this gives a clear approach to anyone looking to implement embeddings quickly with Azure OpenAI


sashabaranov (approved):
Thank you for the PR!

PR #322, opened 2023-05-23
modify authToken to AuthToken

jau1jz:
Cannot set authToken when using openai. NewClientWithConfig because this parameter is private


sashabaranov:
Hey, thank you for the PR! This is by design, please refer to the previous discussion here https://github.com/sashabaranov/go-openai/pull/206

Issue #333, opened 2023-05-31
Temperature param being ignored?


rishsriv:
Hi there! Thank you for maintaining this package. It has made our lives so much simpler.

I wanted to check if the temperature parameter is somehow being ignored in the library?

We have an implementation that is something like this
```
resp, err := Openai.CreateChatCompletion(
	c,
	openai.ChatCompletionRequest{
		Model:       "gpt-3.5-turbo",
		Messages:    messages,
		MaxTokens:   600,
		Temperature: 0.0,
		Stop:        []string{"```"},
	},
)
```

But this is ignoring the temperature parameter and returning a different response every time. This issue does not seem to be present when using the Python library or making a CURL request to the OpenAI API.

vvatanabe:
@rishsriv @sashabaranov I investigated the cause of the problem.
https://github.com/sashabaranov/go-openai/pull/334#issue-1735004506

rishsriv:
Thank you so much for investigating this, @vvatanabe !

vvatanabe:
Duplicate https://github.com/sashabaranov/go-openai/issues/9

vvatanabe:
This is a duplicate issue with https://github.com/sashabaranov/go-openai/issues/9, so please discuss it in https://github.com/sashabaranov/go-openai/issues/9 from now on.

PR #334, opened 2023-05-31
remove omitempty tag from temperature fields (#333)

vvatanabe:
~I confirmed the OpenAI API documentation. The default value for temperature is `1`. The temperature field of the request structure was given an `omitempty` tag, so a value of `0` was removed from the request parameter. It would be better to remove the `omitempty` tag so that `0` can be used.~

Oops, removing the omitempty tag changes the default value in this client library.
I thought it would be better to change the temperature type to pointer(*float32). But that is BREAKING CHANGES.

Refs:
- https://platform.openai.com/docs/api-reference/chat/create#chat/create-temperature
- https://platform.openai.com/docs/api-reference/completions/create#completions/create-temperature
- https://platform.openai.com/docs/api-reference/edits/create#edits/create-temperature

Note: For audio api only, the default value for temperature is `0`.
Refs: https://platform.openai.com/docs/api-reference/audio/create

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -34,7 +34,7 @@ type ChatCompletionRequest struct {
 	Model            string                  `json:"model"`
 	Messages         []ChatCompletionMessage `json:"messages"`
 	MaxTokens        int                     `json:"max_tokens,omitempty"`
-	Temperature      float32                 `json:"temperature,omitempty"`
+	Temperature      float32                 `json:"temperature"`
 	TopP             float32                 `json:"top_p,omitempty"`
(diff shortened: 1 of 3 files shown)


vvatanabe:
Oops, removing the omitempty tag changes the default value in this client library.

I thought it would be better to change the temperature type to pointer(*float32). But that is BREAKING CHANGES.

vvatanabe:
Wait for a major version upgrade.

sashabaranov:
@vvatanabe I like the v2 idea, I guess we should discuss on what highest ROI breakages might be

PR #336, opened 2023-06-01
feature: Rate Limit

GargantuaX:
Whenever you use an API, whether it's OpenAI or Azure, you will be required to limit your rate.
Since everyone has to write this code, perhaps it would be better to include it in the API library.

Document Reference:
https://learn.microsoft.com/en-us/azure/cognitive-services/openai/quotas-limits#quotas-and-limits-reference
https://platform.openai.com/docs/guides/rate-limits/overview


sashabaranov:
Hey, thank you for the PR! This is hard one for me — I'm not I'm not sure we should add rate limiting to this library due to extra complexity and dependencies.

Do you think this might work as a separate library? It might also be expanded to cost limiting (it was mentioned in a couple of issues). I would be happy to link to it from the README.

GargantuaX:
> 

I completely agree that adding rate limiting introduces additional complexity and dependencies, which will be a burden to other maintainers. I have no immediate plans to expand this feature into a separate library. Perhaps it can be pushed forward in this lib when there are more people who need rate limiting.

sashabaranov:
@GargantuaX hey, I think I've found a proper place for rate limiting 🙌🏻  Please check out https://github.com/go-llm/llmlite and my comment here: https://github.com/sashabaranov/go-openai/pull/391#issuecomment-1603043445

PR #339, opened 2023-06-02
Refactor streamReader: Replace goto Statement with Loop in Recv Method

YukiBobier:
@sashabaranov I'd appreciate your review on these changes.

# Description

This pull request introduces a refactor to the Recv method in the streamReader type, which improves code readability and maintainability.

## Main Changes

- Replaced label with for block
- Replaced goto with continue to restart the loop if the condition is met.
- Replaced short variable declaration with regular variable declaration due to scope issue
  - [For is a block scoped while labels are not block scoped](https://go.dev/ref/spec#Goto_statements:~:text=to%20other%20identifiers%2C-,labels%20are%20not%20block%20scoped,-and%20do%20not)

These changes maintain the functionality of the Recv method, while improving the clarity of the control flow within the function. The goto statement can sometimes make the code hard to understand and maintain, hence this refactor aims to resolve that.

## Testing

The changes have been tested to ensure that they don't introduce any regressions.

### DIFF
--- a/stream_reader.go
+++ b/stream_reader.go
@@ -32,41 +32,43 @@ func (stream *streamReader[T]) Recv() (response T, err error) {
 
 	var emptyMessagesCount uint
 
-waitForData:
-	line, err := stream.reader.ReadBytes('\n')
-	if err != nil {
-		respErr := stream.unmarshalError()
-		if respErr != nil {
-			err = fmt.Errorf("error, %w", respErr.Error)
-		}
-		return
-	}
-
-	var headerData = []byte("data: ")
-	line = bytes.TrimSpace(line)
-	if !bytes.HasPrefix(line, headerData) {
-		if writeErr := stream.errAccumulator.Write(line); writeErr != nil {
-			err = writeErr
-			return
-		}
-		emptyMessagesCount++
-		if emptyMessagesCount > stream.emptyMessagesLimit {
-			err = ErrTooManyEmptyStreamMessages
+	for {
+		var line []byte
+		line, err = stream.reader.ReadBytes('\n')
+		if err != nil {
+			respErr := stream.unmarshalError()
+			if respErr != nil {
+				err = fmt.Errorf("error, %w", respErr.Error)
+			}
 			return
 		}
 
-		goto waitForData
-	}
+		var headerData = []byte("data: ")
+		line = bytes.TrimSpace(line)
+		if !bytes.HasPrefix(line, headerData) {
+			if writeErr := stream.errAccumulator.Write(line); writeErr != nil {
+				err = writeErr
+				return
+			}
+			emptyMessagesCount++
+			if emptyMessagesCount > stream.emptyMessagesLimit {
+				err = ErrTooManyEmptyStreamMessages
+				return
+			}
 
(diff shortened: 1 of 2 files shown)


YukiBobier:
@sashabaranov I've added a test and modified the branch to address the coverage issue.

sashabaranov:
Thank you for the PR! While we optimize this section, I've got a couple of comments…

stream_reader.go
```diff
@@ -32,41 +32,43 @@ func (stream *streamReader[T]) Recv() (response T, err error) {
...
 
-	var headerData = []byte("data: ")
-	line = bytes.TrimSpace(line)
-	if !bytes.HasPrefix(line, headerData) {
-		if writeErr := stream.errAccumulator.write(line); writeErr != nil {
-			err = writeErr
-			return
+		var headerData = []byte("data: ")
+		line = bytes.TrimSpace(line)
+		if !bytes.HasPrefix(line, headerData) {
```
sashabaranov, line 48: Can we maybe flip this if condition?

```go
if bytes.HasPrefix(line, headerData) {
    line = bytes.TrimPrefix(line, headerData)
    // ... check for [DONE], etc.
}
```

In that case all prefix-manipulating logic is located close-by

stream_reader.go
```diff
@@ -32,41 +32,43 @@ func (stream *streamReader[T]) Recv() (response T, err error) {
 
 	var emptyMessagesCount uint
 
-waitForData:
-	line, err := stream.reader.ReadBytes('\n')
-	if err != nil {
-		respErr := stream.errAccumulator.unmarshalError()
-		if respErr != nil {
-			err = fmt.Errorf("error, %w", respErr.Error)
+	for {
```
sashabaranov, line 35: Due to added level of indentaion I wonder if we should factor out `for` body in the separate function (or maybe parts of the `for` body)

stream_reader.go
YukiBobier: Yes, we can. Just as you mentioned, if we move the logic for manipulating the prefix closer to the condition statement, it would make sense.

However, I believe it's better to leave it as it is due to the [happy-path idiom](https://maelvls.dev/go-happy-line-of-sight/) and the fail-fast principle. It's a good practice to keep things organized in this way.👍

By the way, I share the same opinion about keeping related processes close to each other. I'm currently working on another pull request to improve that aspect.

stream_reader.go
YukiBobier: I agree with you. Extracting the big, deeply nested for-loop into a separate method would be a good idea.

I will do so👍

YukiBobier:
I've done [this](https://github.com/sashabaranov/go-openai/pull/339#discussion_r1215739508).

But hmm... Coverage issues have occurred...
I will address them.

YukiBobier:
@sashabaranov 
I've addressed all the CI issues by adding two new tests and rebasing the branch. 

Additionally, I have responded to all the comments that were made. 👍

Could you please review it again?

sashabaranov:
@YukiBobier duh, we've got conflicts now due to other stuff

YukiBobier:
@sashabaranov Oops, I will resolve it👌

PR #342, opened 2023-06-06
Enable using http proxy set in the env

hqm19:
When connecting to OpenAI, try to use a proxy set in the environment variables, such as HTTP_PROXY,HTTPS_PROXY and NO_PROXY

### DIFF
--- a/config.go
+++ b/config.go
@@ -44,7 +44,11 @@ func DefaultConfig(authToken string) ClientConfig {
 		APIType:   APITypeOpenAI,
 		OrgID:     "",
 
-		HTTPClient: &http.Client{},
+		HTTPClient: &http.Client{
+			Transport: &http.Transport{
+				Proxy: http.ProxyFromEnvironment, // use http_proxy from env
+			},
+		},
 
 		EmptyMessagesLimit: defaultEmptyMessagesLimit,
 	}


vvatanabe:
It seems that the code in the standard `net/http` package already applies http.ProxyFromEnvironment by default. 

``` go
func (c *Client) transport() RoundTripper {
	if c.Transport != nil {
		return c.Transport
	}
	return DefaultTransport
}
```
Refs: https://github.com/golang/go/blob/38e2376f35907ebbb98419f1f4b8f28125bf6aaf/src/net/http/client.go#L200-L205


``` go
// DefaultTransport is the default implementation of Transport and is
// used by DefaultClient. It establishes network connections as needed
// and caches them for reuse by subsequent calls. It uses HTTP proxies
// as directed by the environment variables HTTP_PROXY, HTTPS_PROXY
// and NO_PROXY (or the lowercase versions thereof).
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: defaultTransportDialContext(&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
	}),
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
```
Refs: https://github.com/golang/go/blob/38e2376f35907ebbb98419f1f4b8f28125bf6aaf/src/net/http/transport.go#L38-L54

hqm19:
@vvatanabe  Yes, you are right!

PR #345, opened 2023-06-08
fix json marshaling error response of azure openai (#343)

### LINKED ISSUE #343: About Error handling
Hi, when I am using Azure openai chatgpt, I refer to the example of Error handling and cannot catch the error code 429. Every time the 429 code appears, the program will run into the default code block. I am not sure whether it is a problem with my operation or The content returned by azure openai is different from that of openai.
go-openai failed to serialize error messages correctly.THANKS!

```GOLANG
func OpenaiErrorAndCode(openaierr error) (code int, err error) {
	var openaiError = &openai.APIError{}
	if errors.As(openaierr, &openaiError) {
		switch openaiError.HTTPStatusCode {
		case 400:
			return 400, errors.New("xxxxxxx")
		case 401:
			return 401, errors.New("xxxxxxx")
		case 429:
			return 429, errors.New("xxxxxxx")
		case 500:
			return 500, errors.New("xxxxxxx")
		default:
				return 500, errors.New("unknown：" + openaierr.Error())
		}
	}
	return 500, openaierr
}
```
```GOLANG
stream, err := client.CreateChatCompletionStream(ctx, req)
if err != nil {
	fmt.Printf("ChatCompletionStream error: %v\n", err)
	_,openaierr := appError.OpenaiErrorAndCode(err)
	fmt.Printf(openaierr.Error())
}
```

The json returned by azure openai is as follows：
```json
{
    "error": {
        "code": "429",
        "message": "Requests to the Creates a completion for the chat message Operation under Azure OpenAI API version 2023-03-15-preview have exceeded token rate limit of your current OpenAI S0 pricing tier. Please retry after 20 seconds. Please go here: https://aka.ms/oai/quotaincrease if you would like to further increase the default rate limit."
    }
}
```

vvatanabe:
Fix #343

Azure OpenAI error response seems to be `type` field missing sometimes. Error (`unexpected end of JSON input`) when parsing `type` in `APIError.UnmarshalJSON` function.

Example error response:
```
# Azure OpenAI
{
   "error":{
      "code":"429",
      "message":"Requests to the Creates a completion for the chat message Operation under Azure OpenAI API version 2023-03-15-preview have exceeded token rate limit of your current OpenAI S0 pricing tier. Please retry after 17 seconds. Please go here: https://aka.ms/oai/quotaincrease if you would like to further increase the default rate limit."
   }
}
```

**Note:**
I also considered changing the `type` field to a pointer, but decided against it because that would be a breaking changes.

### DIFF
--- a/chat_stream_test.go
+++ b/chat_stream_test.go
@@ -255,6 +255,67 @@ func TestCreateChatCompletionStreamRateLimitError(t *testing.T) {
 	t.Logf("%+v\n", apiErr)
 }
 
+func TestAzureCreateChatCompletionStreamRateLimitError(t *testing.T) {
+	wantCode := "429"
+	wantMessage := "Requests to the Creates a completion for the chat message Operation under Azure OpenAI API " +
+		"version 2023-03-15-preview have exceeded token rate limit of your current OpenAI S0 pricing tier. " +
+		"Please retry after 20 seconds. " +
+		"Please go here: https://aka.ms/oai/quotaincrease if you would like to further increase the default rate limit."
+
+	server := test.NewTestServer()
+	server.RegisterHandler("/openai/deployments/gpt-35-turbo/chat/completions",
+		func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set("Content-Type", "application/json")
+			w.WriteHeader(http.StatusTooManyRequests)
+			// Send test responses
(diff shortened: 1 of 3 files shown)


vvatanabe:

sashabaranov (approved):
Thank you! 🙌🏻

PR #352, opened 2023-06-09
Implement Azure DALL-E 2 ImageRequest API

coggsfl:
Microsoft recently updated their DALL-E 2 API to more closely match OpenAI, however as usual there are still significant differences.

This PR implements ImageRequest API for DALL-E 2

I have addressed multiple listing issues, and I think the code is good for review now.


coggsfl:
@sashabaranov I'm assuming the linter is somewhat new as I think it's picking up some pre-existing "issues" and is erroring on some questionable issues.  Please review and advise. Thanks!

PR #355, opened 2023-06-11
Azure DALL-E 2 Implementation

coggsfl:


coggsfl:
@sashabaranov apologies for the reduction in code coverage, however this does add a significant new capability and will happily look for further test coverage opportunities once merged.

PR #356, opened 2023-06-12
refactoring tests with mock servers (#30)

vvatanabe:


vvatanabe:

PR #359, opened 2023-06-13
Improve (#356) to support registration of wildcard URLs

coggsfl:
Improve on (#356) issue to support wildcard URLs required by Azure API path structure
Add Azure ChatCompletions test cases including test of custom deployment name

### DIFF
--- a/chat_test.go
+++ b/chat_test.go
@@ -85,24 +85,6 @@ func TestAzureChatCompletions(t *testing.T) {
 	checks.NoError(t, err, "CreateAzureChatCompletion error")
 }
 
-func TestAzureChatCompletionsWithCustomDeploymentName(t *testing.T) {
-	client, server, teardown := setupAzureTestServerWithCustomDeploymentName()
-	defer teardown()
-	server.RegisterHandler("/openai/deployments/*", handleChatCompletionEndpoint)
-
-	_, err := client.CreateChatCompletion(context.Background(), ChatCompletionRequest{
-		MaxTokens: 5,
-		Model:     GPT3Dot5Turbo,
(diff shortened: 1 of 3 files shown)


coggsfl:
@sashabaranov please review.  I have a larger PR I'm working on for Azure DALL-E 2 support, however this is a simple change which allows test support for Azure OpenAI endpoints.

coggsfl:
@sashabaranov let me know if you have any questions about the need for wildcard support in the test handlers?  

If the project is going to provide first class support for Azure OpenAI then support for wildcard handlers is required due to the deployment name being a component of the URL structure : /openai/deployments/{model}/chat/completions?api-version={api_version}

chat_test.go
```diff
@@ -67,6 +67,42 @@ func TestChatCompletions(t *testing.T) {
...
+		Model:     GPT3Dot5Turbo,
+		Messages: []ChatCompletionMessage{
+			{
+				Role:    ChatMessageRoleUser,
+				Content: "Hello!",
+			},
+		},
+	})
+	checks.NoError(t, err, "CreateAzureChatCompletionWithCustomDeploymentName error")
+}
```
vvatanabe, line 104: @coggsfl Why is this test case necessary? The following test case seems to cover it.
https://github.com/sashabaranov/go-openai/blob/master/config_test.go#L9-L62

coggsfl: Good call @vvatanabe!  I'll pull that one back.

vvatanabe:
LGTM

sashabaranov (approved):
Thank you for the PR!

PR #363, opened 2023-06-13
Add new parameter "functions" for chat completion API

samirkhoja:
Update chat completion API to include new parameter "functions" for model variant "0613" as described here: https://openai.com/blog/function-calling-and-other-api-updates

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -26,7 +26,15 @@ type ChatCompletionMessage struct {
 	// the documentation for the official library for python:
 	// - https://github.com/openai/openai-python/blob/main/chatml.md
 	// - https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
-	Name string `json:"name,omitempty"`
+	Name         string       `json:"name,omitempty"`
+	
+	// New property introduced with model variant "0613"
+	FunctionCall FunctionCall `json:"function_call,omitempty"`
+}
+
+type FunctionCall struct {
+	Name      string `json:"name"`
+	Arguments string `json:"arguments"`
 }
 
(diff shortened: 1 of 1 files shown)


samirkhoja:
Quickly did this in the GitHub UI so lint checks won't pass but just leaving it here as reference.

coggsfl:
@samirkhoja thanks for posting - this is big news!  Azure version is only a few days away too.

chat.go
```diff
@@ -26,23 +27,50 @@ type ChatCompletionMessage struct {
 	// the documentation for the official library for python:
```
j178, line 27: `name` now is in the official documentation: https://platform.openai.com/docs/api-reference/chat/create#chat/create-name

chat.go
```diff
@@ -26,23 +27,50 @@ type ChatCompletionMessage struct {
...
-	N                int                     `json:"n,omitempty"`
-	Stream           bool                    `json:"stream,omitempty"`
-	Stop             []string                `json:"stop,omitempty"`
-	PresencePenalty  float32                 `json:"presence_penalty,omitempty"`
-	FrequencyPenalty float32                 `json:"frequency_penalty,omitempty"`
-	LogitBias        map[string]int          `json:"logit_bias,omitempty"`
-	User             string                  `json:"user,omitempty"`
+	Model            string                   `json:"model"`
+	Messages         []ChatCompletionMessage  `json:"messages"`
+	Functions        []ChatCompletionFunction `json:"functions,omitempty"`
```
j178, line 63: Another top-level parameter `function_call` should be added: https://platform.openai.com/docs/api-reference/chat/create#chat/create-function_call

PR #364, opened 2023-06-14
Support GPT4-0613

carsonkahn-external:
Add support for the new models (https://openai.com/blog/function-calling-and-other-api-updates). 

~TODO in separate PR: support Function Calling (https://platform.openai.com/docs/guides/gpt/function-calling)~ *Addressed in [PR 363](https://github.com/sashabaranov/go-openai/pull/363)


sashabaranov:
Thank you for the PR, it seems that these models are added in the other PRs already :D

Issue #360, opened 2023-06-13
Implement June OpenAI updates


sashabaranov:
See updates at: https://openai.com/blog/function-calling-and-other-api-updates

From the top of my mind
- [x] Implement support for `*-0613` models (e.g. `gpt-4-0613`)
- [ ] Implement function calling

vvatanabe:
Which fields in the `functions` are required or optional? It may be necessary to consider, for example, making the type pointer depending on the field.

refs: https://openai.com/blog/function-calling-and-other-api-updates
``` json
{
   "model":"gpt-3.5-turbo-0613",
   "messages":[
      {
         "role":"user",
         "content":"What is the weather like in Boston?"
      }
   ],
   "functions":[
      {
         "name":"get_current_weather",
         "description":"Get the current weather in a given location",
         "parameters":{
            "type":"object",
            "properties":{
               "location":{
                  "type":"string",
                  "description":"The city and state, e.g. San Francisco, CA"
               },
               "unit":{
                  "type":"string",
                  "enum":[
                     "celsius",
                     "fahrenheit"
                  ]
               }
            },
            "required":[
               "location"
            ]
         }
      }
   ]
}
```

vvatanabe:
The official documentation related to `functions` has been updated.
https://platform.openai.com/docs/api-reference/chat/create#chat/create-name

danielchristianschroeter:
https://platform.openai.com/docs/guides/gpt/function-calling

„note: the model may generate invalid JSON or hallucinate parameters“

Perhaps we should ensure that we handle this situation properly.

stillmatic:
Could I suggest using the [jsonschema](https://github.com/invopop/jsonschema) library to construct JSONSchema, rather than reimplementing a fair bit of the spec? That library is well-tested and maintained and feels like a useful dependency to add, versus trying to check for edge cases.

I'm happy to make a PR that expects an input like https://github.com/stillmatic/gollum/blob/main/functions.go#L10 and simply adds the missing fields / updates enums, without additional logic.

It may be helpful to others: I've included an additional example of end-to-end parsing -- https://github.com/stillmatic/gollum/blob/a66a4664e02743c8b2e0b466493b9632af10d0de/functions_test.go#L135 . So you can take any Go json struct, generate the appropriate JSON schema for OpenAI, and parse the response back to the original struct. The implementation is extremely terse (I will replace the JSON validator stuff with the JSON schema validator) but probably should be standalone vs in this library.

[...]

PR #367, opened 2023-06-14
feat(chat): support function call api

Ccheers:
implement #360

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -13,6 +13,8 @@ const (
 	ChatMessageRoleAssistant = "assistant"
 )
 
+const chatCompletionsSuffix = "/chat/completions"
+
 var (
 	ErrChatCompletionInvalidModel       = errors.New("this model is not supported with this method, please use CreateCompletion client method instead") //nolint:lll
 	ErrChatCompletionStreamNotSupported = errors.New("streaming is not supported with this method, please use CreateChatCompletionStream")              //nolint:lll
@@ -27,6 +29,14 @@ type ChatCompletionMessage struct {
 	// - https://github.com/openai/openai-python/blob/main/chatml.md
 	// - https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
 	Name string `json:"name,omitempty"`
+
+	FunctionCall *FunctionCall `json:"function_call,omitempty"`
+}
+
+type FunctionCall struct {
+	Name string `json:"name,omitempty"`
+	// call function with arguments in JSON format
+	Arguments string `json:"arguments,omitempty"`
 }
 
 // ChatCompletionRequest represents a request structure for chat completion API.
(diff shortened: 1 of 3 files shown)


Ccheers:
@sashabaranov

chat.go
```diff
@@ -43,12 +51,68 @@ type ChatCompletionRequest struct {
...
 	LogitBias        map[string]int          `json:"logit_bias,omitempty"`
 	User             string                  `json:"user,omitempty"`
+	Functions        []*FunctionsDefine      `json:"functions,omitempty"`
+	FunctionCall     string                  `json:"function_call,omitempty"`
 }
 
+type FunctionsDefine struct {
+	Name        string      `json:"name,omitempty"`
+	Description string      `json:"description,omitempty"`
+	Parameters  *FuncParams `json:"parameters,omitempty"`
```
vvatanabe, line 61: @Ccheers The official documentation says `parameters` is optional, but when you actually call the API, it says `parameters` is required.

```
$ curl https://api.openai.com/v1/chat/completions -u :$OPENAI_API_KEY -H 'Content-Type: application/json' -d '{
  "model": "gpt-3.5-turbo-0613",
  "messages": [
    {"role": "user", "content": "What is the weather like in Boston?"}
  ],
  "functions": [
    {
      "name": "get_current_weather"
    }
  ]
}'
```

Response:
```json
{
   "error":{
      "message":"'parameters' is a required property - 'functions.0'",
      "type":"invalid_request_error",
      "param":null,
      "code":null
   }
}
```

Ccheers: You are right and the type must be "object", but as a api we shouldn't limit it. Function `CreateChatCompletionWithFunctionCall` will do it

Ccheers: we can add some comment to user

zbysir:
Looking forward to it

vvatanabe:
@sashabaranov @Ccheers  `CreateChatCompletionWithFunctionCall` function can aggregate multiple API calls related to "Function calling". Very useful. Excellent! 👍 👍 

But go-openai is just a client library for OpenAI API.  Should go-openai provide high-level abstraction like `CreateCompletionWithFunctionCall` function?

It may be difficult to design for abstraction between Azure OpenAI and OpenAI in the future when Azure OpenAI supports "Function calling".

I'm a bit concerned about other additional dependent libraries. Perhaps go-openai values maintenance cost and lightweight and dares to have zero dependencies.

It might be better to first just add some fields for "Function calling" to the `CreateChatCompletion` function request and response, and then gather user feedback to make a decision.

zbysir:
> @sashabaranov @Ccheers  `CreateChatCompletionWithFunctionCall` function can aggregate multiple API calls related to "Function calling". Very useful. Excellent! 👍 👍 
> 
> But go-openai is just a client library for OpenAI API.  Should go-openai provide high-level abstraction like `CreateCompletionWithFunctionCall` function?
> 
> It may be difficult to design for abstraction between Azure OpenAI and OpenAI in the future when Azure OpenAI supports "Function calling".
> 
> I'm a bit concerned about other additional dependent libraries. Perhaps go-openai values maintenance cost and lightweight and dares to have zero dependencies.
> 
> It might be better to first just add some fields for "Function calling" to the `CreateChatCompletion` function request and response, and then gather user feedback to make a decision.

I quite agree! The signature of withfunctioncall is not universal enough, so users should decide for themselves.

ushakov:
My 2p: it seems gjson library is only used to extract `choices.0.message."+fmt.Sprintf("args_%d", i)` field from JSON, which should be relatively easy to work around without introducing the additional dependency

Ccheers:
> My 2p: it seems gjson library is only used to extract `choices.0.message."+fmt.Sprintf("args_%d", i)` field from JSON, which should be relatively easy to work around without introducing the additional dependency

It seems that using map[string]interface{} to deconstruct JSON for parameter extraction is also possible at present, but it is relatively more complicated.

jmacwhyte:
I agree with the issues raised here, I think this library should be kept very simple.

Although the "function call" information that the API returns is meant to be used to call a function in the client's local environment, that might not be how everyone intends to use it. I would prefer to simply have the ChatCompletion/ChatCompletionStream structs updated to include to support the function call details, and leave it up to the client to decide how they will consume that information.

That said, I do think @Ccheers has done some great work which I will use as reference in my own project!

Ccheers:
> > @sashabaranov @Ccheers  `CreateChatCompletionWithFunctionCall` function can aggregate multiple API calls related to "Function calling". Very useful. Excellent! 👍 👍
> > But go-openai is just a client library for OpenAI API.  Should go-openai provide high-level abstraction like `CreateCompletionWithFunctionCall` function?
> > It may be difficult to design for abstraction between Azure OpenAI and OpenAI in the future when Azure OpenAI supports "Function calling".
> > I'm a bit concerned about other additional dependent libraries. Perhaps go-openai values maintenance cost and lightweight and dares to have zero dependencies.
> > It might be better to first just add some fields for "Function calling" to the `CreateChatCompletion` function request and response, and then gather user feedback to make a decision.
> 
> I quite agree! The signature of withfunctioncall is not universal enough, so users should decide for themselves.

In the current implementation, JSON deserialization relies on Go Struct. However, the definition of `function call` seems to contradict our traditional deserialization approach. Therefore, I am considering extracting a separate function to support `function call`.

Ccheers:
> I agree with the issues raised here, I think this library should be kept very simple.
> 
> Although the "function call" information that the API returns is meant to be used to call a function in the client's local environment, that might not be how everyone intends to use it. I would prefer to simply have the ChatCompletionMessage (etc) structs updated to include to support the function call details, and leave it up to the client to decide how they will consume that information.
> 
> That said, I do think @Ccheers has done some great work which I will use as reference in my own project!

I think this is a trade-off. If we consider introducing `function_call` into the native API, we may need to parse JSON twice in the native API, as discussed above. And not everyone needs `function_call`.

jmacwhyte:
My vote is to not accept this PR just yet. We already have CreateChatCompletion() as well as CreateChatCompletionStream(), adding CreateChatCompletionWithFunctionCall() is a bit over the top when it's all the same API! I think the new function call feature should be able to fit nicely into the existing ChatCompletion structure (like @vvatanabe said), and I am willing to work on an alternate solution (tomorrow, probably) if no one else gets to it first.

PR #373, opened 2023-06-15
add items, which is required for array type

awormuth:
When using openai.JSONSchemaTypeArray, you must define and "Items" object, otherwise OpenAI will return a 400.


sashabaranov (approved):
Thank you!

PR #376, opened 2023-06-15
feat(config): can use the proxy URL to call the OpenAI API

gclhaha:
In some regions, OpenAI is not accessible. This commit can be completed by calling openai.NewClientWithConfig(openai.DefaultConfigWithUrl(authToken, proxyUrl)). This will be helpful to those people.:-)

### DIFF
--- a/config.go
+++ b/config.go
@@ -50,6 +50,22 @@ func DefaultConfig(authToken string) ClientConfig {
 	}
 }
 
+func DefaultConfigWithUrl(authToken string, proxyUrl string) ClientConfig {
+	if proxyUrl == "" {
+		proxyUrl = openaiAPIURLv1
+	}
+	return ClientConfig{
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -50,6 +50,22 @@ func DefaultConfig(authToken string) ClientConfig {
...
+		BaseURL:   proxyUrl,
+		APIType:   APITypeOpenAI,
+		OrgID:     "",
+
+		HTTPClient: &http.Client{},
+
+		EmptyMessagesLimit: defaultEmptyMessagesLimit,
+	}
+}
+
```
vvatanabe, line 68: FIY We can already.
```
cfg.BaseURL = "proxyUrl"
```

gclhaha: cool, looks like i made a wrong attempt earlier, thanks a lot.

PR #375, opened 2023-06-15
Add Items to JSONSchemaDefine for JsonTypeArray.

ryomak:
In the current version (v1.11.0), when trying to use the `JsonTypeArray` in FunctionCall, it is not possible to define the array type, resulting in a 400 error.

I propose adding `Items` to `JSONSchemaDefine`. This would allow the array type to be specified and would eliminate the aforementioned error.

Could you please have a moment to take a look at it?

### example
```
func main() {
	ctx := context.Background()
	c := openai.NewClient("key")
	response, err := c.CreateChatCompletion(ctx, openai.ChatCompletionRequest{
		Model: openai.GPT40613,
		Messages: []openai.ChatCompletionMessage{
			{
				Role:    openai.ChatMessageRoleUser,
				Content: "「What is the weather like in Boston?」Break down the above sentence into words",
			},
		},
		N:                0,
		Stream:           false,
		Stop:             nil,
		MaxTokens:        1024,
		PresencePenalty:  0,
		FrequencyPenalty: 0,
		LogitBias:        nil,
		Functions: []*openai.FunctionDefine{
			{
				Name:        "split_word",
				Description: "Break sentences into words",
				Parameters: &openai.FunctionParams{
					Type: openai.JSONSchemaTypeObject,
					Properties: map[string]*openai.JSONSchemaDefine{
						"list": {
							Type: openai.JSONSchemaTypeArray,
							Items: &openai.JSONSchemaDefine{
								Type: openai.JSONSchemaTypeString,
							},
							Description: "word list",
						},
						"count": {
							Type:        openai.JSONSchemaTypeNumber,
							Description: "word count",
						},
					},
					Required: []string{"list", "count"},
				},
			},
		},
	})
	if err != nil {
		panic(err)
	}
	if response.Choices[0].FinishReason != openai.FinishReasonFunctionCall {
		panic(err)
	}
	fmt.Println(response.Choices[0].Message.FunctionCall.Arguments)
}

/*
output: 
{
  "count": 7,
  "list": ["What", "is", "the", "weather", "like", "in", "Boston?"]
}
*/
```

### referrence
https://json-schema.org/understanding-json-schema/reference/array.html#id6

PR #377, opened 2023-06-15
feat: use json.rawMessage in functions, add test

stillmatic:


jmacwhyte:
I think this is a much cleaner solution! I will test it out and give feedback (I realize this is kind of a pointless comment, but I want you to know there is interest!)

chat.go
```diff
@@ -61,40 +62,8 @@ type ChatCompletionRequest struct {
 type FunctionDefine struct {
 	Name        string `json:"name"`
 	Description string `json:"description,omitempty"`
-	// it's required in function call
-	Parameters *FunctionParams `json:"parameters"`
```
vvatanabe, line 65: go-openai keeps compatible. This is breaking changes. So why not add a field like `ParametersRaw`?
Note: You need to implement `MarshalJSON` func in `FunctionDefine` since the "parameters" tag is duplicated.

stillmatic: heh, yeah, was a bit sad to see the release go out before I had a chance to upstream this. I agree it's a bit undesirable to have a breaking change like this, even if the release was only out for a few hours. 

It should be relatively straightforward to implement a custom marshaler here, but I think it's worth having the larger conversation about how this library intends to handle the Parameters field moving forward - will this library implement (and maintain) its own JSONSchemaDefine (and related structs) or not?

If so - then ParametersRaw is the only path forward, and I will write the custom functions. GPT-4 gets us pretty close I think: https://chat.openai.com/share/bec0bb09-41f5-4746-8b21-8e40087509b1

If not - then the cleanest path is to deprecate Parameters as an argument and ask people to use ParametersRaw. That's not the end of the world, but it would be much more intuitive to match the API schema. 

My view is that the JSON Schema definition is too large to reasonably support in this library and is better suited to be handled in a specialized library. The problem with supporting a subset is that once somebody hits an edge case, then they have to completely rewrite their code here to use either the raw bytes or another library's struct that can serialize properly.

jmacwhyte: I have been testing out this PR, and I like it a lot! I think it's a lot smoother and easier to work with, and probably easier for new devs, too (for quick testing you can just paste in a schema as a text string, instead of building one from nested Go structs!)

I think this is the way we should do it.

stillmatic:
I've added back the removed interfaces to the `Params` field and renamed the json.RawMessage field to `ParametersRaw`. I feel like this is an anti-pattern though and greatly complicates the library -- I barely understand what's going on in the marshal/unmarshal functions.  While the marshaled struct is byte compatible with the OpenAI spec, it's quite unclear how we handle unmarshal, as there are two possible fields where that can be set (and if we want to maintain strict compatibility, we also need to maintain this unmarshal function, not just the marshal function). 

The PR now also contains essentially the same fix as #375 and #373 -- my view is that so long as the option to use the structs is around, users will gradually request more and more functionality and it's better to draw a line in the sand sooner. Those 2 PR's also suggest that this code path was already broken for anyone using it, so breaking compatibility this time should not be an issue -- itwas already broken.

jmacwhyte:
@stillmatic I agree with the ease-of-use issues. If there was only a RawMessage field, it's pretty easy for the docs to say "put your schema (marshalled into []byte) here!"

If a struct and the rawmessage are both populated, which takes precedence when the request is submitted to the API? How do we convey that to users of the libray? I'm a big fan of a "single source of truth" instead of having two copies of the same data, just in different formats.

chat.go
nekomeowww: I think it should also be marked as deprecated by prepending `// Deprecated:` at the very last line of comment and stated out that users should go and use `ParametersRaw` instead.

chat.go
```diff
@@ -61,8 +62,58 @@ type ChatCompletionRequest struct {
...
+	// Parameters is a JSONSchema object describing the function. Generate it via another library.
+	Parameters    *FunctionParams `json:"-"`
+	ParametersRaw json.RawMessage `json:"-"`
+}
+
+func (fd FunctionDefine) MarshalJSON() ([]byte, error) {
```
stillmatic, line 70: Could somebody please review this marshal/unmarshal code? I barely understand it, if we are going to keep Parameters and ParametersRaw then this code _needs_ to be correct.

jmacwhyte: It looks okay to me. The custom marshalling is almost exactly the same as [in this article](https://medium.com/picus-security-engineering/custom-json-marshaller-in-go-and-common-pitfalls-c43fa774db05).

The only thing I would change would be to rename `FunctionDefine` to `FunctionDefinition`...

chat.go
```diff
@@ -61,8 +62,61 @@ type ChatCompletionRequest struct {
...
+	// ParametersRaw is a JSONSchema object describing the function.
+	// You can pass a raw byte array describing the schema,
+	// or you can generate the array from a JSONSchema object, using another library.
+	ParametersRaw json.RawMessage `json:"-"`
+	// Deprecated: DO NOT USE. Use ParametersRaw instead.
+	Parameters *FunctionParams `json:"-"`
+}
+
+func (fd FunctionDefine) MarshalJSON() ([]byte, error) {
+	type Alias FunctionDefine
```
vvatanabe, line 74: How about adding a comment indicating that FunctionDefine type aliases are used to prevent recursion during JSON marshalling?

chat.go
```diff
@@ -61,8 +62,61 @@ type ChatCompletionRequest struct {
...
+		*Alias
+		Parameters json.RawMessage `json:"parameters"`
+	}{
+		Alias:      (*Alias)(&fd),
+		Parameters: parameters,
+	})
+}
+
+func (fd *FunctionDefine) UnmarshalJSON(data []byte) error {
+	type Alias FunctionDefine
```
vvatanabe, line 97: How about adding a comment indicating that FunctionDefine type aliases are used to prevent recursion during JSON unmarshalling?

jmacwhyte:
@stillmatic I've submitted a PR to your repo with some improvements (including the comment @vvatanabe suggested!)

stillmatic:
thanks @jmacwhyte ! my only issue with that is that `FunctionDefine -> FunctionDefinition` also breaks compatibility :)

jmacwhyte:
> thanks @jmacwhyte ! my only issue with that is that `FunctionDefine -> FunctionDefinition` also breaks compatibility :)

How long has `FunctionDefine` even existed? One day? XD

I agree we don't want to break anything, but it's frustrating when poor word choices are rushed out the door... it makes it much harder for people to understand the code in the future.

stillmatic:
That's basically how I feel  - the Parameters implementation with a custom struct adds quite a lot of complexity and is quite hard for future users to understand (and deprecation without removal is also confusing). 

If we're going to break compatibility with the FunctionDefinition change (which I agree is much easier to read), we might as well break compatibility and focus only on the `json.RawMessage` implementation. It's only been around for 2 days, and for 1 of those days, it was broken without the `items` field.

jmacwhyte:
@stillmatic Should we just put together what we think is the optimal solution, and let @sashabaranov (or another maintainer) pull it in if they want? I don't know who is making the decisions around here, I just want a library that makes my project easy to do.

stillmatic:
@sashabaranov - my proposal is that we merge this PR as of https://github.com/stillmatic/go-openai/tree/7bdebae1046d119a139a997fb6c0bf4c90fe63a4 -- that diff converts the `Parameters` object from a custom JSON Schema implementation to a `json.RawMessage` object, which is more generally supported and relieves the library maintainers from the need to implement another JSON Schema definition. It's closer to what OpenAI expects and much easier to work with using external libraries. I've written a testing and working implementation https://github.com/stillmatic/gollum/blob/main/functions.go#L16 using the github.com/invopop/jsonschema library but have confirmed that it is also compatible with the github.com/santhosh-tekuri/jsonschema/v5 library/. Tests and checks also pass completely, as the implementation is extremely simple.

This change is the most compatible with the rest of the golang ecosystem and the least confusing for users - a single code path with no required dependencies and clear debugging paths. It technically breaks compatibility, but I believe it is not a big deal - as discussed above, the release has not been out for very long and it was bugged most of that time. The future maintenance burden of keeping a deprecated tag around is quite high in this case and should outweigh the disruption to workflows right now.

nekomeowww (approved):

jmacwhyte (changes requested):
Also change FunctionCall.Arguments to type json.RawMessage.

stillmatic:
@jmacwhyte is the change you're referring to covered by https://github.com/sashabaranov/go-openai/pull/388/files?

jmacwhyte:
> @jmacwhyte is the change you're referring to covered by https://github.com/sashabaranov/go-openai/pull/388/files?

No. I explained why in a comment on that PR ;)

sashabaranov:
Folks, thank you so much for the great effort you are putting in here! Please excuse me if I've missed some part of the discussion. Here are a couple of thoughts after briefly studying the discussion:

1. I'm fine with breaking API in this case
2. I can see the appeal of both simplified in-library JSON Schema and `json.RawMessage`. Could we have a type that can take either of them and do the proper thing during marshaling? Or would it break during unmarshalling? If this approach is feasible — let's also move JSON schema stuff into a separate directory.

jmacwhyte:
I wonder if we could change `Parameters` to type `any`, and then during submission:

1. If `Parameters` is a `[]byte` or `string`, treat it like `json.RawMessage`.
2. If it is anything else, run it through `json.Marshal()`.

That way, users can pass in a preformatted schema as text, but they can also use any library that provides nested structs that are JSON compliant (and we don't have to know anything about which libraries they are using).

stillmatic:
I've pushed up a sample commit now that converts `Parameters` to `interface{}` and relies on the JSON package to handle marshaling/unmarshaling here. Tests pass using both `json.RawMessage` and custom structs that serialize to JSON here. This is very similar to @jmacwhyte suggestion but does not require a custom marshaller. 

If you want to add back the JSONSchema implementation, ok, we can add that back, preferably it should note the limitations on it, but I think the PR should be good to go right afterwards.

chat.go
```diff
@@ -54,50 +54,22 @@ type ChatCompletionRequest struct {
...
+type FunctionDefinition struct {
 	Name        string `json:"name"`
 	Description string `json:"description,omitempty"`
-	// it's required in function call
-	Parameters *FunctionParams `json:"parameters"`
+	// Parameters is an object describing the function.
+	// You can pass a raw byte array describing the schema,
+	// or you can generate the array from a JSONSchema object, using another library.
+	// The only constraint is that the input must implement the JSONEncodable interface.
+	Parameters interface{} `json:"parameters"`
```
sashabaranov, line 68: nit: `interface{}` is just `any` now

stillmatic: TIL :) still write `map[string]interface{}` out of habit

chat.go
```diff
@@ -54,50 +54,22 @@ type ChatCompletionRequest struct {
...
-	// it's required in function call
-	Parameters *FunctionParams `json:"parameters"`
+	// Parameters is an object describing the function.
+	// You can pass a raw byte array describing the schema,
+	// or you can generate the array from a JSONSchema object, using another library.
+	// The only constraint is that the input must implement the JSONEncodable interface.
+	Parameters any `json:"parameters"`
 }
 
-type FunctionParams struct {
```
sashabaranov, line 68: I guess we should keep `FunctionParams` struct to provide a default way to specify function parameters? So in README we would recomment doing `Parameters: &FunctionParams{...}` or using your own `json.RawMessage` (as per @jmacwhyte suggestion).

(See my second point in the earlier [comment](https://github.com/sashabaranov/go-openai/pull/377#issuecomment-1596097407))



cc @vvatanabe @jmacwhyte

jmacwhyte:
@sashabaranov Sounds good to me, as long as it's simple and easy to maintain.

@stillmatic I submitted another PR to your fork, it's quite minor but I would like it included before the final merge.

jmacwhyte:
Regarding @sashabaranov 's [point number 2 here](https://github.com/sashabaranov/go-openai/pull/377#issuecomment-1596097407), I have submitted a PR to stillmatic's fork which moves all the json schema stuff to a separate directory.

chat.go
```diff
@@ -54,23 +54,23 @@ type ChatCompletionRequest struct {
...
-	// the Type must be JSONSchemaTypeObject
-	Type       JSONSchemaType               `json:"type"`
-	Properties map[string]*JSONSchemaDefine `json:"properties,omitempty"`
-	Required   []string                     `json:"required,omitempty"`
-}
+// Deprecated: use FunctionDefinition instead.
+type FunctionDefine FunctionDefinition
```
vvatanabe, line 73: This is a defined type; cannot use `FunctionDefine` for `Functions` fields. Compatibility can be maintained by using a type alias as follows:

`type FunctionDefine = FunctionDefinition`

vvatanabe:
@sashabaranov  Sorry for the late reply.

> 1. I'm fine with breaking API in this case.

I agree. Please note: if the API is to be changed in a way that breaks compatibility, it should be announced that these breaking changes will be made in a forthcoming minor update. Subsequently, these changes should be implemented and released in a major update. Make sure to document the breaking changes in the release notes of the major update. [go-github](https://github.com/google/go-github/releases) is a good reference for this process, as it frequently makes breaking changes while following these steps.

> 2. I can see the appeal of both simplified in-library JSON Schema and json.RawMessage. 

I agree. Changing the `Parameters` field to the type any would allow for a minor update to be released. When marshalling the `Parameters` field, we can use type assertions to differentiate `[]byte(json.RawMessage)` or `struct`. The next major update should then support only []byte(json.RawMessage).

stillmatic:
Thanks for the tip on `defined type` vs `type alias` -- with that, I've added a test that covers the old behavior as well. So we should actually have 100% compatibility. 

I haven't added any custom marshaling code since this works now. I can also see the appeal of leaving it as an `any` field i nstead of `json.RawMessage`. This is not the greatest example, but imagine if `msg` here was a properly formed JSONSchema struct, e.g. from a custom JSONSchema library.

image

Then, this code will marshal properly as is. The advantage is that you save a marshal step (instead of marshal struct to get bytes to pass to parameters, then marshal again). The disadvantage is that you add an unmarshal step (need to reflect to get the type). However, most people will never unmarshal this particular `Request` struct as they don't run their own servers, so probably better this way. We can revisit that in the future though :) 

should be good to merge?

jmacwhyte:
Thanks for the comments @vvatanabe , I'm learning a lot from them!

I really like the solution we've come up with to use `any` for the schema. It lets developers use whatever they want, and as far as I can see it's pretty hard to use incorrectly, so there aren't really any downsides.

vvatanabe:
My apologies, I overlooked the TestChatCompletionsFunctions test. There's no need to implement a custom marshalling function. I believe the solution proposed in this PR is both simple and fitting. LGTM !

Thanks to the change to the any type, there's no need to specifically plan for a major update. This change can simply be released as part of a minor update.

PR #382, opened 2023-06-16
audio: add items to AudioResponseFormat enum

romazu:
Added "text" and "verbose_json" output formats.

See reference:
https://platform.openai.com/docs/api-reference/audio/create

### DIFF
--- a/audio.go
+++ b/audio.go
@@ -20,9 +20,11 @@ const (
 type AudioResponseFormat string
 
 const (
-	AudioResponseFormatJSON AudioResponseFormat = "json"
-	AudioResponseFormatSRT  AudioResponseFormat = "srt"
-	AudioResponseFormatVTT  AudioResponseFormat = "vtt"
+	AudioResponseFormatJSON        AudioResponseFormat = "json"
+	AudioResponseFormatText        AudioResponseFormat = "text"
+	AudioResponseFormatSRT         AudioResponseFormat = "srt"
+	AudioResponseFormatVerboseJSON AudioResponseFormat = "verbose_json"
+	AudioResponseFormatVTT         AudioResponseFormat = "vtt"
 )
 
 // AudioRequest represents a request structure for audio API.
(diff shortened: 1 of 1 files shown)


audio.go
```diff
@@ -20,9 +20,11 @@ const (
 type AudioResponseFormat string
 
 const (
-	AudioResponseFormatJSON AudioResponseFormat = "json"
-	AudioResponseFormatSRT  AudioResponseFormat = "srt"
-	AudioResponseFormatVTT  AudioResponseFormat = "vtt"
+	AudioResponseFormatJSON        AudioResponseFormat = "json"
+	AudioResponseFormatText        AudioResponseFormat = "text"
+	AudioResponseFormatSRT         AudioResponseFormat = "srt"
+	AudioResponseFormatVerboseJSON AudioResponseFormat = "verbose_json"
```
vvatanabe, line 26: @romazu What format is `verbose_json`? If it is json, you may need to add it to the conditions of the function below.

```
// HasJSONResponse returns true if the response format is JSON.
func (r AudioRequest) HasJSONResponse() bool {
	return r.Format == "" || r.Format == AudioResponseFormatJSON
}
```
Refs:
https://github.com/sashabaranov/go-openai/blob/master/audio.go#L97-L100

romazu: Yes, you are right, it's a json in the following format:
```
{
  "task": "transcribe",
  "language": "english",
  "duration": 10.0,
  "segments": [
    {
      "id": 0,
      "seek": 0,
      "start": 0.0,
      "end": 10.16,
      "text": " Chapter 1 Looming",
      "tokens": [
        50364,
        18874,
        502,
        6130,
        10539,
        50872
      ],
      "temperature": 0.0,
      "avg_logprob": -0.6817335401262555,
      "compression_ratio": 0.68,
      "no_speech_prob": 0.01032273843884468,
      "transient": false
    }
  ],
  "text": "Chapter 1 Looming"
}
```

I also expanded AudioResponse struct to accommodate for this.

Issue #378, opened 2023-06-15
ChatCompletionRequest.FunctionCall does not support specific functions


chrikoch:
Setting ChatCompletionRequest.FunctionCall to "auto" or "none" is no problem at all, but the OpenAI API will reject the request when using a specific function.

When using this:

```
FunctionCall: `{"name": "getWeather"}`,
```

it is sent like this "on the wire":

```
"function_call":"{\"name\": \"getWeather\"}
```
Obviously "FunctionCall" is intepreted as a string (as it is one), but not as a JSON object. Correct would be this:
```
"function_call":{"name": "getWeather"}
```

zbysir:
This issue will fixed in the PR: 
https://github.com/sashabaranov/go-openai/pull/374

Just wait a while.

MrCher55:
Could you please provide some examples for the Function call part where we are calling our own specific functions within the OpenAI API?
https://platform.openai.com/docs/guides/gpt/function-calling

vvatanabe:
@MrCher55 Your question is beyond the scope of this issue. Could you possibly create a new issue or ask in the [discussions](https://github.com/sashabaranov/go-openai/discussions)?

vvatanabe:
Already fixed #https://github.com/sashabaranov/go-openai/pull/374

PR #374, opened 2023-06-15
fix: function_call can be a string or an object

j178:
According to the document, `function_call` parameter can a string literal: "none", "auto", or an object with a key "name".

https://platform.openai.com/docs/api-reference/chat/create#chat/create-function_call

Also added `FunctionCall` to `ChatCompletionStreamChoiceDelta`.

PR #388, opened 2023-06-17
fix: support objects for function_call in chat completion

ryanskidmore:
This PR is a duplicate of #374, and addresses issue #378. I've copied a majority of the changes from @j178's PR (thank you) and addressed the feedback that was raised in that PR. I've only applied this to the chat completion endpoint as that's what I needed these changes for. Unfortunately there was already a struct named `FunctionCall` so unfortunately had to go with the rather ugly naming of `ChatCompletionRequestFunctionCall`. Happy to hear suggestions for better names!

Happy to close this in favour of @j178's PR if they wish given most of this is their code. I needed this for a project I'm working on, otherwise I would have waited on the other PR.

### DIFF
--- a/examples/completion-with-tool/main.go
+++ b/examples/completion-with-tool/main.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/sashabaranov/go-openai"
+	"github.com/sashabaranov/go-openai/jsonschema"
+)
+
+func main() {
+	ctx := context.Background()
+	client := openai.NewClient(os.Getenv("OPENAI_API_KEY"))
+
+	// describe the function & its inputs
+	params := jsonschema.Definition{
+		Type: jsonschema.Object,
+		Properties: map[string]jsonschema.Definition{
+			"location": {
+				Type:        jsonschema.String,
+				Description: "The city and state, e.g. San Francisco, CA",
+			},
+			"unit": {
+				Type: jsonschema.String,
+				Enum: []string{"celsius", "fahrenheit"},
+			},
+		},
+		Required: []string{"location"},
+	}
+	f := openai.FunctionDefinition{
+		Name:        "get_current_weather",
+		Description: "Get the current weather in a given location",
+		Parameters:  params,
+	}
+	t := openai.Tool{
+		Type:     openai.ToolTypeFunction,
+		Function: f,
+	}
+
+	// simulate user asking a question that requires the function
+	dialogue := []openai.ChatCompletionMessage{
+		{Role: openai.ChatMessageRoleUser, Content: "What is the weather in Boston today?"},
+	}
+	fmt.Printf("Asking OpenAI '%v' and providing it a '%v()' function...\n",
+		dialogue[0].Content, f.Name)
+	resp, err := client.CreateChatCompletion(ctx,
+		openai.ChatCompletionRequest{
+			Model:    openai.GPT4TurboPreview,
+			Messages: dialogue,
+			Tools:    []openai.Tool{t},
+		},
+	)
+	if err != nil || len(resp.Choices) != 1 {
+		fmt.Printf("Completion error: err:%v len(choices):%v\n", err,
+			len(resp.Choices))
+		return
+	}
+	msg := resp.Choices[0].Message
+	if len(msg.ToolCalls) != 1 {
+		fmt.Printf("Completion error: len(toolcalls): %v\n", len(msg.ToolCalls))
+		return
+	}
+
+	// simulate calling the function & responding to OpenAI
+	dialogue = append(dialogue, msg)
+	fmt.Printf("OpenAI called us back wanting to invoke our function '%v' with params '%v'\n",
+		msg.ToolCalls[0].Function.Name, msg.ToolCalls[0].Function.Arguments)
+	dialogue = append(dialogue, openai.ChatCompletionMessage{
+		Role:       openai.ChatMessageRoleTool,
+		Content:    "Sunny and 80 degrees.",
+		Name:       msg.ToolCalls[0].Function.Name,
+		ToolCallID: msg.ToolCalls[0].ID,
+	})
+	fmt.Printf("Sending OpenAI our '%v()' function's response and requesting the reply to the original question...\n",
+		f.Name)
+	resp, err = client.CreateChatCompletion(ctx,
+		openai.ChatCompletionRequest{
+			Model:    openai.GPT4TurboPreview,
+			Messages: dialogue,
+			Tools:    []openai.Tool{t},
+		},
+	)
+	if err != nil || len(resp.Choices) != 1 {
+		fmt.Printf("2nd completion error: err:%v len(choices):%v\n", err,
+			len(resp.Choices))
+		return
+	}
+
+	// display OpenAI's response to the original question utilizing our function
+	msg = resp.Choices[0].Message
+	fmt.Printf("OpenAI answered the original request with: %v\n",
+		msg.Content)
+}


jmacwhyte:
This is incorrect. There are two different properties called "function_call" in the API:

1. When making a request. ChatCompletionRequest.FunctionCall is what is being discussed in the issues you mentioned, and can be either "none", "auto", or an object with a "name" property specifying which function you want to force the AI to call.
2. When the AI responds to a request, and decides to call a function you have defined. In this instance, ChatCompletionMessage.FunctionCall will contain an object of the following format:

`{
Name: [function name],
Arguments: {[argument name]: [argument value], ...},
}
`

Your PR appears to treat number 2 as if it is number 1.

ryanskidmore:
> This is incorrect. There are two different properties called "function_call" in the API:
> 
> 1. When making a request. ChatCompletionRequest.FunctionCall is what is being discussed in the issues you mentioned, and can be either "none", "auto", or an object with a "name" property specifying which function you want to force the AI to call.
> 2. When the AI responds to a request, and decides to call a function you have defined. In this instance, FinishReason will be set to `function_call` and ChatCompletionMessage.FunctionCall will contain an object of the following format:
> 
> ```
> {
> Name: [function name],
> Arguments: {[argument name]: [argument value], ...},
> }
> ```
> 
> Your PR appears to treat number 2 as if it is number 1.

@jmacwhyte I think you may have misread the code. This PR adds a `FunctionCall` field to the `ChatCompletionRequest` struct - addressing (1). This PR does not make any changes related to (2), which is already functional.

jmacwhyte:
> @jmacwhyte I think you may have misread the code. This PR adds a `FunctionCall` field to the `ChatCompletionRequest` struct - addressing (1). This PR does not make any changes related to (2), which is already functional.

Crap, you're right. It looked like the FunctionCall you were changing was the one inside of ChatCompletionMessage, because the code in between was hidden. My mistake!

jmacwhyte:
I've look at it again (correctly, this time) and I think this is a very good solution. Nice job!

ryanskidmore:
> I've look at it again (correctly, this time) and I think this is a very good solution. Nice job!

Thanks! I can't take much (or really, any) credit for it though 😅

chat.go
```diff
@@ -55,7 +56,26 @@ type ChatCompletionRequest struct {
 	LogitBias        map[string]int          `json:"logit_bias,omitempty"`
 	User             string                  `json:"user,omitempty"`
 	Functions        []*FunctionDefine       `json:"functions,omitempty"`
-	FunctionCall     string                  `json:"function_call,omitempty"`
+	// FunctionCall can either be a string ("none" or "auto") or an instance of ChatCompletionRequestFunctionCall
+	FunctionCall any `json:"function_call,omitempty"`
+}
+
+func checkFunctionCall(value any) bool {
```
sashabaranov, line 63: Hey, thank you for the PR! The name of this function sounds a bit too vague. Maybe let's go with something a bit more specific like `checkIsChatCompletionFunctionCall`?

ryanskidmore: @sashabaranov Good idea - I've just pushed a change to rename the function to exactly that.

steeve:
I have another take at this, that is type safe and without reflection if you folks want:

```go
var (
	FunctionCallNone = &FunctionCall{}
	FunctionCallAuto = &FunctionCall{}
)

func (f *FunctionCall) MarshalJSON() ([]byte, error) {
	switch f {
	case FunctionCallNone:
		return []byte(`"none"`), nil
	case FunctionCallAuto:
		return []byte(`"auto"`), nil
	default:
	}
	type functionCallAlias FunctionCall
	return json.Marshal((*functionCallAlias)(f))
}
```

```go
	FunctionCall     *FunctionCall           `json:"function_call,omitempty"`
```

tartavull:
What's missing to get this merged? :)

sashabaranov:
Okay, we have a conflict here now based on work done in #377 and #407

@ryanskidmore I love the additional testing and type-checking. Would it be possible to resolve conflicts with the latest updates? Thanks!

chat.go
```diff
@@ -55,7 +56,26 @@ type ChatCompletionRequest struct {
 	LogitBias        map[string]int          `json:"logit_bias,omitempty"`
 	User             string                  `json:"user,omitempty"`
 	Functions        []*FunctionDefine       `json:"functions,omitempty"`
-	FunctionCall     string                  `json:"function_call,omitempty"`
+	// FunctionCall can either be a string ("none" or "auto") or an instance of ChatCompletionRequestFunctionCall
+	FunctionCall any `json:"function_call,omitempty"`
+}
+
+func checkIsChatCompletionFunctionCall(value any) bool {
```
jmacwhyte, line 63: Should this maybe be named something like `validateChatCompletionFunctionCall`? The current name makes me think a return of `true` would mean it is a function call, but `true` means it is a valid value (even if there is no function call involved).

jmacwhyte:
I also think @steeve 's solution is very clever (and compact), how do we decide what is more maintainable?

ryanskidmore:
@sashabaranov (cc @jmacwhyte @tartavull @steeve & any others interested) I've rebased my fork onto master and resolved the outstanding conflicts. With the recent changes that were made in other PRs I took a bit of a different direction which I appreciate may conflict with intentional decisions made in other PRs - so please feel free to give me feedback on those directional changes.

The `FunctionCall` field moving to be an `any` type made the usage very non-intuitive and confusing (from the perspective of someone who uses this library). You have to read the documentation to understand what to put in the `FunctionCall` field which IMO isn't great from a design perspective. I've moved this field back to being a pointer to the `FunctionCall` struct and adopted @steeve's novel approach which removes the need for this field to be an `any` type. I've changed the validation to match this change accordingly which ensures the field is populated with a valid value.

chat.go
```diff
@@ -34,12 +36,37 @@ type ChatCompletionMessage struct {
...
+	case FunctionCallAuto:
+		return []byte(`"auto"`), nil
+	default:
+	}
+	type functionCallAlias FunctionCall
+	return json.Marshal((*functionCallAlias)(f))
+}
+
+func validateChatCompletionFunctionCall(v *FunctionCall) bool {
+	if v == nil || v == FunctionCallNone || v == FunctionCallAuto || v.Name != "" {
```
jmacwhyte, line 63: Now that this is so simple (one line), instead of making it a separate function could you just put it inline at line number 146 below?

vvatanabe: I kindly suggest we keep this as a function. The reasons being, it promotes enhances readability, facilitates testing, and simplifies refactoring. I believe these are crucial elements in preserving the quality of our code.

jmacwhyte: I'm good with that!

jmacwhyte:
I think it all looks great!

chat.go
```diff
@@ -34,12 +36,37 @@ type ChatCompletionMessage struct {
...
+func (f *FunctionCall) MarshalJSON() ([]byte, error) {
+	switch f {
+	case FunctionCallNone:
+		return []byte(`"none"`), nil
+	case FunctionCallAuto:
+		return []byte(`"auto"`), nil
+	default:
+	}
+	type functionCallAlias FunctionCall
+	return json.Marshal((*functionCallAlias)(f))
```
vvatanabe, line 59: Developers unfamiliar with Go may be a bit confused, so please add a comment that the use of type aliases prevents recursive calls.

arrix: Learned something from this comment 👍. It would be helpful to clarify this is a type definition rather than an alias. In go's terms, an alias is an identifier bound to a type so it receives all the methods from it too.

sullrich:
Hoping we can keep this alive, this is needed with modern ai api calls.  Can someone review + commit?

jmacwhyte:
I just assumed this was accepted 6 months ago, and haven't looked at it much since! @sashabaranov can you update us on the latest?

ryanskidmore:
I'm going to close this for now given the number of changes that have happened in this code over the last ~6 months. These fields have been deprecated now so I don't see a need to revive this.

PR #390, opened 2023-06-18
feat: added function call info to chat completions

cem-unuvar:
Function calls are currently inaccessible in chat completion streams. This addresses that issue.

### DIFF
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -8,8 +8,9 @@ import (
 )
 
 type ChatCompletionStreamChoiceDelta struct {
-	Content string `json:"content,omitempty"`
-	Role    string `json:"role,omitempty"`
+	Content      string        `json:"content,omitempty"`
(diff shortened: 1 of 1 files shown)


jmacwhyte:
@cem-unuvar Have you been able to get function calls to work when streaming from the API? In my tests, it just returned an empty response and no function call information. I had assumed function calls aren't supported when the `streaming` flag is set to true, but maybe I missed something?

vvatanabe:
FYI
https://github.com/sashabaranov/go-openai/pull/374#issuecomment-1593715129

cem-unuvar:
@jmacwhyte I am running this in prod right now without any issues

jmacwhyte:
I was able to get it to work! I agree with merging this.

giefferre (approved):

sashabaranov (approved):
Thank you for the PR!

PR #394, opened 2023-06-20
CI

NoahAmethyst:
-- add sync workflow to auto fetch upstream updates


vvatanabe:
@NoahAmethyst  Is this GitHub Action for periodically synchronizing the fork repository with the origin?

NoahAmethyst:
> @NoahAmethyst Is this GitHub Action for periodically synchronizing the fork repository with the origin?

yep

NoahAmethyst:
> @NoahAmethyst Is this GitHub Action for periodically synchronizing the fork repository with the origin?

you can check out the [workflow log](https://github.com/NoahAmethyst/go-openai/actions/runs/5317576569/jobs/9628240103) from my repo

vvatanabe:
@NoahAmethyst Does this action automatically take effect for the forked repository? Does the owner of the forked repository need to add any settings?

NoahAmethyst:
> Does this action automatically take effect for the forked repository? Does the owner of the forked repository need to add any settings?

Of course it  automatically take effect for the forked repository,they don't need to set anything except they have their own requirements like merging upstream master into forked repo's specific branch,they need change the script in this case

vvatanabe:
> Of course it automatically take effect for the forked repository

@NoahAmethyst   This action is very useful. Excellent work! However, I believe there are some concerns related to this action:

- If it is not up-to-date when you begin work, manual synchronization is required.
- Some users or organizations that have forked may prefer that the master branch not be automatically updated.
- If the forked repository is private, it could inadvertently consume credits.

NoahAmethyst:
> > Of course it automatically take effect for the forked repository
> 
> @NoahAmethyst This action is very useful. Excellent work! However, I believe there are some concerns related to this action:
> 
> * If it is not up-to-date when you begin work, manual synchronization is required.
> * Some users or organizations that have forked may prefer that the master branch not be automatically updated.
> * If the forked repository is private, it could inadvertently consume credits.

You're very thoughtful,i did not consider these cases, thank you for reminding.It's useful in some forked repo but not suitable at the origin.

PR #395, opened 2023-06-20
refactor: refactoring http request creation and sending

vvatanabe:
`http.Request` creation and sending were duplicated, so they have been aggregated into a common function.

- `RequestBuilder#Build` function calls were aggregated into the `Client#newRequest` function.
- All public functions build `http.Request` in the `Client#newRequest` function.
- The sending of `http.Request` is now aggregated into the following three functions.
  - `Client#sendRequest` function: marshalling the response into a structure
  - `Client#sendRequestRaw` function: returns the Body of the response as is 
  - `Client#sendRequestStream` function: use generics to specify a `streamable` type.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1,6 +1,7 @@
 package openai
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -45,6 +46,42 @@ func NewOrgClient(authToken, org string) *Client {
 	return NewClientWithConfig(config)
 }
 
+type requestOptions struct {
+	body   any
+	header http.Header
+}
+
+type requestOption func(*requestOptions)
+
+func withBody(body any) requestOption {
+	return func(args *requestOptions) {
+		args.body = body
+	}
+}
+
+func withContentType(contentType string) requestOption {
+	return func(args *requestOptions) {
+		args.header.Set("Content-Type", contentType)
+	}
+}
+
+func (c *Client) newRequest(ctx context.Context, method, url string, setters ...requestOption) (*http.Request, error) {
+	// Default Options
+	args := &requestOptions{
+		body:   nil,
+		header: make(http.Header),
+	}
+	for _, setter := range setters {
+		setter(args)
+	}
+	req, err := c.requestBuilder.Build(ctx, method, url, args.body, args.header)
+	if err != nil {
+		return nil, err
+	}
+	c.setCommonHeaders(req)
+	return req, nil
+}
+
 func (c *Client) sendRequest(req *http.Request, v any) error {
 	req.Header.Set("Accept", "application/json; charset=utf-8")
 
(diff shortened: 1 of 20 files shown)


vvatanabe:

sashabaranov:
This is super cool! 🚀

internal/request_builder.go
```diff
@@ -20,21 +21,27 @@ func NewRequestBuilder() *HTTPRequestBuilder {
 	}
 }
 
-func (b *HTTPRequestBuilder) Build(ctx context.Context, method, url string, request any) (*http.Request, error) {
-	if request == nil {
-		return http.NewRequestWithContext(ctx, method, url, nil)
+func (b *HTTPRequestBuilder) Build(ctx context.Context,
+	method, url string, body any, header http.Header) (req *http.Request, err error) {
```
sashabaranov, line 25: nit: maybe let's go one-argument-per-line style here, it's a bit longer but seems cleaner. What do you think?

```
func (b *HTTPRequestBuilder) Build(
	ctx context.Context,
	method,
	url string,
	body any,
	header http.Header
)
```

models_test.go
```diff
@@ -56,3 +59,22 @@ func handleGetModelEndpoint(w http.ResponseWriter, _ *http.Request) {
...
 }
+
+func TestGetModelReturnTimeoutError(t *testing.T) {
+	client, server, teardown := setupOpenAITestServer()
+	defer teardown()
+	server.RegisterHandler("/v1/models/text-davinci-003", func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(10 * time.Nanosecond)
+	})
+	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(ctx, time.Nanosecond)
```
sashabaranov, line 70: Wow, that's some bleeding edge timeout! 😄

sashabaranov (approved):
This is a really beautiful PR!

internal/request_builder.go
vvatanabe: @sashabaranov 
> one-argument-per-line style

Looks Good! I will update to this style.

models_test.go
vvatanabe: @sashabaranov I didn't want to spend time running tests.;D

Here is some background on this timeout test. 
`sendRequest`, `sendRequestRaw`, and `sendRequestStream` each exec `c.config.HTTPClient.Do(req)` internally. This error handling was not covered test. That is why we implemented a workaround for `c.config.HTTPClient.Do(req)` with a timeout, to intentionally cause it to return an error.

It's hard to convey the intent of the test, isn't it? Would it be better to write a test that executes `sendRequest`, `sendRequestRaw`, and `sendRequestStream` directly?

client.go
```diff
@@ -55,8 +92,6 @@ func (c *Client) sendRequest(req *http.Request, v any) error {
 		req.Header.Set("Content-Type", "application/json; charset=utf-8")
 	}
 
-	c.setCommonHeaders(req)
-
 	res, err := c.config.HTTPClient.Do(req)
 	if err != nil {
 		return err
```
vvatanabe, line 97: I wanted to cover the above line.
https://github.com/sashabaranov/go-openai/pull/395#discussion_r1237604960

internal/request_builder.go
vvatanabe: DONE!

sashabaranov (approved):

PR #396, opened 2023-06-20
fix: chat stream returns an error response with a 'data: ' prefix

liushuangls:
Occasionally, OpenAI may return the following error message:
```
data: {"error":{"message":"The server experienced an error while processing your request. We apologize for the inconvenience!", "type":"server_error", "param":null,"code":null)}

data: [DONE]
```
In such cases, the stream.Recv() function will return a successful status along with a zero-value response.


sashabaranov (approved):
Awesome, thank you!

liushuangls:
@sashabaranov Please release it.

pmilanez:
+1 for release it, i'm facing this problem and not been able to retry it based on the error

PR #402, opened 2023-06-22
add vvatanabe to FUNDING.yml

vvatanabe:
Hey @sashabaranov, my GitHub Sponsors application got approved! So, as we talked about before, I've made a PR to add myself to the FUNDING.yml. Check it out when you can!


sashabaranov (approved):
🙌

PR #407, opened 2023-06-23
move json schema to directory/package

jmacwhyte:
@sashabaranov When discussing #377 you made [a comment](https://github.com/sashabaranov/go-openai/pull/377#issuecomment-1596097407) about separating out the json schema definitions, which [I did in a subsequent commit](https://github.com/sashabaranov/go-openai/pull/377#issuecomment-1597897146). Although it was added to @stillmatic 's fork before #377 was accepted, it looks like that commit didn't make it into the final merge!

I am including it here for consideration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,6 +517,66 @@ func main() {
 </details>
 
 <details>
+<summary>JSON Schema for function calling</summary>
+
+It is now possible for chat completion to choose to call a function for more information ([see developer docs here](https://platform.openai.com/docs/guides/gpt/function-calling)).
+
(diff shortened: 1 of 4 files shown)


sashabaranov:
@jmacwhyte thank you for the PR! Would you be up to adding a README example while we are on it?

sashabaranov (approved):

stillmatic:
oops, it got clobbered in a rebase + force push :| 

fwiw it does break the import path of old code (hence updating the tests). I am fine with it but should make a note!

jmacwhyte:
@sashabaranov I've added some details to the README, please let me know if that looks sufficient!

@stillmatic I'm not really sure what details to include about import path breaking, but would be happy to add something if you have a suggestion.

PR #412, opened 2023-06-26
chore: add a pull request template

vvatanabe:
@sashabaranov I've created a pull request template for bug reports. Could you please review the content of the template?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,0 +1,27 @@
+---
+name: Pull Request
+about: Propose changes to the codebase
+title: ''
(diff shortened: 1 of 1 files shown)


vvatanabe:
@sashabaranov 
How about prefixing the title of pull requests that contain breaking changes with [BREAKING_CHANGES]? This would make it immediately clear on the release page under the 'What's Changed' section that a breaking change has been introduced. (I presume the release page is auto-generated by a tool, right?) This approach is very developer-friendly.

Here's an example (Note that no actual changes have been made).

image

If there are no objections, I want to incorporate that rule into this pull request template.

sashabaranov (approved):
Thank you so much! I agree with the `[BREAKING_CHANGES]` suggestion, it would definitely help!

Issue #429, opened 2023-07-08
Function Call Error Due to Nil Properties


vvatanabe:
**Describe the bug**
An error occurs when using the Function Call feature. The relevant error message is as follows:
```
Completion error: error, status code: 400, message: Invalid schema for function 'func_name': None is not of type 'object'
```
This issue arises when the map in the Properties field of jsonschema.Definition is set to nil.

**To Reproduce**
This bug can be reproduced through the following steps:

1. Use jsonschema.Definition when constructing parameters for the Function Call.
2. Set the map in the Properties field of jsonschema.Definition to nil.
3. Send a request using the Function Call feature.

Please note that the process works correctly if the map in the Properties field of jsonschema.Definition is initialized.

PR for reproduction:
https://github.com/sashabaranov/go-openai/pull/430

**Expected behavior**
Even if the map in the Properties field of jsonschema.Definition is nil, it should be possible to send a request using the Function Call feature.

**Environment**
go-openai version: v1.13.0

**Additional context**
This bug was caused by changes made in pull request #419. Specifically, omitempty was removed from the Properties field of jsonschema.Definition by pull request #419. This change caused issues when the map in the Properties field was set to nil.

For detailed reasons why omitempty was removed, please refer to the description in pull request #419.

As a bug handling measure, it seems better to initialize the map in the Properties field of jsonschema.Definition to an empty map if it is set to nil. Also, since jsonschema.Definition has a recursive structure that can have jsonschema.Definition as a child, the check for nil and initialization should also be done recursively.

vvatanabe:
Fixed #431

vvatanabe:
The problem recurs when Parameters is a value instead of a pointer.
Refs: https://github.com/sashabaranov/go-openai/issues/432

vvatanabe:
Fixed https://github.com/sashabaranov/go-openai/pull/434

PR #431, opened 2023-07-08
fix: function call error due to nil properties (429)

vvatanabe:
**Describe the change**
This PR solves issue #429.
PR to reproduce the problem: #430 

**Describe your solution**
If the map in the Properties field of jsonschema.Definition is set to nil, initialize it to an empty map. Also, as jsonschema.Definition has a recursive structure that can contain jsonschema.Definition as a child, perform the nil check and initialization recursively.

**Tests**
I have tested that the Properties field is initialized with an empty map in all of the following cases.
- Empty Definition
- Definition properties set
- Nested Definition properties
- Complex nested Definition
- Array type Definition

**Additional context**
Issue: #429

### DIFF
--- a/jsonschema/json.go
+++ b/jsonschema/json.go
@@ -50,10 +50,9 @@ func (d *Definition) initializeProperties() {
 	for k, v := range d.Properties {
 		if v.Properties == nil {
 			v.Properties = make(map[string]Definition)
-			d.Properties[k] = v
 		} else {
 			v.initializeProperties()
-			d.Properties[k] = v
 		}
+		d.Properties[k] = v
 	}
 }
--- a/jsonschema/json.go
+++ b/jsonschema/json.go
@@ -4,6 +4,8 @@
 // and/or pass in the schema in []byte format.
 package jsonschema
 
+import "encoding/json"
+
 type DataType string
 
 const (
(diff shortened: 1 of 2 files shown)


vvatanabe:
It appears that the Function Call of the OpenAI API tolerates an empty object in the properties field but does not tolerate nil.

However, I question whether there really is a use case where the properties field in the root Definition is left empty. If there is no realistic use case for this, it might be better to re-add omitempty to the properties field.

jsonschema/json.go
```diff
@@ -33,3 +35,25 @@ type Definition struct {
...
+	}
+
+	for k, v := range d.Properties {
+		if v.Properties == nil {
+			v.Properties = make(map[string]Definition)
+			d.Properties[k] = v
+		} else {
+			v.initializeProperties()
+			d.Properties[k] = v
+		}
```
sashabaranov, line 57: nit:

```go
if v.Properties == nil {
	v.Properties = make(map[string]Definition)
} else {
	v.initializeProperties()
}

d.Properties[k] = v
```

sashabaranov (approved):
Good catch, thank you!

jsonschema/json.go
vvatanabe: @sashabaranov Oops! Thanks for the review.

PR #433, opened 2023-07-10
fix error 400 during function call

mehulgohil:
**Describe the change**
Added omitempty as json tag for properties field

**Describe your solution**
Was getting error 400 while calling open api function call. So have added omitempty json tag. As properties was going as null

**Tests**
Fixed tests accordingly

Issue:  #432

Issue #432, opened 2023-07-10
 Invalid schema for function 'get_current_weather': None is not of type 'object'


mehulgohil:
**Describe the bug**
When calling  CreateChatCompletion with function call getting error 400.
ChatCompletion error: error, status code: 400, message: Invalid schema for function 'get_current_weather': None is not of type 'object'

**Expected behavior**
Expected behaviorwas to return the function call definition

**Screenshots/Logs**
<img width="354" alt="image" src="https://github.com/sashabaranov/go-openai/assets/28588619/7f8639e3-b394-4d46-8160-8d7def74b92a">

**Environment (please complete the following information):**
 - go-openai version: [e.g. v1.12.0]
 - Go version: [e.g. 1.18]
 - OpenAI API version: [e.g. v1]
 - OS: [e.g. Ubuntu 20.04]

vvatanabe:
Duplicate #429

vvatanabe:
Fixed https://github.com/sashabaranov/go-openai/pull/434

PR #434, opened 2023-07-10
fix: invalid schema for function 'func_name': None is not of type 'object' (#429)(#432)

vvatanabe:
**Describe the change**
The problem recurs when Parameters is a value instead of a pointer. MarshalJSON in Definition does not execute MarshalJSON if Definition is a value because the receiver is a pointer.
```
func (d *Definition) MarshalJSON() ([]byte, error)
```
Refs:
- https://github.com/sashabaranov/go-openai/pull/433#issuecomment-1629401793
- https://github.com/sashabaranov/go-openai/pull/433#issuecomment-1629430142

error message:
```
 invalid schema for function 'func_name': None is not of type 'object' 
```

request to reproduce:
```
	req := openai.ChatCompletionRequest{
		Model: openai.GPT3Dot5Turbo,
			{
				Content: "What is the weather like in Boston?",
			},
		},
		Functions: []openai.FunctionDefinition{{
			Name: "get_current_weather",
			Parameters: jsonschema.Definition{ // <= The problem recurs when Parameters is a value instead of a pointer.
				Type: jsonschema.Object,
				Properties: map[string]jsonschema.Definition{
					"location": {
						Type:        jsonschema.String,
						Description: "The city and state, e.g. San Francisco, CA",
					},
					"unit": {
						Type: jsonschema.String,
						Enum: []string{"celsius", "fahrenheit"},
					},
				},
				Required: []string{"location"},
			},
		}},
	}
```

**Describe your solution**
- MarshalJSON method is executed whether jsonschema.Definition in Parameters is a pointer or a value
- Initialization with an empty map if Properties is nil.

**Tests**
- Test MarshalJSON with both value and pointer.
- add integration test for function call.

**Additional context**
Issue: #429, #432

### DIFF
--- a/api_integration_test.go
+++ b/api_integration_test.go
@@ -9,8 +9,10 @@ import (
 	"os"
 	"testing"
 
+	"github.com/sashabaranov/go-openai"
 	. "github.com/sashabaranov/go-openai"
 	"github.com/sashabaranov/go-openai/internal/test/checks"
+	"github.com/sashabaranov/go-openai/jsonschema"
 )
 
 func TestAPI(t *testing.T) {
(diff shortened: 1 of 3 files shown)


vvatanabe:

sashabaranov (approved):
Thank you for the PR!

PR #437, opened 2023-07-12
Check if the model param is valid for moderations endpoint

MunaerYesiyan:
Currently there are only [two models available](https://platform.openai.com/docs/api-reference/moderations/create) for OpenAI /moderations endpoint, which are `text-moderation-stable` and `text-moderation-latest`. In the official python library `openai-python`. The method Moderation will check if the model param is in `VALID_MODEL_NAMES` and returns an error when it's not. 

```python
 def _prepare_create(cls, input, model, api_key):
        if model is not None and model not in cls.VALID_MODEL_NAMES:
            raise ValueError(
                f"The parameter model should be chosen from {cls.VALID_MODEL_NAMES} "
                f"and it is default to be None."
            )

        instance = cls(api_key=api_key)
        params = {"input": input}
        if model is not None:
            params["model"] = model
        return instance, params
```

It would be great if we can add similar checks before passing the request to `/v1/moderations` endpoint.

### DIFF
--- a/moderation.go
+++ b/moderation.go
@@ -2,6 +2,7 @@ package openai
 
 import (
 	"context"
+	"errors"
 	"net/http"
 )
 
@@ -17,6 +18,15 @@ const (
 	ModerationTextLatest = "text-moderation-latest"
 	ModerationText001    = "text-moderation-001"
 )
+
+var (
+	ErrModerationInvalidModel = errors.New("this model is not supported with moderation, please use text-moderation-stable or text-moderation-latest instead") //nolint:lll
+)
+
+var validModerationModel = map[string]struct{}{
+	ModerationTextStable: {},
+	ModerationTextLatest: {},
+}
 
 // ModerationRequest represents a request structure for moderation API.
 type ModerationRequest struct {
@@ -63,6 +73,10 @@ type ModerationResponse struct {
 // Moderations — perform a moderation api call over a string.
 // Input can be an array or slice but a string will reduce the complexity.
 func (c *Client) Moderations(ctx context.Context, request ModerationRequest) (response ModerationResponse, err error) {
+	if _, ok := validModerationModel[request.Model]; len(request.Model) > 0 && !ok {
+		err = ErrModerationInvalidModel
+		return
+	}
 	req, err := c.newRequest(ctx, http.MethodPost, c.fullURL("/moderations", request.Model), withBody(&request))
 	if err != nil {
 		return
--- a/moderation.go
+++ b/moderation.go
@@ -16,7 +16,8 @@ import (
 const (
 	ModerationTextStable = "text-moderation-stable"
 	ModerationTextLatest = "text-moderation-latest"
-	ModerationText001    = "text-moderation-001"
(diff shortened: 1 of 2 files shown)


moderation_test.go
```diff
@@ -27,6 +27,18 @@ func TestModerations(t *testing.T) {
...
+func TestModerationsWithIncorrectModel(t *testing.T) {
+	client, server, teardown := setupOpenAITestServer()
+	defer teardown()
+	server.RegisterHandler("/v1/moderations", handleModerationEndpoint)
+	_, err := client.Moderations(context.Background(), ModerationRequest{
+		Model: GPT3Dot5Turbo,
+		Input: "I want to kill them.",
+	})
+	checks.ErrorIs(t, err, ErrModerationInvalidModel)
+}
```
vvatanabe, line 40: @MunaerYesiyan Thanks for PR! How about trying to write exhaustive tests for both valid and invalid models in table driven tests?

MunaerYesiyan: https://github.com/sashabaranov/go-openai/pull/437/commits/e1ae0560840e0df535a9fe27b790801280bdefe6

vvatanabe:
Confirmed not supported just in case. 👍 
```
$ curl https://api.openai.com/v1/moderations \
                    -H "Content-Type: application/json" \
                    -H "Authorization: Bearer $OPENAI_API_KEY" \
                    -d '{
                  "input": "I want to kill them.", "model": "text-moderation-001"
                }'
{

  "error": {
    "message": "Invalid value for 'model' = text-moderation-001. Please check the OpenAI documentation and try again.",
    "type": "invalid_request_error",
    "param": "model",
    "code": null
  }
}
```

MunaerYesiyan:
> Confirmed not supported just in case. 👍
> 
> ```
> $ curl https://api.openai.com/v1/moderations \
>                     -H "Content-Type: application/json" \
>                     -H "Authorization: Bearer $OPENAI_API_KEY" \
>                     -d '{
>                   "input": "I want to kill them.", "model": "text-moderation-001"
>                 }'
> {
> 
>   "error": {
>     "message": "Invalid value for 'model' = text-moderation-001. Please check the OpenAI documentation and try again.",
>     "type": "invalid_request_error",
>     "param": "model",
>     "code": null
>   }
> }
> ```

And it's kinda confusing as the official document uses `text-moderation-005` in the response [example ](https://platform.openai.com/docs/api-reference/moderations/create)

moderation.go
```diff
@@ -18,6 +19,15 @@ const (
 	ModerationText001    = "text-moderation-001"
```
vvatanabe, line 19: @MunaerYesiyan I would like you to add a comment indicating that this constant is deprecated.

MunaerYesiyan: Deprecating the `text-moderation-001` model in `moderation.go` https://github.com/sashabaranov/go-openai/pull/437/commits/e579559797ea9199d0bd902248c58830a4a23283

vvatanabe (approved):

vvatanabe:
@MunaerYesiyan Thanks for the pull request!

PR #440, opened 2023-07-14
EscapeString

rhinogo:
The official Python package, when converting objects to JSON strings, will convert Chinese and other text into Unicode encoding, and can effectively avoid sensitive words in responses. This modification is to enable the Golang package to have this capability.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"strconv"
 )
 
 // Chat message role defined by the OpenAI API.
@@ -21,9 +22,15 @@ var (
 	ErrChatCompletionStreamNotSupported = errors.New("streaming is not supported with this method, please use CreateChatCompletionStream")              //nolint:lll
 )
 
+type EscapeString string
+
+func (esc EscapeString) MarshalJSON() ([]byte, error) {
+	return []byte(strconv.QuoteToASCII(string(esc))), nil
+}
+
 type ChatCompletionMessage struct {
-	Role    string `json:"role"`
-	Content string `json:"content"`
+	Role    string       `json:"role"`
+	Content EscapeString `json:"content"`
 
 	// This property isn't in the official documentation, but it's in
 	// the documentation for the official library for python:
--- a/chat_test.go
+++ b/chat_test.go
@@ -268,12 +268,12 @@ func handleChatCompletionEndpoint(w http.ResponseWriter, r *http.Request) {
 		res.Choices = append(res.Choices, ChatCompletionChoice{
 			Message: ChatCompletionMessage{
 				Role:    ChatMessageRoleAssistant,
-				Content: completionStr,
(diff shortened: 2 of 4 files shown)


chat.go
```diff
@@ -21,9 +22,15 @@ var (
...
+
+func (esc EscapeString) MarshalJSON() ([]byte, error) {
+	return []byte(strconv.QuoteToASCII(string(esc))), nil
+}
+
 type ChatCompletionMessage struct {
-	Role    string `json:"role"`
-	Content string `json:"content"`
+	Role    string       `json:"role"`
+	Content EscapeString `json:"content"`
```
vvatanabe, line 33: Can you adopt a method that doesn't involve breaking changes?

sashabaranov:
Hey, thank you for the PR! Could you please provide an illustration where default Go's string handling doesn't work? And what could be our alternatives here?

chat.go
rhinogo: I am not very familiar with other implementation methods.

codingknees:
The response has to be `unquote` ?

```
{chatcmpl-7gTRgwdik36hYBzQKVuFr0wBifCsr chat.completion 1690357320 gpt-35-turbo [{0 {assistant \u54C8\u54C8\uff0c\u6211\u4E5F\u662F\u54EA\u4E2A\u5BF9\u8C61\u7684\u597D\u670B\u53CB\u3002\u5728\u4E3A\u4F60\u6D88\u606F\u65F6\u8BF7\u8BB0\u4F4F\u4EE5\u4E0B\u8981\u70B9\uff1A\n1. \u6211\u7684\u76EE\u7684\u662F\u8BA9\u4F60\u611F\u5230\u5F00\u5FC3\uff0C\u613F\u610F\u548C\u6211\u4E00\u76F4\u804A\u5929\n2. \u5982\u679C\u6211\u8981\u6C42\uFF  <nil>} 
```

codingknees:
when there are Emoji in string, Marshal would fail: 

`json: error calling MarshalJSON for type openai.EscapeString: invalid character 'U' in string escape code`

for example, 😗 will be quoted to ASCII \U0001f617 , but it is not valid in json (encoding/json/scanner.go:353).
Because of the  \U

codingknees:
> when there are Emoji in string, Marshal would fail:
> 
> `json: error calling MarshalJSON for type openai.EscapeString: invalid character 'U' in string escape code`
> 
> for example, 😗 will be quoted to ASCII \U0001f617 , but it is not valid in json (encoding/json/scanner.go:353). Because of the \U

I find that in python, json.dumps encode that Emoji to  \ud83d\ude17. So strange

codingknees:
hack as 
` EscapeString string

func (esc EscapeString) MarshalJSON() ([]byte, error) {
	x := strings.ReplaceAll(strconv.QuoteToASCII(string(esc)), "\\U", "\\u")
	return []byte(x), nil
}`

PR #441, opened 2023-07-15
Update comment for `jsonSchema.FunctionDefinition`

savannahostrowski:
**Describe the change**
Currently, the comment for the `Parameters` property in the `jsonschema.FunctionDefinition` struct says that you can pass in a `[]byte{}`. However, this does not work and you need to use `json.RawMessage()` instead. 

**Describe your solution**
Just updated the comment :)

**Tests**
Not applicable!

**Additional context**
Passing in `[]byte{}` leads to base64'd value
image

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -65,7 +65,7 @@ type FunctionDefinition struct {
 	Name        string `json:"name"`
 	Description string `json:"description,omitempty"`
 	// Parameters is an object describing the function.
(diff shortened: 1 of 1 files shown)


vvatanabe (approved):

vvatanabe:
@savannahostrowski Thanks for the correction of the comment!

PR #449, opened 2023-07-19
feat: make finish reason nullable in json marshal

yuikns:
**Describe the change**

this PR added a `MarshalJSON` to `FinishReason`. 

**Describe your solution**

`MarshalJSON` implemented `Marshaler` interface in package `encoding/json` 

If `FinishReason` is `FinishReasonNull` or empty string, we can render it as `null` instead of `"null"`.

**Tests**

```
c := &ChatCompletionChoice{
	FinishReason: FinishReasonNull,
}
jsonBytes, _ := json.Marshal(c)
fmt.Println(string(jsonBytes)) // it will print { .... "finish_reason": null }
```

Since i am not sure is this PR expected, i didn't add a lot of test cases in *_test.go file in the PR. but it was test locally.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -85,6 +85,13 @@ const (
 	FinishReasonNull          FinishReason = "null"
 )
 
+func (r FinishReason) MarshalJSON() ([]byte, error) {
+	if r == FinishReasonNull || r == "" {
+		return []byte("null"), nil
+	}
+	return []byte(`"` + string(r) + `"`), nil // best effort to not break future API changes
+}
+
 type ChatCompletionChoice struct {
 	Index   int                   `json:"index"`
 	Message ChatCompletionMessage `json:"message"`
--- a/chat_test.go
+++ b/chat_test.go
@@ -298,3 +298,34 @@ func getChatCompletionBody(r *http.Request) (ChatCompletionRequest, error) {
 	}
 	return completion, nil
 }
+
+func TestFinishReason(t *testing.T) {
+	c := &ChatCompletionChoice{
+		FinishReason: FinishReasonNull,
+	}
+	resBytes, _ := json.Marshal(c)
+	if !strings.Contains(string(resBytes), `"finish_reason":null`) {
+		t.Error("null should not be quoted")
+	}
+
+	c.FinishReason = ""
+
+	resBytes, _ = json.Marshal(c)
+	if !strings.Contains(string(resBytes), `"finish_reason":null`) {
+		t.Error("null should not be quoted")
+	}
+
+	otherReasons := []FinishReason{
+		FinishReasonStop,
+		FinishReasonLength,
+		FinishReasonFunctionCall,
+		FinishReasonContentFilter,
+	}
+	for _, r := range otherReasons {
+		c.FinishReason = r
+		resBytes, _ = json.Marshal(c)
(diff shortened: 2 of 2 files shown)


chat.go
```diff
@@ -85,6 +85,17 @@ const (
 	FinishReasonNull          FinishReason = "null"
 )
 
+func (r FinishReason) MarshalJSON() ([]byte, error) {
```
vvatanabe, line 88: @yuikns Thanks for the suggestion. Sorry for the late reply.
The premise is that `FinishReason` is used only in responses. Therefore, there is usually no need to marshal it. Could you tell us in what situations you would like to marshal the response?

yuikns: @vvatanabe 
thank you for your attention.
in some of our projects, it is playing the role of gateway. we need to make sure it will correctly forward the message in the same way.
this string FinishReason broken the assumption.

vvatanabe: @yuikns I see. Understood. Thank you.
The cause of this issue is that the nullable value finish_reason doesn't allow null. Changing the field that uses FinishReason to a pointer might resolve this.

However, turning a primitive string value into a pointer might degrade the developer experience.

Your approach might be a good compromise.

On a different note, currently when finish_reason is null, it gets unmarshalled to an empty string. So, there's no chance that the "null" string indicated by FinishReasonNull is assigned to the FinishReason field, right?

yuikns: @vvatanabe 
exactly, I do not want to alter the API, particularly as changing a value type may cause some breaking change, which is hurting. This implementation could meets my requirements, allowing me to avoid any ugly workarounds.

Thank you for your understanding.

chat.go
```diff
@@ -85,6 +85,17 @@ const (
...
+	switch r {
+	case FinishReasonNull, "":
+		return []byte("null"), nil
+	case FinishReasonStop, FinishReasonLength, FinishReasonFunctionCall, FinishReasonContentFilter:
```
vvatanabe, line 92: @yuikns Is this conditional branching necessary? It seems to do the same thing as the default. How about just having the default? It looks like we might forget to add to this statement if a new FinishReason is added in the future.

yuikns: @vvatanabe it is not necessary.
I was supposed to do some validation here, but i think it is not a good idea.

yuikns: @vvatanabe unfortunately if I just having the default will failed in CI checking. 

ref https://github.com/sashabaranov/go-openai/actions/runs/5684433643/job/15407141816?pr=449

```
  Error: missing cases in switch of type openai.FinishReason: openai.FinishReasonStop, openai.FinishReasonLength, openai.FinishReasonFunctionCall, openai.FinishReasonContentFilter (exhaustive)
  
  Error: issues found
  Ran golangci-lint in 9366ms
```

vvatanabe: @yuikns You can fix like this.

```go
func (r FinishReason) MarshalJSON() ([]byte, error) {
	if r == FinishReasonNull || r == "" {
		return []byte("null"), nil
	}
	return []byte(`"` + string(r) + `"`), nil // best effort to not break future API changes
}
```

yuikns: @vvatanabe I got caught in path dependency before. thank you, fixed

vvatanabe: LGTM

vvatanabe (approved):

vvatanabe:
@yuikns Thank you for patiently discussing! I merged!

PR #454, opened 2023-07-22
Allow modifying requests to satisfy custom needs

arrix:
Some company / organization service gateways may look for additional http headers for authentication or  tracing purpose. Having a request modifier hook in config allows such customization to be implemented without having to fork and patch the sdk.

Use ClientConfig.CustomRequestModifier to make any changes required to the request.

`TestChatCustomRequest` demonstrate a use case where a tracing header is added to the request.

Please let me know if any further changes are needed for this PR.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,6 +1,7 @@
 package openai
 
 import (
+	"context"
 	"net/http"
 	"regexp"
 )
@@ -33,6 +34,9 @@ type ClientConfig struct {
 	APIVersion           string                    // required when APIType is APITypeAzure or APITypeAzureAD
 	AzureModelMapperFunc func(model string) string // replace model to azure deployment name func
 	HTTPClient           *http.Client
+
+	// allows additional modification of request before sending
+	CustomRequestModifier func(ctx context.Context, req *http.Request) error
 
 	EmptyMessagesLimit uint
 }
--- a/chat_test.go
+++ b/chat_test.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	. "github.com/sashabaranov/go-openai"
+	"github.com/sashabaranov/go-openai/internal/test"
 	"github.com/sashabaranov/go-openai/internal/test/checks"
 	"github.com/sashabaranov/go-openai/jsonschema"
 )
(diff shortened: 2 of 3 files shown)


config.go
```diff
@@ -34,6 +35,9 @@ type ClientConfig struct {
 	AzureModelMapperFunc func(model string) string // replace model to azure deployment name func
 	HTTPClient           *http.Client
 
+	// allows additional modification of request before sending
+	CustomRequestModifier func(ctx context.Context, req *http.Request) error
```
vvatanabe, line 39: @arrix  Thank you for the suggestion. However, this can already be achieved with the existing mechanisms. You can modify request by customizing http.Client as follows.

```go
package main

import (
	"net/http"

	"github.com/sashabaranov/go-openai"
)

type CustomRequestTransport struct {
	Origin http.RoundTripper
}

func (t *CustomRequestTransport) RoundTrip(req *http.Request) (*http.Response, error) {
	// Modify request here
	return t.Origin.RoundTrip(req)
}

func main() {
	cfg.HTTPClient = &http.Client{
		Transport: &CustomRequestTransport{
			Origin: http.DefaultTransport,
		},
	}

	client := openai.NewClientWithConfig(cfg)
}

```

arrix: @vvatanabe Thank you for the code snippet. Custom RoundTripper can do the job.
I'm working on some integration SDK for an API gateway and many of my clients already use their own HTTPClient. It's more cumbersome for them to inject request modification code into `RoundTrip` compared to a dedicate config hook. And the go documentation for RoundTrip recommends against modifying the request.

https://pkg.go.dev/net/http#RoundTripper

> RoundTrip should not modify the request, except for
> consuming and closing the Request's Body.

I see both way can do the job. Would a config hook be more explicit and easier to use?

vvatanabe: > RoundTrip should not modify the request, except for consuming and closing the Request's Body.

@arrix When interpreting the description of RoundTripper strictly, it can be said that modifying any part of the request, including the headers, is discouraged. However, in practical operations, it's quite common to add or modify headers for specific purposes, such as authentication or tracing.

Generally, when implementing a custom RoundTripper and there's a need to alter headers, it's recommended to clone the original request before making the modifications. This approach prevents changes to the original request object and ensures operations are carried out without impacting other middleware or code.

In summary, while it's fundamentally discouraged to directly modify headers using RoundTripper, this constraint can be bypassed by cloning the request and then making the necessary changes.

PR #461, opened 2023-07-27
support openrouter client

liu1700:
**Describe the change**
Add a new type of AI backend integration [OpenRouter](https://openrouter.ai/docs)
Can read the doc for more detail, 
In general, it is providing a Model-agnostic ai service backend that provides a variety of service backends including Openai, claude, etc.

Its API schema is compatible with the OpenAI request, so it should be easy to introduce into this project.

**Describe your solution**
Add a new config constructor for open router ai and set the required HTTP header before sending out the request.

**Tests**
I'm using this client in my project to replace my Azure open ai integration.

### DIFF
--- a/client.go
+++ b/client.go
@@ -142,7 +142,6 @@ func sendRequestStream[T streamable](client *Client, req *http.Request) (*stream
 }
 
 func (c *Client) setCommonHeaders(req *http.Request) {
-
 	// https://openrouter.ai/docs#api-keys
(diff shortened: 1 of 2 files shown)


sashabaranov:
It doesn't seem like OpenRouter is a widely used thing. Please take a look at this approach https://github.com/sashabaranov/go-openai/pull/454#discussion_r1271361949

PR #465, opened 2023-08-01
When the azure service is unavailable, the error message returned does not contain the error node

zjy282:
Describe the change
The Json deserialization function of the custom structure ErrorResponse returns the correct and expected message copy for the case where there is no error node in the APIError return information.

Describe your solution
In the custom ErrorResponse structure json deserialization function, judge whether the error node is included, if not, get the message node as the error message, if not, use the response body as the error message.

Tests
Please refer to the TestErrorResponse function in error_test.go.

Issue: When the azure service is unavailable, the error message returned does not contain the error node
For example:
```json
{
    "statusCode": 500,
    "message": "Internal server error",
    "activityId": ""
}
```

### DIFF
--- a/examples/completion-with-tool/main.go
+++ b/examples/completion-with-tool/main.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/sashabaranov/go-openai"
+	"github.com/sashabaranov/go-openai/jsonschema"
+)
+
+func main() {
+	ctx := context.Background()
+	client := openai.NewClient(os.Getenv("OPENAI_API_KEY"))
+
+	// describe the function & its inputs
+	params := jsonschema.Definition{
+		Type: jsonschema.Object,
+		Properties: map[string]jsonschema.Definition{
+			"location": {
+				Type:        jsonschema.String,
+				Description: "The city and state, e.g. San Francisco, CA",
+			},
+			"unit": {
+				Type: jsonschema.String,
+				Enum: []string{"celsius", "fahrenheit"},
+			},
+		},
+		Required: []string{"location"},
+	}
+	f := openai.FunctionDefinition{
+		Name:        "get_current_weather",
+		Description: "Get the current weather in a given location",
+		Parameters:  params,
+	}
+	t := openai.Tool{
+		Type:     openai.ToolTypeFunction,
+		Function: f,
+	}
+
+	// simulate user asking a question that requires the function
+	dialogue := []openai.ChatCompletionMessage{
+		{Role: openai.ChatMessageRoleUser, Content: "What is the weather in Boston today?"},
+	}
+	fmt.Printf("Asking OpenAI '%v' and providing it a '%v()' function...\n",
+		dialogue[0].Content, f.Name)
+	resp, err := client.CreateChatCompletion(ctx,
+		openai.ChatCompletionRequest{
+			Model:    openai.GPT4TurboPreview,
+			Messages: dialogue,
+			Tools:    []openai.Tool{t},
+		},
+	)
+	if err != nil || len(resp.Choices) != 1 {
+		fmt.Printf("Completion error: err:%v len(choices):%v\n", err,
+			len(resp.Choices))
+		return
+	}
+	msg := resp.Choices[0].Message
+	if len(msg.ToolCalls) != 1 {
+		fmt.Printf("Completion error: len(toolcalls): %v\n", len(msg.ToolCalls))
+		return
+	}
+
+	// simulate calling the function & responding to OpenAI
+	dialogue = append(dialogue, msg)
+	fmt.Printf("OpenAI called us back wanting to invoke our function '%v' with params '%v'\n",
+		msg.ToolCalls[0].Function.Name, msg.ToolCalls[0].Function.Arguments)
+	dialogue = append(dialogue, openai.ChatCompletionMessage{
+		Role:       openai.ChatMessageRoleTool,
+		Content:    "Sunny and 80 degrees.",
+		Name:       msg.ToolCalls[0].Function.Name,
+		ToolCallID: msg.ToolCalls[0].ID,
+	})
+	fmt.Printf("Sending OpenAI our '%v()' function's response and requesting the reply to the original question...\n",
+		f.Name)
+	resp, err = client.CreateChatCompletion(ctx,
+		openai.ChatCompletionRequest{
+			Model:    openai.GPT4TurboPreview,
+			Messages: dialogue,
+			Tools:    []openai.Tool{t},
+		},
+	)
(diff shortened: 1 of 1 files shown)


error.go
```diff
@@ -92,3 +92,24 @@ func (e *RequestError) Error() string {
...
+	err = json.Unmarshal(data, &rawMap)
+	if err != nil {
+		return
+	}
+	if _, ok := rawMap["error"]; !ok {
+		e.Error = &APIError{
+			Code: "unknown",
+		}
+		if _, okMessage := rawMap["message"]; okMessage {
+			if json.Unmarshal(rawMap["message"], &e.Error.Message) == nil {
```
vvatanabe, line 107: @zjy282 The activityId gets lost here.

error.go
```diff
@@ -92,3 +92,24 @@ func (e *RequestError) Error() string {
...
+
+func (e *ErrorResponse) UnmarshalJSON(data []byte) (err error) {
+	var rawMap map[string]json.RawMessage
+	err = json.Unmarshal(data, &rawMap)
+	if err != nil {
+		return
+	}
+	if _, ok := rawMap["error"]; !ok {
+		e.Error = &APIError{
+			Code: "unknown",
```
vvatanabe, line 104: @zjy282 This is not the code returned by Azure, so users might get confused.

error.go
zjy282: @vvatanabe Thanks for pointing out the error, I'll fix it next week

error.go
zjy282: @vvatanabe Thank you for your suggestion. In the process of using Azure, I found a variety of error returns that do not conform to the OpenAI specification. I suggest adding the unknown error code type and putting it in the readme as a description to ensure that new unknown errors will be generated in future updates of Azure. Can let the user be treated as a separate logic, what do you think

zjy282:
> 感谢您的善意建议！这是特定于 Azure 的错误响应，对吗？如何仅在使用 Azure 时处理此错误？例如，首先，定义一个名为 AzureServiceError 的类型，如下所示：
> 
> ```
> type AzureServiceError struct {
> 	StatusCode int    `json:"statusCode"`
> 	Message    string `json:"message"`
> 	ActivityId string `json:"activityId"`
> }
> ```
> 
> 接下来，如果它不能作为 APIError 取消封送，那么只有当它与 Azure 相关时，它才作为 AzureServiceError 解送。 https://github.com/sashabaranov/go-openai/blob/master/client.go#L211-L227

Thanks for your suggestion, but I think that at present, users use APIError for error logic processing. If AzureError is separated separately, it may destroy the code compatibility of Azure users.

``` go
e := &openai.APIError{}
if errors.As(err, &e) {
  switch e.HTTPStatusCode {
    case 401:
      // invalid auth or key (do not retry)
    case 429:
      // rate limiting or engine overload (wait and retry) 
    case 500:
      // openai server error (retry)
    default:
      // unhandled
  }
}
```

error.go
zjy282: > @zjy282 The activityId gets lost here.

@vvatanabe done!

vvatanabe:
@zjy282 

> Thanks for your suggestion, but I think that at present, users use APIError for error logic processing. If AzureError is separated separately, it may destroy the code compatibility of Azure users.

How about decoding with AzureServiceError only when using Azure, if the decoding of ErrorResponse fails in the following handleErrorResp function? That way, we won't break compatibility.

```go
func (c *Client) handleErrorResp(resp *http.Response) error {
	var errRes ErrorResponse
	err := json.NewDecoder(resp.Body).Decode(&errRes)
	if err != nil || errRes.Error == nil {
		reqErr := &RequestError{
			Err:            err,
		}
		if errRes.Error != nil {
			reqErr.Err = errRes.Error
		}
		return reqErr
	}

	errRes.Error.HTTPStatusCode = resp.StatusCode
	return errRes.Error
}

type AzureServiceError struct {
	StatusCode int    `json:"statusCode"`
	Message    string `json:"message"`
	ActivityID string `json:"activityId"`
}

func (e *AzureServiceError) Error() string {
	return "" // fixme
}
```

zjy282:
> @zjy282
> 
> > Thanks for your suggestion, but I think that at present, users use APIError for error logic processing. If AzureError is separated separately, it may destroy the code compatibility of Azure users.
> 
> How about decoding with AzureServiceError only when using Azure, if the decoding of ErrorResponse fails in the following handleErrorResp function? That way, we won't break compatibility.
> 
> ```go
> func (c *Client) handleErrorResp(resp *http.Response) error {
> 	var errRes ErrorResponse
> 	err := json.NewDecoder(resp.Body).Decode(&errRes)
> 	if err != nil || errRes.Error == nil {
> 		reqErr := &RequestError{
> 			HTTPStatusCode: resp.StatusCode,
> 			Err:            err,
> 		}
> 		if errRes.Error != nil {
> 			reqErr.Err = errRes.Error
> 		}
> 		return reqErr
> 	}
> 
> 	errRes.Error.HTTPStatusCode = resp.StatusCode
> 	return errRes.Error
> }
> 
> type AzureServiceError struct {
> 	StatusCode int    `json:"statusCode"`
> 	Message    string `json:"message"`
> 	ActivityID string `json:"activityId"`
> }
> 
> func (e *AzureServiceError) Error() string {
> 	return "" // fixme
> }
> ```

@vvatanabe 
I feel that it will still break the compatibility of Azure user Error handling.
My modification is as follows, I don't know if I understand your meaning correctly, if I understand wrong, please point out

```go
func (c *Client) handleErrorResp(resp *http.Response) error {
	var errRes ErrorResponse
	respBody, err := io.ReadAll(resp.Body)
	if err != nil {
		reqErr := &RequestError{
			Err:            err,
		}
		return reqErr
	}
	err = json.Unmarshal(respBody, &errRes)

	if err != nil || errRes.Error == nil {
			azureRes := &AzureServiceError{}
			err := json.Unmarshal(respBody, azureRes)
			if err == nil {
				return azureRes
			}
		}

		reqErr := &RequestError{
			Err:            err,
		}
		if errRes.Error != nil {
			reqErr.Err = errRes.Error
		}
		return reqErr
	}

	errRes.Error.HTTPStatusCode = resp.StatusCode
	return errRes.Error
}

type AzureServiceError struct {
	StatusCode int    `json:"statusCode"`
	Message    string `json:"message"`
	ActivityID string `json:"activityId"`
}

func (e *AzureServiceError) Error() string {
	return fmt.Sprintf("error, status code: %d, message: %s", e.StatusCode, e.Message)
}

````
According to the original error handling method, this way of writing cannot pass the unit test

https://github.com/sashabaranov/go-openai/blob/71a24931dbc5b7029901ff963dc4d0d2509aa7ed/client_test.go#L175-L179

In addition, because resp.Body is repeatedly parsed, the content needs to be read before processing

Issue #344, opened 2023-06-06
Can CreateEmbeddings support custom models?


lizc2003:
Integrate with the third-party large models, their API is compatible with openai, but need to specify their model, so I hope to support custom model.

jaffee:
I believe this can now be closed due to #629

PR #476, opened 2023-08-17
fix: use string instead enum for name of embedding model

usoonees:
**Describe the change**
Please consider using a string instead of an Enum for the name of embedding model. 

This approach provides more flexibility, especially when applying for Azure OpenAI services, as the desired model name may not always be available in the list. 

This change would allow for easier customization. Thank you!

Issue: #344

### DIFF
--- a/embeddings.go
+++ b/embeddings.go
@@ -7,107 +7,42 @@ import (
 
 // EmbeddingModel enumerates the models which can be used
 // to generate Embedding vectors.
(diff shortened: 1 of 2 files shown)


usoonees:
This is also consistent with the model name in `completions.go`, using a string instead of an Enum.
https://github.com/sashabaranov/go-openai/blob/master/completion.go#L20

PR #481, opened 2023-08-25
Allow user to pass in auth token when using NewClientWithConfig

MarkusZhang:
**Describe the change**
Some organisations may have their own GPT proxy, so when using this library, they would need to specify their proxy url. In this case, seems `NewClientWithConfig` is the only function that they can use, but right now the `authToken` field in `ClientConfig` is not exposed, so they won't be able to pass in auth token. To make `NewClientWithConfig` usable for the user, we need to change `authToken` into an expose field. 

**Describe your solution**
Change `authToken` in `ClientConfig` into an expose field. 

**Tests**
Just ran the unit tests, no new tests added as there is no logic change.

Issue: 480

### DIFF
--- a/client.go
+++ b/client.go
@@ -145,10 +145,10 @@ func (c *Client) setCommonHeaders(req *http.Request) {
 	// https://learn.microsoft.com/en-us/azure/cognitive-services/openai/reference#authentication
 	// Azure API Key authentication
 	if c.config.APIType == APITypeAzure {
-		req.Header.Set(AzureAPIKeyHeader, c.config.authToken)
+		req.Header.Set(AzureAPIKeyHeader, c.config.AuthToken)
 	} else {
 		// OpenAI or Azure AD authentication
-		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.authToken))
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.config.AuthToken))
 	}
 	if c.config.OrgID != "" {
 		req.Header.Set("OpenAI-Organization", c.config.OrgID)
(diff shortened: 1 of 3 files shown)


vvatanabe:
@MarkusZhang 

FIY We can already.
```
cfg.BaseURL = "proxyUrl"
```

MarkusZhang:
@vvatanabe Oh thanks for telling, but this kinda unintuitive for the users IMO

MarkusZhang:
@vvatanabe Hi, I did a refactoring on `NewClient` method, by using the options pattern. It makes it very intuitive to the users what input is compulsory, and what is optionally configurable. This does affect any existing usage. Please help review again. Thanks

More about options pattern: https://golang.cafe/blog/golang-functional-options-pattern.html

sashabaranov:
@MarkusZhang You are right, it's less intuitive, but prevents token to be accidentally leaked.

MarkusZhang:
@sashabaranov could you elaborate a bit more? As in leak to who? what issue do you envision?

sashabaranov:
@MarkusZhang basically you can accidentally log the client struct and have auth token printed out

MarkusZhang:
@sashabaranov Ok, then. what about this:
`type Config struct {
	BaseURL              string
	OrgID                string
	APIType              APIType
	APIVersion           string                    
	AzureModelMapperFunc func(model string) string 
	EmptyMessagesLimit uint
}

func NewClientWithConfig(authToken string, config Config) *Client {
  // initialise client ....
}`

Then we change `ClientConfig` to be for the library's internal use only, this way it's more intuitive to use and won't have the accidental logging issue

Issue #488, opened 2023-09-07
Add new "List fine-tuning jobs" endpoint


BrendanMartin:
The new fine-tuning API has a new jobs listing endpoint: 
**GET https://api.openai.com/v1/fine_tuning/jobs**

Docs: https://platform.openai.com/docs/api-reference/fine-tuning/

PR #489, opened 2023-09-07
fine tuning job listing

BrendanMartin:
**Describe the change**
Added support for the listing of fine-tuning jobs.

**Describe your solution**
Essentially copied and modified the fine-tuning events listing code that was already present.

**Tests**
Added unit test.

**Additional context**

Issue: #488

### DIFF
--- a/fine_tuning_job.go
+++ b/fine_tuning_job.go
@@ -51,6 +51,12 @@ type FineTuningJobEvent struct {
 	Type      string `json:"type"`
 }
 
+type FineTuningJobsList struct {
+	Object  string          `json:"object"`
+	Data    []FineTuningJob `json:"data"`
+	HasMore bool            `json:"has_more"`
+}
+
 // CreateFineTuningJob create a fine tuning job.
 func (c *Client) CreateFineTuningJob(
 	ctx context.Context,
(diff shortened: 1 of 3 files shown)


BrendanMartin:
I'm not sure why coverage is lower. I used the same code as FineTuningJobEvent.

sashabaranov:
@BrendanMartin Hey, thank you for the PR! Please throw in a test case here, it should fix the coverage issue: https://github.com/sashabaranov/go-openai/blob/master/client_test.go#L195

fine_tuning_job.go
```diff
@@ -151,3 +157,61 @@ func (c *Client) ListFineTuningJobEvents(
 	err = c.sendRequest(req, &response)
 	return
 }
+
+type listFineTuningJobsParameters struct {
+	after *string
+	limit *int
+}
+
+type ListFineTuningJobsParameter func(*listFineTuningJobsParameters)
```
sashabaranov, line 166: Should we unify all our listing setters? (Maybe even with a small-ish API break?) 

@BrendanMartin Would you have the bandwidth to do this unification? (No worries if not, we can merge and update this afterward!)

@vvatanabe would love to hear your thoughts here.

The other listing setters: https://github.com/sashabaranov/go-openai/blob/master/fine_tuning_job.go#L100

BrendanMartin: Unfortunately, I can't tackle this at the moment. To be honest, I don't even know what unification would look like.

sashabaranov: @BrendanMartin got it, thanks you!

vvatanabe: @sashabaranov Sorry for the late reply! I'd like to consolidate using a concise and generic structure like the one below. It's clear at a glance that it's for pagination purposes.

```
type PaginationQuery struct {
    After *string
    Limit *int
}
```

If we're going with this design, I can create a new PR after we merge this one.

sashabaranov: Merged into a non-master branch, let's take our time and unify the pagination approach

PR #491, opened 2023-09-12
Resolve #487: Enhance tests and fix linting issues

### LINKED ISSUE #487: Use non-empty structs in test server handlers
Currently, in many [test cases](https://github.com/search?q=repo%3Asashabaranov%2Fgo-openai+resBytes&type=code), we return empty structs from the test server. For example:

```go
server.RegisterHandler("/v1/engines/text-davinci-003", func(w http.ResponseWriter, r *http.Request) {
	resBytes, _ := json.Marshal(Engine{}) // <------------- Here we return empty Engine struct
	fmt.Fprintln(w, string(resBytes))
})
```

It is possible for such tests not to cover a case where client does not return the proper response, see for example this discussion: https://github.com/sashabaranov/go-openai/pull/485#discussion_r1316008197

It would be good to make these structs non-empty, and the best solution would probably be to randomize them (keeping randomization logic in the `internal/test` directory).

ealvar3z:
**Describe the change**
- Add randomized fields to Engine structs in tests

**Describe your solution**
- Implement cryptographically secure random utility functions

**Tests**
- Make these structs non-empty, by randomizing them, while keeping randomization logic in the internal/test directory (`internal/test/random.go`). So in the end we'll get something like:

```go
server.RegisterHandler("/v1/engines/text-davinci-003", func(w http.ResponseWriter, r *http.Request) {
	resBytes, _ := json.Marshal(Engine{ID: test.randomString(), Object: test.randomString()})
	fmt.Fprintln(w, string(resBytes))
})
```
- Address golangci-lint warnings

**Additional context**
- None

Issue: #487

### DIFF
--- a/engines_test.go
+++ b/engines_test.go
@@ -8,15 +8,27 @@ import (
 	"testing"
 
 	. "github.com/sashabaranov/go-openai"
+	"github.com/sashabaranov/go-openai/internal/test"
 	"github.com/sashabaranov/go-openai/internal/test/checks"
 )
+
+// Helper function to test the ListEngines endpoint.
+func RandomEngine() Engine {
+	return Engine{
+		ID:     test.RandomString(),
+		Object: test.RandomString(),
+		Owner:  test.RandomString(),
+		Ready:  test.RandomBool(),
+	}
+}
 
 // TestGetEngine Tests the retrieve engine endpoint of the API using the mocked server.
 func TestGetEngine(t *testing.T) {
 	client, server, teardown := setupOpenAITestServer()
 	defer teardown()
 	server.RegisterHandler("/v1/engines/text-davinci-003", func(w http.ResponseWriter, r *http.Request) {
-		resBytes, _ := json.Marshal(Engine{})
+		engine := RandomEngine()
+		resBytes, _ := json.Marshal(engine)
 		fmt.Fprintln(w, string(resBytes))
 	})
 	_, err := client.GetEngine(context.Background(), "text-davinci-003")
@@ -28,7 +40,11 @@ func TestListEngines(t *testing.T) {
 	client, server, teardown := setupOpenAITestServer()
 	defer teardown()
 	server.RegisterHandler("/v1/engines", func(w http.ResponseWriter, r *http.Request) {
-		resBytes, _ := json.Marshal(EnginesList{})
+		engines := make([]Engine, 5)
+		for i := range engines {
+			engines[i] = RandomEngine()
+		}
+		resBytes, _ := json.Marshal(EnginesList{Engines: engines})
 		fmt.Fprintln(w, string(resBytes))
 	})
 	_, err := client.ListEngines(context.Background())
--- a/internal/test/random.go
+++ b/internal/test/random.go
@@ -1,0 +1,35 @@
+package test
+
+import (
+	"crypto/rand"
+	"math/big"
+)
+
+var (
+	strLen = 10
+	//nolint:gomnd // this avoids the golangci-lint "magic number" warning
+	n = big.NewInt(2)
+)
+
+// RandomString generates a cryptographically secure random string of a fixed
+// length. The string is composed of alphanumeric characters and is generated
+// using the crypto/rand library. The length of the string is determined by the
+// constant strLen.
+func RandomString() string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	s := make([]rune, strLen)
+	max := big.NewInt(int64(len(letters)))
+	for i := range s {
+		randomInt, _ := rand.Int(rand.Reader, max)
+		s[i] = letters[randomInt.Int64()]
+	}
+	return string(s)
+}
+
+// RandomBool generates a cryptographically secure random boolean value. It
+// uses the crypto/rand library to generate a random integer (either 0 or 1),
+// and returns true if the integer is 1, and false otherwise.
+func RandomBool() bool {
+	randomInt, _ := rand.Int(rand.Reader, n)
+	return randomInt.Int64() == 1
+}


sashabaranov:
Thank you for the PR!

internal/test/random.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"crypto/rand"
+	"math/big"
+)
+
+var (
+	strLen = 10
+	//nolint:gomnd // this avoids the golangci-lint "magic number" warning
+	n = big.NewInt(2)
```
sashabaranov, line 11: Could we make these `const`?

engines_test.go
```diff
@@ -28,7 +40,11 @@ func TestListEngines(t *testing.T) {
 	client, server, teardown := setupOpenAITestServer()
 	defer teardown()
 	server.RegisterHandler("/v1/engines", func(w http.ResponseWriter, r *http.Request) {
-		resBytes, _ := json.Marshal(EnginesList{})
+		engines := make([]Engine, 5)
```
sashabaranov, line 43: I guess we can use random int here instead

internal/test/random.go
```diff
@@ -0,0 +1,35 @@
...
+// length. The string is composed of alphanumeric characters and is generated
+// using the crypto/rand library. The length of the string is determined by the
+// constant strLen.
+func RandomString() string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	s := make([]rune, strLen)
+	max := big.NewInt(int64(len(letters)))
+	for i := range s {
+		randomInt, _ := rand.Int(rand.Reader, max)
+		s[i] = letters[randomInt.Int64()]
```
sashabaranov, line 24: I suggest we avoid `big` package and go with most basic implementation here e.g., https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go

ealvar3z: Done

engines_test.go
ealvar3z: Done

internal/test/random.go
ealvar3z: Done

engines_test.go
```diff
@@ -8,15 +8,27 @@ import (
...
+	}
+}
+
 // TestGetEngine Tests the retrieve engine endpoint of the API using the mocked server.
 func TestGetEngine(t *testing.T) {
 	client, server, teardown := setupOpenAITestServer()
 	defer teardown()
 	server.RegisterHandler("/v1/engines/text-davinci-003", func(w http.ResponseWriter, r *http.Request) {
-		resBytes, _ := json.Marshal(Engine{})
+		engine := RandomEngine()
```
sashabaranov, line 30: By keeping `engine` variable inside the handler we have no way to compare it to the client-returned result. Also, the comparison would need to be pretty long as we would need to compare all four fields (one would probably be enough)

internal/test/random.go
```diff
@@ -0,0 +1,60 @@
+package test
+
+import (
+	"crypto/rand"
```
sashabaranov, line 4: I suggest we use [math/rand](https://pkg.go.dev/math/rand) package instead. It would allow for [seeding](https://pkg.go.dev/math/rand#Rand.Seed) tests and making them reproducible.

engines_test.go
ealvar3z: I'll move `engine := RandomEngine()` outside of the handler and add some comparison logic that compares only one field.

internal/test/random.go
ealvar3z: Initially I implemented the `Random` funcs with `math/rand` and then ran in to golangci-lint errors. That's when I pivoted to `crypto/rand`. However, I made all the changes as instructed. Will push the code when I get off work here shortly. 

Thank you for the review.

engines_test.go
```diff
@@ -8,27 +8,51 @@ import (
...
+
+	// Compare the two using only one field per code review comment
+	if actualEngine.ID != expectedEngine.ID {
+		t.Errorf("Engine ID mismatch: got %s, expected %s", actualEngine.ID, expectedEngine.ID)
+	}
 }
 
 // TestListEngines Tests the list engines endpoint of the API using the mocked server.
 func TestListEngines(t *testing.T) {
+	test.Seed(42) // Seed the RNG
```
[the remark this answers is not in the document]
ealvar3z: Thank you for your review, @sashabaranov.

I've can go ahead and remove the RNG seeding from `TestListEngines()` as you suggested. I agree that there could be specific scenarios like fuzz testing where non-deterministic behavior is desirable.

// Possible compromise scenario though
We could introduce a flag or environment variable to decide whether or not to seed the RNG, allowing us to keep both deterministic and non-deterministic behavior depending on the use-case. Would that be an acceptable solution for future requirements?

Here's what I am thinking:

```go
func MaybeSeedRNG() {
	seed := os.Getenv("TEST_RNG_SEED")

	if seedValue, err := strconv.ParseInt(seed, 10, 64); err == nil {
		rand.Seed(seedValue)
		return
	}

	// However, if the environment variable is set to "random", 
	// we can use the current time as the seed
	if seedEnv == "random" {
		rand.Seed(time.Now().UnixNano())
	}
}
```
With this setup, we can control the RNG seeding behavior through the `TEST_RNG_SEED` environment variable:

So to make the tests deterministic with a specific seed, we can set the environment variable to that seed (e.g., `TEST_RNG_SEED=42`). To make tests non-deterministic, we can set the environment variable to "random" (`TEST_RNG_SEED=random`).

If we don't set the environment variable, the RNG will not be seeded, maintaining its default non-deterministic behavior.

Then, we could call it whenever we want determinism, e.g:

```go
func TestListEngines(t *testing.T) {
	test.MaybeSeedRNG()  // Call it here
	client, server, teardown := setupOpenAITestServer()
	defer teardown()
	// ... rest of the function remains the same ...
}
```

Please advise.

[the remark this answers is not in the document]
sashabaranov: @ealvar3z I like the `MaybeSeedRNG` approach! Let's factor it out to the separate PR, trying to keep the PR size small-ish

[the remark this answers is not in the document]
ealvar3z: Done. See PR #509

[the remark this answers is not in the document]
sashabaranov: Alright, let's remove `test.Seed(42)` from this PR and merge!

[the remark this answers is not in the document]
ealvar3z: Done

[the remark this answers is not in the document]
sashabaranov: @ealvar3z it's not removed :D

engines_test.go
```diff
@@ -8,27 +8,50 @@ import (
...
 	server.RegisterHandler("/v1/engines/text-davinci-003", func(w http.ResponseWriter, r *http.Request) {
-		resBytes, _ := json.Marshal(Engine{})
+		resBytes, _ := json.Marshal(expectedEngine)
 		fmt.Fprintln(w, string(resBytes))
 	})
-	_, err := client.GetEngine(context.Background(), "text-davinci-003")
+	actualEngine, err := client.GetEngine(context.Background(), "text-davinci-003")
 	checks.NoError(t, err, "GetEngine error")
+
+	// Compare the two using only one field per code review comment
```
sashabaranov, line 38: Could we please remove these comments? They are relevant to review, but not for the codebase going forward

engines_test.go
```diff
@@ -8,27 +8,50 @@ import (
...
+		Ready:  test.RandomBool(),
+	}
+}
+
 // TestGetEngine Tests the retrieve engine endpoint of the API using the mocked server.
 func TestGetEngine(t *testing.T) {
 	client, server, teardown := setupOpenAITestServer()
 	defer teardown()
+
+	expectedEngine := RandomEngine() // move outside of handler per code review comment
```
sashabaranov, line 30: (2) Could we please remove these comments? They are relevant to review but not for the codebase going forward

ealvar3z: Will do. Thank you for your patience.

PR #497, opened 2023-09-23
added delete fine tune model endpoint

BrendanMartin:
**Describe the change**
With the deprecation of the fine-tunes endpoint, OpenAI moved the deletion of fine-tuned models to `DELETE /models/{model}`. This PR adds this new endpoint in `models.go`.

**Describe your solution**
I reworked the code present in `fine_tunes.go` that was already written for deleting fine-tuned models. Because OpenAI organized the new delete under the "models" prefix, this code is now located in the `models.go` file.

**Tests**
The tests used for the previous fine-tune model delete were reworked to use the new endpoint.


sashabaranov (approved):
Thank you so much!

Issue #496, opened 2023-09-22
json: cannot unmarshal string into Go struct field Hyperparameters.hyperparameters.n_epochs of type int


BrendanMartin:
CreateFineTuningJob results in the error:
`json: cannot unmarshal string into Go struct field Hyperparameters.hyperparameters.n_epochs of type int`

It looks like OpenAI is returning the following default for Hyperparameters now:
```json
"hyperparameters": {
    "n_epochs": "auto"
},
```
And CreateFineTuningJob's response decodes to FineTuningJob, where Hyperparameters are defined as an int type.

PR #499, opened 2023-09-25
fix: use any for n_epochs

henomis:
Issue: #496 

**Describe the change**
Hyper parameter `n_epochs` can be either a string or a int value.

**Describe your solution**
Develop a custom un/marshaller for `n_epochs` parameter.

### DIFF
--- a/fine_tuning_job.go
+++ b/fine_tuning_job.go
@@ -2,6 +2,7 @@ package openai
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


shenyute:
When will the PR being merged~

fine_tuning_job.go
```diff
@@ -23,8 +24,40 @@ type FineTuningJob struct {
...
+	} else if h.StringValue != nil {
+		return json.Marshal(*h.StringValue)
+	}
+
+	return nil, fmt.Errorf("invalid hyperparameter n_epochs")
+}
+
 type Hyperparameters struct {
-	Epochs int `json:"n_epochs"`
+	Epochs *HyperparameterNEpochs `json:"n_epochs,omitempty"`
```
sashabaranov, line 60: Hey, thank you so much for the PR! I wonder if we can use `any` here to avoid `Epochs: &HyperparameterNEpochs{` initialization

henomis: done

sashabaranov (approved):

Issue #503, opened 2023-09-29
Add method convert str to embedding model enum


warjiang:
Want featurs like 
```go
ConvertStr2EmbeddingModel("text-embedding-ada-002") 
// return openai.AdaEmbeddingV2
```

henomis:
You can use this:

```go
var m openai.EmbeddingModel
m.UnmarshalText([]byte("text-similarity-ada-002"))
```

PR #504, opened 2023-09-29
feat: add method convert str to embedding model enum

warjiang:
**Describe the change**
Use go-openai as part of my application, params passed to application may be string, but there is no method to convert model name to enum for embedding, so i add this helper method for convenient. 

**Describe your solution**
Reuse the implementation of `stringToEnum`. If the target mapping is found in the `stringToEnum` mapping, use it directly. Otherwise, return the Unknown enumeration type.

**Tests**
ConvertStr2EmbeddingModel("text-embedding-ada-002")  should return AdaEmbeddingV2

**Additional context**
Nop

Issue: #503

### DIFF
--- a/embeddings.go
+++ b/embeddings.go
@@ -12,9 +12,19 @@ import (
 // to generate Embedding vectors.
 type EmbeddingModel int
 
+func ConvertStr2EmbeddingModel(modelName string) EmbeddingModel {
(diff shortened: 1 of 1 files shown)


vvatanabe:
@warjiang  How about the following approach?
https://github.com/sashabaranov/go-openai/issues/503#issuecomment-1749638202

PR #517, opened 2023-11-02
[BREAKING_CHANGES] ChatCompletionRequest in chat.go, chagne data type float32 to float64

pinerchen:
Solve Issue: #9 

**Describe the change**
1) ChatCompletionRequest struct attributes annotation with *omitempty* treats value 0 as not set, for example `temperature`, `top_p` attribute. And this will trigger OpenAI use default value `1`, temperature attribute specifically. [OpenAI Chat Temperature Doc](https://platform.openai.com/docs/api-reference/chat/create#chat-create-temperature)

2) Data type `float32` is not competent for some DBs like MongoDB data type. The only two options to store float data in  MongoDB are `Double` or `Decimal128`, either way will arise error of "float64 can only be truncated to an integer type when truncation is enable" and "cannot decode 128-bit decimal into a float32 or float64 type" respectively in Golang float32.

**Describe your solution**
1) remove some attributes' `omitempty` tag in ChatCompletionRequest.
2) Change `float32` to `float64`.

**Tests**
In MongoDB document, set attribute value data type as `Double` and temperature value to `0` or `0.01`, won't cause error.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -74,13 +74,13 @@ type ChatCompletionRequest struct {
 	Model            string                  `json:"model"`
 	Messages         []ChatCompletionMessage `json:"messages"`
 	MaxTokens        int                     `json:"max_tokens,omitempty"`
-	Temperature      float32                 `json:"temperature,omitempty"`
-	TopP             float32                 `json:"top_p,omitempty"`
+	Temperature      float64                 `json:"temperature"`
+	TopP             float64                 `json:"top_p"`
 	N                int                     `json:"n,omitempty"`
 	Stream           bool                    `json:"stream,omitempty"`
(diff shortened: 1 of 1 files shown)


sashabaranov:
@pinerchen thank you for the PR! 

I don't think removing `omitempty` is the way to go with this issue. I have hopes about `encoding/json/v2` package, though (see https://github.com/sashabaranov/go-openai/issues/9#issuecomment-1791475931). 

I also have a hard time agreeing with `float32` to `float64` conversion as the argument is very ad hoc. By that logic, we would need to convert every `float32` to `float64` due to some unrelated database limitation, which doesn't make sense. (e.g., it would blow up embedding size by a factor of two).

PR #525, opened 2023-11-06
Feat Support chat completion response format

henomis:
**Describe the change**
This branch implement the support for chat completion [response format](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format) optional field

**Describe your solution**
A new type has been added to support this new field.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -69,18 +69,30 @@ type FunctionCall struct {
 	Arguments string `json:"arguments,omitempty"`
 }
 
+type ChatCompletionResponseFormatType string
+
+const (
+	ChatCompletionResponseFormatTypeJSONObject ChatCompletionResponseFormatType = "json_object"
+	ChatCompletionResponseFormatTypeText       ChatCompletionResponseFormatType = "text"
+)
+
+type ChatCompletionResponseFormat struct {
+	Type ChatCompletionResponseFormatType `json:"type"`
+}
+
 // ChatCompletionRequest represents a request structure for chat completion API.
 type ChatCompletionRequest struct {
-	Model            string                  `json:"model"`
-	Messages         []ChatCompletionMessage `json:"messages"`
-	MaxTokens        int                     `json:"max_tokens,omitempty"`
-	Temperature      float32                 `json:"temperature,omitempty"`
-	TopP             float32                 `json:"top_p,omitempty"`
-	N                int                     `json:"n,omitempty"`
-	Stream           bool                    `json:"stream,omitempty"`
-	Stop             []string                `json:"stop,omitempty"`
-	PresencePenalty  float32                 `json:"presence_penalty,omitempty"`
-	FrequencyPenalty float32                 `json:"frequency_penalty,omitempty"`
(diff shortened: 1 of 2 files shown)


stillmatic:
IMO lint should be a separate PR but this LGTM -- consider also adding `seed` here? It's also part of the new release, should similarly also require adding a single int parameter

https://platform.openai.com/docs/api-reference/chat/create#chat-create-seed

chat.go
```diff
@@ -69,18 +69,30 @@ type FunctionCall struct {
...
+	Model            string                        `json:"model"`
+	Messages         []ChatCompletionMessage       `json:"messages"`
+	MaxTokens        int                           `json:"max_tokens,omitempty"`
+	Temperature      float32                       `json:"temperature,omitempty"`
+	TopP             float32                       `json:"top_p,omitempty"`
+	N                int                           `json:"n,omitempty"`
+	Stream           bool                          `json:"stream,omitempty"`
+	Stop             []string                      `json:"stop,omitempty"`
+	PresencePenalty  float32                       `json:"presence_penalty,omitempty"`
+	ResponseFormat   *ChatCompletionResponseFormat `json:"response_format,omitempty"`
```
carsonkahn-external, line 94: Let's [address `seed`](https://github.com/sashabaranov/go-openai/pull/525/#issuecomment-1797332520):

```suggestion
	ResponseFormat   *ChatCompletionResponseFormat `json:"response_format,omitempty"`
	Seed             *int                          `json:"seed,omitempty"`

**Comment by henomis** [comment]
I agree @stillmatic
Thanks @carsonkahn-external

**Comment by sashabaranov** [comment]
@henomis thank you for the PR! Could you please rebase on the latest master version so we could have clean CI runs here? (works for other PRs too!)

File: chat.go [hunk 2]
```diff
@@ -69,18 +69,31 @@ type FunctionCall struct {
...
```
iflamed, line 94: May not the best choice to set the type as a pointer. Should we use the value type as the other field.

henomis: Thanks

chat.go
```diff
@@ -69,18 +69,31 @@ type FunctionCall struct {
...
+	Model            string                       `json:"model"`
+	Messages         []ChatCompletionMessage      `json:"messages"`
+	MaxTokens        int                          `json:"max_tokens,omitempty"`
+	Temperature      float32                      `json:"temperature,omitempty"`
+	TopP             float32                      `json:"top_p,omitempty"`
+	N                int                          `json:"n,omitempty"`
+	Stream           bool                         `json:"stream,omitempty"`
+	Stop             []string                     `json:"stop,omitempty"`
+	PresencePenalty  float32                      `json:"presence_penalty,omitempty"`
+	ResponseFormat   ChatCompletionResponseFormat `json:"response_format,omitempty"`
```
jackieli-tes, line 94: This is optional. So it should have been a pointer: `*ChatCompletionResponseFormat`. Now I get:

```
status code: 400, message: '' is not one of ['json_object', 'text'] - 'response_format.type'
```

PR #528, opened 2023-11-07
feat: initial TTS support

stillmatic:
**Describe the change**
Adds very basic support for `create_speech` endpoint. 

See also https://platform.openai.com/docs/api-reference/audio/createSpeech and https://platform.openai.com/docs/guides/text-to-speech.

**Describe your solution**
This is an extremely minimal approach that returns the http body (`io.ReadCloser`). An additional method that saves the result directly to a file is likely to be useful. For streaming, having the raw chunks is most helpful. The test includes an example for saving to file.

I've added the relevant constants, and the basic API should work, could use more tests for sure. 

**Tests**
I likely don't have time to write the necessary tests. I've added a quick and dirty integration test that succeeds - it synthesizes some text and saves to mp3. Listening to the mp3 confirms that the audio is correct.

It would be useful for somebody to write the rest of the mock methods and add a test mp3 file to return.

**Additional context**
Saving to a particular file format is trivial, but streaming particular forms of audio is tricky and I want to leave that to the end user. Most implementations will require additional format-specific libraries, no need for that in this library.

### DIFF
--- a/speech.go
+++ b/speech.go
@@ -34,11 +34,11 @@ const (
 )
 
 type CreateSpeechRequest struct {
-	Model          SpeechModel           `json:"model"`
-	Input          string                `json:"input"`
-	Voice          SpeechVoice           `json:"voice"`
-	ResponseFormat *SpeechResponseFormat `json:"response_format,omitempty"` // Optional, default to mp3
-	Speed          *float64              `json:"speed,omitempty"`           // Optional, default to 1.0
+	Model          SpeechModel          `json:"model"`
+	Input          string               `json:"input"`
+	Voice          SpeechVoice          `json:"voice"`
+	ResponseFormat SpeechResponseFormat `json:"response_format,omitempty"` // Optional, default to mp3
+	Speed          float64              `json:"speed,omitempty"`           // Optional, default to 1.0
 }
 
 func (c *Client) CreateSpeech(ctx context.Context, request CreateSpeechRequest) (response io.ReadCloser, err error) {
--- a/speech.go
+++ b/speech.go
@@ -9,28 +9,28 @@ import (
 type SpeechModel string
 
 const (
-	TTS_MODEL_1    SpeechModel = "tts-1"
-	TTS_MODEL_1_HD SpeechModel = "tts-1-hd"
+	TTSModel1   SpeechModel = "tts-1"
+	TTsModel1HD SpeechModel = "tts-1-hd"
 )
 
 type SpeechVoice string
 
 const (
-	VOICE_ALLOY   SpeechVoice = "alloy"
-	VOICE_ECHO    SpeechVoice = "echo"
-	VOICE_FABLE   SpeechVoice = "fable"
-	VOICE_ONYX    SpeechVoice = "onyx"
-	VOICE_NOVA    SpeechVoice = "nova"
-	VOICE_SHIMMER SpeechVoice = "shimmer"
+	VoiceAlloy   SpeechVoice = "alloy"
+	VoiceEcho    SpeechVoice = "echo"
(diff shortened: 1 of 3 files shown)


speech_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"io"
+	"os"
+	"testing"
+
+	"github.com/sashabaranov/go-openai"
+	"github.com/sashabaranov/go-openai/internal/test/checks"
+)
+
+func TestSpeechIntegration(t *testing.T) {
+	client := openai.NewClient("sk-xyz")
```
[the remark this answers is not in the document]
stillmatic: [🙏](https://emojipedia.org/folded-hands)

stillmatic:
I feel like the error check that codecov is complaining should be covered by the code here?

Petrzlen:
> I feel like the error check that codecov is complaining should be covered by the code here?

Thanks for the PR! I am giving out a shot. 

Regarding your codecov question
```
speech.go#L50-L51

Added lines #L50 - L51 were not covered by tests
```
then the computer is correct. 

Your test returns an error from the server request handler `response, err = c.sendRequestRaw(req)`;
whereas the coverage complains about `req, err := c.newRequest(ctx, http.MethodPost, c.fullURL("/audio/speech", request.Model)`

I am new to this codebase - it's awesome though - so unsure if even should be tested.

speech.go
```diff
@@ -0,0 +1,56 @@
...
+
+type CreateSpeechRequest struct {
+	Model          SpeechModel          `json:"model"`
+	Input          string               `json:"input"`
+	Voice          SpeechVoice          `json:"voice"`
+	ResponseFormat SpeechResponseFormat `json:"response_format,omitempty"` // Optional, default to mp3
+	Speed          float64              `json:"speed,omitempty"`           // Optional, default to 1.0
+}
+
+func (c *Client) CreateSpeech(ctx context.Context, request CreateSpeechRequest) (response io.ReadCloser, err error) {
```
Petrzlen, line 44: reading other endpoints, it feels that we should check the request parameters here
e.g. how `Completion` does this:
```
func (c *Client) CreateCompletion(
	ctx context.Context,
	request CompletionRequest,
) (response CompletionResponse, err error) {
	if request.Stream {
		err = ErrCompletionStreamNotSupported
		return
	}

	urlSuffix := "/completions"
	if !checkEndpointSupportsModel(urlSuffix, request.Model) {
		err = ErrCompletionUnsupportedModel
		return
	}

	if !checkPromptType(request.Prompt) {
		err = ErrCompletionRequestPromptTypeNotSupported
		return
	}
```

speech.go
siryur, line 1: Quick question: why isn't this under the `audio.go`, since speech is an resource under the audio endpoint(s)?

stillmatic: I thought about this too, but there's very little that actually is shared between this endpoint and the existing audio file. text to speech and speech to text are pretty different.

stillmatic:
hi @nguyendc-systran -- think this should be good to review!

speech.go
```diff
@@ -0,0 +1,87 @@
...
+		err = ErrInvalidVoice
+		return
+	}
+	req, err := c.newRequest(ctx, http.MethodPost, c.fullURL("/audio/speech", request.Model),
+		withBody(request),
+		withContentType("application/json; charset=utf-8"),
+	)
+	if err != nil {
+		return
+	}
```
sashabaranov, line 82: You might want to add a test case here https://github.com/sashabaranov/go-openai/blob/master/client_test.go#L195

stillmatic: thanks! done, and hitting 100% codecov

PR #522, opened 2023-11-06
lint: fix linter warnings reported by golangci-lint

simonklee:
Fix linter warning reported by golangci-lint.

Fixes #519

Issue #529, opened 2023-11-07
DALL•E 3 support


hehua2018:
The big guys strongly want to support the DALL•E 3 interface Well, DALL•E 3 feels too good to use on the web

vvatanabe:
@hehua2018 DONE! Please chk latest version!

PR #530, opened 2023-11-07
Support new fields for /v1/images/generation API

mikecutalo:
**Describe the change**
Updating the [/v1/images/generation](https://platform.openai.com/docs/api-reference/images/create) api with the new changes announced at [devday](https://openai.com/blog/new-models-and-developer-products-announced-at-devday).

**Describe your solution**
Simply add new fields to `ImageRequest` & `ImageResponseDataInner` as well a bunch of new const's for supported options.

**Tests**
Added more fields to the existing `TestImages` test, as these are later unmarshaled. Although if you would like to see tests that simply assert marshalling I could add some, let me know.

Issue: https://github.com/sashabaranov/go-openai/issues/529

### DIFF
--- a/image.go
+++ b/image.go
@@ -13,7 +13,7 @@ const (
 	CreateImageSize256x256   = "256x256"
 	CreateImageSize512x512   = "512x512"
 	CreateImageSize1024x1024 = "1024x1024"
-	// dall-e-3 supported only
+	// dall-e-3 supported only.
 	CreateImageSize1792x1024 = "1792x1024"
 	CreateImageSize1024x1792 = "1024x1792"
 )
--- a/image.go
+++ b/image.go
@@ -13,6 +13,9 @@ const (
 	CreateImageSize256x256   = "256x256"
 	CreateImageSize512x512   = "512x512"
 	CreateImageSize1024x1024 = "1024x1024"
+	// dall-e-3 supported only
+	CreateImageSize1792x1024 = "1792x1024"
+	CreateImageSize1024x1792 = "1024x1792"
 )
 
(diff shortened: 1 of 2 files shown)


vuguet:
could you  set the response on this way ? : 

```
	type ImageResponseDataInner struct {
		URL     string `json:"url,omitempty"`
		B64JSON string `json:"b64_json,omitempty"`
		RevisedPrompt string `json:"revised_prompt"`
	}
```

They now return a prompt in the response : 
image

vuguet:
It is not related to your changes and sorry for so many comments but the linter issue is because of this : 
image

mikecutalo:
Nice good catch let me update the response object. There is already a lint fix PR https://github.com/sashabaranov/go-openai/pull/522 thats been opened.

image.go
```diff
@@ -13,18 +13,38 @@ const (
...
 	CreateImageResponseFormatB64JSON = "b64_json"
 )
 
+const (
+	CreateImageModelDallE2 = "dall-e-2"
+	CreateImageModelDallE3 = "dall-e-3"
+)
+
+const (
+	CreateImageQualityHD = "hd"
```
vvatanabe, line 32: @mikecutalo 
Sometimes you may want to explicitly specify default values in the code. Please add the following as well.

`CreateImageQualityStandard = "standard"`

image

https://platform.openai.com/docs/api-reference/images/create#images-create-quality

vvatanabe (approved):

vvatanabe:
@mikecutalo Thanks for implementing the new API so quickly!

PR #534, opened 2023-11-07
feat(response_format): Add response_format field and values

pluja:
**Description**
I have added the new `response_format` field, see: https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format. I added the field and constant values. Let me know if there's anything else I shold do for this.

**Tests**
No tests needed.

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -67,6 +67,13 @@ const (
 	CodexCodeDavinci001 = "code-davinci-001"
 )
 
+// Response formats for the completion API.
+const (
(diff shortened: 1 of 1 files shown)


henomis:
There is already a [PR](https://github.com/sashabaranov/go-openai/pull/525) for this

pluja:
Oh whoa, I don't know how I searched, but I did not find that. Closing, and sorry for the duplication.

PR #543, opened 2023-11-07
Add SystemFingerprint and chatMsg.ToolCallID field

gburt:
A few small improvements:
- capture the new `system_fingerprint` response field to chat calls
- on ChatCompletionMessage capture the `tool_call_id` that needs to be set for role=TOOL prompts
- fix a typo (ToolChoiche)
- Add omitempty to ChatCompletionMessage optional fields
- Add required ToolCall.type field


sashabaranov:
@gburt thank you for the PR! Seems like there's a conflicting change here

gburt:
> @gburt thank you for the PR! Seems like there's a conflicting change here

you're welcome! thanks for the great library

chat.go
```diff
@@ -62,11 +62,17 @@ type ChatCompletionMessage struct {
 	Name string `json:"name,omitempty"`
 
 	FunctionCall *FunctionCall `json:"function_call,omitempty"`
-	ToolCalls    []ToolCall    `json:"tool_calls,omitempty"`
+
+	// For Role=ASSISTANT prompts this may be set to the tool calls generated by the model, such as function calls.
+	ToolCalls []ToolCall `json:"tool_calls,omitempty"`
```
sashabaranov, line 67: Could you please point out this field in the documentation?

sashabaranov: also, could we please de-capslock the `Role=ASSISTANT` and `Role=TOOL` here? :)

gburt: Ah yeah good call re: caps (we use SCREAMING enums internally...).  [Here are the docs](https://platform.openai.com/docs/api-reference/chat/create) for CreateChatRequest.messages.tool_calls:
image

sashabaranov (approved):

Issue #541, opened 2023-11-07
CreateChatCompletion() fails with error, status code: 400, message: '' is not one of ['json_object', 'text'] - 'response_format.type'


mikeb26:
**Describe the bug**
v1.17.0 introduced a regression into CreateChatCompletion() when using the 3.5 Turbo model. Now this function is failing with:
error, status code: 400, message: '' is not one of ['json_object', 'text'] - 'response_format.type'

If I switch back to v1.16.1 the problem goes away.

**To Reproduce**
```
	client := openai.NewClient(MYKEY)
	dialogue := []openai.ChatCompletionMessage{
		{Role: openai.ChatMessageRoleSystem, Content: MYSYSTEMMSG},
		{Role: openai.ChatMessageRoleUser, Content: MYPROMPT},
	}

	resp, err := client.CreateChatCompletion(
		context.Background(),
		openai.ChatCompletionRequest{
			Model:    openai.GPT3Dot5Turbo,
			Messages: dialogue,
		},
	)
	if err != nil {
		return err  <-- v1.17.0 fails here
	}
```

**Expected behavior**
Non-error return.

**Screenshots/Logs**

**Environment (please complete the following information):**
 - go-openai version: v1.17.0
 - Go version: 1.21.3
 - OpenAI API version: v1
 - OS: Ubuntu 20.04.6 LTS

**Additional context**
n/a

mikeb26:
It looks like commit 6d9c3a6365643d02692ecc6f0b34a5fa3e7fea45 introduced this regression. If I revert it locally the problem goes away. @henomis can you take a look?

jackieli-tes:
oops, filed the same one as you. I searched before drafting :D #542

henomis:
PR here: https://github.com/sashabaranov/go-openai/pull/544

vvatanabe:
@sashabaranov @mikeb26 @jackieli-tes @henomis 
Released the fix for this issue:
https://github.com/sashabaranov/go-openai/releases/tag/v1.17.2

PR #544, opened 2023-11-07
Fix Use pointer for ChatCompletionResponseFormat

henomis:
**Describe the change**
The field `ChatCompletionResponseFormat` must be a pointer
 
Issue: #541

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -89,18 +89,18 @@ type ChatCompletionResponseFormat struct {
 
 // ChatCompletionRequest represents a request structure for chat completion API.
 type ChatCompletionRequest struct {
-	Model            string                       `json:"model"`
-	Messages         []ChatCompletionMessage      `json:"messages"`
-	MaxTokens        int                          `json:"max_tokens,omitempty"`
-	Temperature      float32                      `json:"temperature,omitempty"`
-	TopP             float32                      `json:"top_p,omitempty"`
-	N                int                          `json:"n,omitempty"`
-	Stream           bool                         `json:"stream,omitempty"`
-	Stop             []string                     `json:"stop,omitempty"`
-	PresencePenalty  float32                      `json:"presence_penalty,omitempty"`
-	ResponseFormat   ChatCompletionResponseFormat `json:"response_format,omitempty"`
-	Seed             *int                         `json:"seed,omitempty"`
-	FrequencyPenalty float32                      `json:"frequency_penalty,omitempty"`
+	Model            string                        `json:"model"`
+	Messages         []ChatCompletionMessage       `json:"messages"`
+	MaxTokens        int                           `json:"max_tokens,omitempty"`
+	Temperature      float32                       `json:"temperature,omitempty"`
+	TopP             float32                       `json:"top_p,omitempty"`
+	N                int                           `json:"n,omitempty"`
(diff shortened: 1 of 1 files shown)


jackieli-tes (approved):
Was just drafting  the same thing 👍

simonklee (approved):
Confirmed this as well.

chat.go
```diff
@@ -89,18 +89,18 @@ type ChatCompletionResponseFormat struct {
...
+	Model            string                        `json:"model"`
+	Messages         []ChatCompletionMessage       `json:"messages"`
+	MaxTokens        int                           `json:"max_tokens,omitempty"`
+	Temperature      float32                       `json:"temperature,omitempty"`
+	TopP             float32                       `json:"top_p,omitempty"`
+	N                int                           `json:"n,omitempty"`
+	Stream           bool                          `json:"stream,omitempty"`
+	Stop             []string                      `json:"stop,omitempty"`
+	PresencePenalty  float32                       `json:"presence_penalty,omitempty"`
+	ResponseFormat   *ChatCompletionResponseFormat `json:"response_format,omitempty"`
```
mikeb26, line 101: Change itself LG; maybe consider adding a unit test for issue #541?

staugaard (approved):

chat.go
vvatanabe: @mikeb26 @henomis Thanks for the quick fix. It's self-evident that if the struct is nil, the field will be omitted due to 'omitempty', so I think it's unnecessary to write a unit test for that. It would be better to ensure that the integration test succeeds locally.
refs: https://github.com/sashabaranov/go-openai/blob/master/CONTRIBUTING.md#integration-test

vvatanabe: Oops, it seems like the integration test is currently failing to compile.

vvatanabe: I fixed the integration test on my local machine and tested it. The response was successful!

**Before (in master):**
```sh
$ go test -v -tags=integration ./api_integration_test.go
=== RUN   TestAPI
    api_integration_test.go:61: error, status code: 400, message: '' is not one of ['json_object', 'text'] - 'response_format.type' [CreateChatCompletion (without name) returned error]
    api_integration_test.go:76: error, status code: 400, message: '' is not one of ['json_object', 'text'] - 'response_format.type' [CreateChatCompletion (with name) returned error]
    api_integration_test.go:132: error, status code: 400, message: '' is not one of ['json_object', 'text'] - 'response_format.type' [CreateChatCompletion (with functions) returned error]
--- FAIL: TestAPI (2.76s)
=== RUN   TestAPIError
--- PASS: TestAPIError (0.19s)
FAIL
FAIL	command-line-arguments	3.230s
FAIL
```

**After(in henomis:fix/514-use-pointer-for-ChatCompletionResponseFormat):**
```sh
go test -v -tags=integration ./api_integration_test.go
=== RUN   TestAPI
--- PASS: TestAPI (5.51s)
=== RUN   TestAPIError
--- PASS: TestAPIError (0.19s)
PASS
ok  	command-line-arguments	5.991s
```

vvatanabe (approved):

chat.go
henomis: Thanks!

PR #536, opened 2023-11-07
Feat Implement threads API

henomis:
Describe the change
This branch implements [threads API](https://platform.openai.com/docs/api-reference/threads)

Describe your solution
Create a new implementation file dedicated to this API

Tests
tests attached in `thread_test.go`

Issue: https://github.com/sashabaranov/go-openai/issues/533

PR #546, opened 2023-11-07
Feat/messages api

urjitbhatia:
This branch implements [thread messages API](https://platform.openai.com/docs/api-reference/messages)

Sibling to the threads api implementation https://github.com/sashabaranov/go-openai/pull/536


zhangpeihaoks:
i real need this pl merge,this help me so more

henomis:
Hi, to cover your tests you need to add entries in `TestCases` in `client_test.go`

messages.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+type Message struct {
+	ID          string           `json:"id"`
+	Object      string           `json:"object"`
+	CreatedAt   int              `json:"created_at"`
+	ThreadID    string           `json:"thread_id"`
+	Role        string           `json:"role"`
+	Content     []MessageContent `json:"content"`
+	FileIds     []interface{}    `json:"file_ids"`
```
henomis, line 21:
```suggestion
	FileIds     []string    `json:"file_ids"`
```

messages.go
```diff
@@ -0,0 +1,173 @@
...
+
+type MessagesList struct {
+	Messages []Message `json:"data"`
+
+	httpHeader
+}
+
+type MessageContent struct {
+	Type string      `json:"type"`
+	Text MessageText `json:"text"`
```
henomis, line 37:
```suggestion
	Text *MessageText `json:"text,omitempty"`
	ImageFile *ImageFile `json:"image_file,omitempty"` // <- define this one
```

henomis (changes requested):

wti806:
May I ask when the message API would be ready?

urjitbhatia:
Thanks, I'll work on the fixes this weekend.

urjitbhatia:
> TestCases

Adding the test cases...

However, when looking at these tests - It makes sense to push for a high coverage but some of the work done to keep it high seems forced. For example, covering the branch when the API returns an error.

I'd suggest rethinking the requirement a little bit - it makes sense to cover the main logic but this test seem to add very little or no value - https://github.com/sashabaranov/go-openai/blob/master/client_test.go#L184 - am I missing something in the value of this test? It doesn't actually test that the request building failed due to bad/missing params, only that an `if err` was added to the calls - lmk if that's the intended use anyways.

@henomis @sashabaranov

messages.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+type MessageContent struct {
+	Type      string       `json:"type"`
+	Text      *MessageText `json:"text,omitempty"`
+	ImageFile *ImageFile   `json:"image_file,omitempty"`
+}
+type MessageText struct {
+	Value       string        `json:"value"`
+	Annotations []interface{} `json:"annotations"`
```
sashabaranov, line 42: let's use `any` instead of `interface{}`

messages.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if before != nil {
+		urlValues.Add("before", *before)
+	}
+	encodedValues := ""
+	if len(urlValues) > 0 {
+		encodedValues = "?" + urlValues.Encode()
+	}
+
+	urlSuffix := fmt.Sprintf("/threads/%s%s%s", threadID, messagesSuffix, encodedValues)
```
sashabaranov, line 108: Just a suggestion: changing `messagesSuffix` from `/messages` to `messages` would make formatting more consistent without `%s%s` mixed with `%s/%s`

sashabaranov:
@urjitbhatia thank you so much for the PR! A couple of minor additions and let's merge it 🙌🏻 

> However, when looking at these tests - It makes sense to push for a high coverage but some of the work done to keep it high seems forced. For example, covering the branch when the API returns an error.
> I'd suggest rethinking the requirement a little bit - it makes sense to cover the main logic but this test seem to add very little or no value - https://github.com/sashabaranov/go-openai/blob/master/client_test.go#L184 - am I missing something in the value of this test? 

You know, when I started working on this library a couple of years earlier, I would agree with you. I didn't see any point in extensive testing except for integration tests with the actual API to check that the client actually works. 

Now, I have come to realise a couple of things about testing:

- It's easy to test the "happy path", but to test error cases, you need to design the code with clear interfaces and separation of concerns (`RequestBuilder` and `ErrorAccumulator` are good examples of that)
- You are either forcing code coverage or dealing with the code base with a constant decline of code coverage sometime in the future. I would prefer the former, especially for such a widely-used library
- ~100% code coverage makes it simpler to accept changes into the codebase — it's just easier to trust the code that way

I think the requirement of copy-pasting [3 lines](https://github.com/sashabaranov/go-openai/pull/546/files#diff-4dc56e64f310cd8c12bb00f75c09e83f5d875563ea9901788e525487dc6788b3R304-R306) of tests per client method is a reasonable requirement for all these benefits.

So I think the way forward is to actually reach 100% coverage, add more randomisation(see #491), fuzzing, and integration tests in CI.

> It doesn't actually test that the request building failed due to bad/missing params

Currently the validation happens on the OpenAI backend side, I'll be happy to accept any client validation though 👍🏻

urjitbhatia:
@sashabaranov added the style fixes from the last review :)

496672097:
"I need help. How can I obtain the result of a run after executing the CreateRun() method?"
image

vsumit89:
HI @sashabaranov @urjitbhatia any updates by when this PR will be merged. 
Thanks for all the work you guys have been putting in!

sashabaranov (approved):

lowczarc:
Any reason why `withBetaAssistantV1()` has not been added here ? https://github.com/sashabaranov/go-openai/pull/546/commits/17673a12f5e939f46c64a4761ffd786795aba1aa#diff-a24c457dd8089cef80ea36580edebca6e5dac3a6b277909fffe3cced7c5ba812L69

When using CreateMessage I get an error `You must provide the 'OpenAI-Beta' header to access the Assistants API. Please try again by setting the header 'OpenAI-Beta: assistants=v1'`

messages.go
```diff
@@ -0,0 +1,178 @@
...
+type MessageFilesList struct {
+	MessageFiles []MessageFile `json:"data"`
+
+	httpHeader
+}
+
+// CreateMessage creates a new message.
+func (c *Client) CreateMessage(ctx context.Context, threadID string, request MessageRequest) (msg Message, err error) {
+	urlSuffix := fmt.Sprintf("/threads/%s/%s", threadID, messagesSuffix)
+	req, err := c.newRequest(ctx, http.MethodPost, c.fullURL(urlSuffix), withBody(request))
```
lowczarc, line 74: Any reason for not adding `withBetaAssistantV1()` here ? When using CreateMessage I still get the error `You must provide the 'OpenAI-Beta' header to access the Assistants API. Please try again by setting the header 'OpenAI-Beta: assistants=v1'.`

PR #549, opened 2023-11-07
fix test server setup:

urjitbhatia:
- go map access is not deterministic
- this can lead to a route: /foo/bar/1 matching /foo/bar before matching /foo/bar/1 if the map iteration go through /foo/bar first since the regex match wasn't bound to start and end anchors
- registering handlers now converts * in routes to .* for proper regex matching
- test server route handling now tries to fully match the handler route

This makes testing less flaky

### DIFF
--- a/fine_tuning_job_test.go
+++ b/fine_tuning_job_test.go
@@ -42,7 +42,7 @@ func TestFineTuningJob(t *testing.T) {
 	)
 
 	server.RegisterHandler(
-		"/fine_tuning/jobs/"+testFineTuninigJobID+"/cancel",
+		"/v1/fine_tuning/jobs/"+testFineTuninigJobID+"/cancel",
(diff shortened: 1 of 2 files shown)


sashabaranov:
Thank you for the PR!

fine_tuning_job_test.go
```diff
@@ -42,7 +42,7 @@ func TestFineTuningJob(t *testing.T) {
 	)
 
 	server.RegisterHandler(
-		"/fine_tuning/jobs/"+testFineTuninigJobID+"/cancel",
+		"/v1/fine_tuning/jobs/"+testFineTuninigJobID+"/cancel",
```
sashabaranov, line 45: Why do we need to add `v1` here?

sashabaranov (approved):

PR #526, opened 2023-11-07
Feat Support tools and tools choice new fileds

henomis:
**Describe the change**
This branch adds support for [new fields](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools) `tools` and `tool_choice`

**Describe your solution**
New types have been added to support new features

PR #554, opened 2023-11-08
Fix struct type spelling s/ToolChoiche/ToolChoice

kb:
Correct misspelling of chat.go struct ToolChoice that was added in https://github.com/sashabaranov/go-openai/pull/526

Issue: #522


sashabaranov:
@kb thank you for the PR! I think this was fixed in an earlier PR

Issue #555, opened 2023-11-08
"type" field missing in "tool_calls" of the new gpt parallel function calling feature 


Benciveni:
- OpenAI API version: v1.17.3

openai.ChatCompletionMessage{
					Role: openai.ChatMessageRoleAssistant,

					ToolCalls: []openai.ToolCall{
						{
							ID: FuncId,
							Type: "function",  //<<<<<<<<<<<<<<<<<<<<<<<< Missing type field
							Function: openai.FunctionCall{
								Name:      FuncName,
								Arguments: FuncArgs,
							},
						}},
				}

henomis:
This one should fix: https://github.com/sashabaranov/go-openai/pull/558

Benciveni:
Tested and it works ! 
Thank you for the great API.

PR #558, opened 2023-11-09
fix: add missing fields in tool_calls

henomis:
**Describe the change**
There are [missing fields](https://platform.openai.com/docs/api-reference/chat/streaming) in `tool_calls` object. Following the documentation the [chat completion chunk object](https://platform.openai.com/docs/api-reference/chat/streaming) has the `index` field also.

**Describe your solution**
Just added missing fields

Issue: #555

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -66,7 +66,10 @@ type ChatCompletionMessage struct {
 }
 
 type ToolCall struct {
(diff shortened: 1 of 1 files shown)


sashabaranov (approved):
Thank you!!

Issue #533, opened 2023-11-07
assistants API


Koeng101:
**Is your feature request related to a problem? Please describe.**
Today, OpenAI announced their assistants API. I would like support for this API in golang. https://platform.openai.com/docs/api-reference/assistants

**Describe the solution you'd like**
Support for the assistants API

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

I expect others also will want this feature, and hopefully someone is working on a fork right now! If not, I'll probably eventually make one, since I want to use it.

danai-antoniou:
@henomis thanks so much for adding assistants and threads! 🚀 OOI are you planning on adding [runs](https://platform.openai.com/docs/api-reference/runs) along with the threads? I can give it a stab if you're not adding it in the threads PR

Koeng101:
Holy crap. That was fast.

henomis:
I think threads and runs must have a separated PRs

henomis:
@sashabaranov @vvatanabe 
[Run API PR](https://github.com/sashabaranov/go-openai/pull/560) is completed. It's a lot of code, I need some help from the go-openai community to test/review this one before merging it.

jacewalker:
I can't see that we can add messages to an existing thread, have I missed this somewhere?

kb:
@henomis I've been working on some code that is using the Assistant, Thread, Run, (and yet to be merged Message) APIs and things so far have worked out. Just put up a small PR to fix one issue I had with the Runs logic. https://github.com/sashabaranov/go-openai/pull/561

henomis:
@kb Great job! Thanks for testing! I'll check your PR

PR #560, opened 2023-11-09
Feat implement Run APIs

henomis:
**Describe the change**
This branch implements [Run APIs](https://platform.openai.com/docs/api-reference/runs)

Issue: #533 

⚠️  It's a lot of code, I need some help from the go-openai community to test/review this one before merging it.


sashabaranov:
Thank you for the PR! ❤️  We would need to add some integration tests for Assistants API at some point in the future

run.go
```diff
@@ -0,0 +1,358 @@
...
+
+// SubmitToolOutputs submits tool outputs.
+func (c *Client) SubmitToolOutputs(
+	ctx context.Context,
+	threadID string,
+	runID string,
+	request SubmitToolOutputsRequest) (response Run, err error) {
+	urlSuffix := fmt.Sprintf("/threads/%s/runs/%s/submit_tool_outputs", threadID, runID)
+	req, err := c.newRequest(ctx, http.MethodPost, c.fullURL(urlSuffix), withBody(request),
+		withBetaAssistantV1())
```
sashabaranov, line 262: nit: could we please either use single-line or all-in multiple lines, so here it would be 

```go
func (c *Client) SubmitToolOutputs(
	ctx context.Context,
	threadID string,
	runID string,
	request SubmitToolOutputsRequest,
) (response Run, err error) {
	urlSuffix := fmt.Sprintf("/threads/%s/runs/%s/submit_tool_outputs", threadID, runID)
	req, err := c.newRequest(
		ctx,
		http.MethodPost,
		c.fullURL(urlSuffix),
		withBody(request),
		withBetaAssistantV1(),
	)
```

run.go
```diff
@@ -0,0 +1,358 @@
...
+
+// ListRunSteps lists run steps.
+func (c *Client) ListRunSteps(
+	ctx context.Context,
+	threadID string,
+	runID string,
+	limit *int,
+	order *string,
+	after *string,
+	before *string,
```
sashabaranov, line 328: I wonder if we should pack all these arguments into a struct?

henomis: Yeah, I agree

henomis: done

sashabaranov (approved):

PR #561, opened 2023-11-10
Add `json:"metadata,omitempty"` to RunRequest struct

kb:
Metadata is an optional field per the api spec
https://platform.openai.com/docs/api-reference/runs/createRun


sashabaranov:
thank you!

sashabaranov (approved):

PR #566, opened 2023-11-13
fix: add beta assistant header to CreateMessage call

floodfx:
**Describe the change**
Calling the `CreateMessage` method currently fails with the following error message: `You must provide the 'OpenAI-Beta' header to access the Assistants API. Please try again by setting the header 'OpenAI-Beta: assistants=v1'.`

**Describe your solution**
I've updated the `CreateMessage` call in `messages.go` to include the `withBetaAssistantV1()` request option.

**Tests**
It was unclear how to update the tests as I didn't see a way to test the header was set.  I have tested that the call succeeds in my application now.


urjitbhatia:
Good catch, I probably forgot to add it for that call.

sashabaranov (approved):

sashabaranov:
Thank you for the PR!

PR #568, opened 2023-11-13
feat: support direct bytes for file upload

floodfx:
**Describe the change**
`CreateFile` currently requires a "local" file which it reads from the filesystem and uploads.  For various reasons, we require uploading bytes directly rather than relying on local filesystem.

**Describe your solution**
Add `CreateFileBytes` which takes a `FileBytesRequest` that allows directly passing a `[]byte`.  This reuses the exsting `CreateFormFileReader` passing a `*bytes.Reader` directly.

**Tests**
Added a test case to `files_api_test.go`.   Also validated that it works in a separate app.


sashabaranov (approved):
Thank you so much!

PR #569, opened 2023-11-14
Update messages.go

496672097:
Add a header to the CreateMessage method; otherwise, the message will not be created successfully


floodfx:
Dupe of #566

sashabaranov:
Thank you for the PR, this is already implemented!

PR #577, opened 2023-11-14
Add File purposes as constants

revett:
## Describe the change

- When uploading a `File`, you need to define a `.Purpose`
- Currently this is a `string` that is passed to the `FileRequest` type
- It would be nicer to have these as constants

image

> See: https://platform.openai.com/docs/api-reference/files

## Describe your solution

Add a constant for each supported purpose as defined in the OpenAI API.

## Tests

No tests needed.

## Additional context

Not related to an issue, just something I saw when using the new Assistants API.

### DIFF
--- a/files.go
+++ b/files.go
@@ -20,9 +20,9 @@ type PurposeType string
 
 const (
 	PurposeFineTune         PurposeType = "fine-tune"
-  PurposeFineTuneResults  PurposeType  = "fine-tune-results"
+	PurposeFineTuneResults  PurposeType = "fine-tune-results"
 	PurposeAssistants       PurposeType = "assistants"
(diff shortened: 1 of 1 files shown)


floodfx:
I added something similar in #568 but with more type safety.  We should probably merge these efforts?  I can add the additional types to #568.  I am unclear whether the type safe approach I used is typical of this project though?

revett:
> I added something similar in #568 but with more type safety. We should probably merge these efforts? I can add the additional types to #568. I am unclear whether the type safe approach I used is typical of this project though?

Thanks! I've rebased your merged changes in to this PR. Looks like we still need two more purposes added.

@sashabaranov This is ready for review 👍

sashabaranov (approved):
Thank you!

PR #557, opened 2023-11-09
Add support for vision chat completion models

pufferffish:
**Describe the change**
This PR let's you pass in `image_url` so we can pass in images to vision models like `gpt-4-vision-preview`

**Describe your solution**
The definition of `ChatCompletionMessage` is changed such that we can pass in a struct containing text and images for `Content` instead of a string.

**Tests**
This PR has been tested in `example/chatvision`

Issue: #539

Issue #539, opened 2023-11-07
Add support for image links and base64 images in chat completions


KaffeDiem:
Currently the package does not support adding either base64 encoded images to the chat completions or an image link. 

[This is a new feature just introduced.](https://platform.openai.com/docs/guides/vision)

rkintzi:
Is anyone already working on this? Any ideas on how to implement this feature?

The OpenAI API has been updated to allow the message content to be either a string or an array of objects. To ensure backward compatibility, I  propose adding a new field to the ChatCompletionMessage. Here's my suggested approach (just a draft): 



```go
package main

import (
	"encoding/json"
	"fmt"
	"strings"
	"testing"

	"github.com/stretchr/testify/require"
)

type ContentType string

const (
	ContentTypeText  ContentType = "text"
	ContentTypeImage ContentType = "image_url"
)

type Part struct {
	Type     ContentType `json:"type"`
	ImageUrl string      `json:"image_url,omitempty"`
	Text     string      `json:"text,omitempty"`
}

type Parts []Part

type ChatCompletionMessage struct {
	Role    string `json:"role"`
	Content string `json:"-"`
	Parts   Parts  `json:"content"`
}

func (ps Parts) MarshalJSON() ([]byte, error) {
	if len(ps) == 0 {
		return []byte("\"\""), nil
	}
	cc := []Part(ps)
	if len(cc) == 1 && cc[0].Type == "text" {
		if cc[0].Text != "" {
			return json.Marshal(cc[0].Text)
		}
		return []byte("\"\""), nil
	}
	return json.Marshal([]Part(cc))
}

func (ps *Parts) UnmarshalJSON(bs []byte) error {
	if bs[0] == '"' && bs[len(bs)-1] == '"' {
		if len(bs) == 2 {
			*ps = nil
			return nil
		}
		var s string
		err := json.Unmarshal(bs, &s)
		if err != nil {
			return err
		}
		*ps = Parts{{Type: ContentTypeText, Text: s}}
		return nil
	}
	var parts []Part
	err := json.Unmarshal(bs, &parts)
	if err != nil {
		return err
	}
	*ps = parts
	return nil
}

func (m *ChatCompletionMessage) UnmarshalJSON(bs []byte) error {
	msg := struct {
		Role    string `json:"role"`
		Content string `json:"-"`
		Parts   Parts  `json:"content"`
	}(*m)
	err := json.Unmarshal(bs, &msg)
	if err != nil {
		return err
	}
[...]

PR #580, opened 2023-11-14
Add support for multi part chat messages

rkintzi:
OpenAI has recently introduced a new model called gpt-4-visual-preview, which now supports images as input. The chat completion endpoint accepts multi-part chat messages, where the content can be an array of structs in addition to the usual string format.

This commit introduces new structures and constants to represent different types of content parts. It also implements the json.Marshaler and json.Unmarshaler interfaces on ChatCompletionMessage.

Issue: #539
Similar PR: #557


emil14:
any updates?

rkintzi:
We are waiting for review.

sashabaranov:
Hey, thank you for the PR! Sorry for being so slow with the review!

chat.go
```diff
@@ -51,9 +52,32 @@ type PromptAnnotation struct {
...
 
+const (
+	ImageURLDetailHigh = "high"
+	ImageURLDetailLow  = "low"
+	ImageURLDetailAuto = "auto"
+)
+
+type ChatMessageImageURL struct {
+	URL    string `json:"url,omitempty"`
+	Detail string `json:"detail,omitempty"`
```
sashabaranov, line 63: Could we please make it more type-safe, by using a separate type for `Detail`?

Something like

```go
type ImageURLDetail string

const (
	ImageURLDetailHigh ImageURLDetail = "high"
	ImageURLDetailLow  ImageURLDetail = "low"
	ImageURLDetailAuto ImageURLDetail = "auto"
)

type ChatMessageImageURL struct {
	URL    string `json:"url,omitempty"`
	Detail ImageURLDetail `json:"detail,omitempty"`
}
```

chat.go
```diff
@@ -70,6 +94,64 @@ type ChatCompletionMessage struct {
 	ToolCallID string `json:"tool_call_id,omitempty"`
 }
 
+func (m ChatCompletionMessage) MarshalJSON() ([]byte, error) {
+	if m.Content != "" && m.MultiContent != nil {
+		return nil, errors.New("can't use both Content and MultiContent files simultaneously")
```
sashabaranov, line 99: Let's define this error as a public variable and test that we are actually getting it

chat.go
```diff
@@ -70,6 +94,64 @@ type ChatCompletionMessage struct {
...
+	}
+	msg := struct {
+		Role         string            `json:"role"`
+		Content      string            `json:"content"`
+		MultiContent []ChatMessagePart `json:"-"`
+		Name         string            `json:"name,omitempty"`
+		FunctionCall *FunctionCall     `json:"function_call,omitempty"`
+		ToolCalls    []ToolCall        `json:"tool_calls,omitempty"`
+		ToolCallID   string            `json:"tool_call_id,omitempty"`
+	}(m)
```
sashabaranov, line 121: I think I see this `struct{...}(m)` pattern for the first time :D How is it called?

sashabaranov: Found it! https://go.dev/ref/spec#Conversions

chat.go
```diff
@@ -70,6 +94,64 @@ type ChatCompletionMessage struct {
...
+	multiMsg := struct {
+		Role         string `json:"role"`
+		Content      string
+		MultiContent []ChatMessagePart `json:"content"`
+		Name         string            `json:"name,omitempty"`
+		FunctionCall *FunctionCall     `json:"function_call,omitempty"`
+		ToolCalls    []ToolCall        `json:"tool_calls,omitempty"`
+		ToolCallID   string            `json:"tool_call_id,omitempty"`
+	}{}
+	if err := json.Unmarshal(bs, &multiMsg); err == nil {
```
sashabaranov, line 135: We probably should start unmarshalling from the non-image case, as it's the most frequent

sashabaranov:
Looks good! Just a couple of minor fixes and let's merge

sashabaranov (approved):
Thank you for the updates! Note: there were more tests, do you want to return them or can we merge like that?

chat.go
rkintzi: I did this, even that I don't really like the idea. Could you explain why this is useful here and not with message roles?

rkintzi:
I introduced requested changes.

chat.go
sashabaranov: @rkintzi it adds type safety, so the compiler can do more checks for us. I think we should do this for roles too!

rkintzi:
Not sure what you are referring with "more tests". Do you mean that we catch errors in tests? If so, then my answer is no. I don't thing it is worth wrapping json.Unmarshal/json.Marshal errors only to be able to catch them in test. This could potentially break someones code.

sashabaranov:
@rkintzi There was this whole block of checks https://github.com/sashabaranov/go-openai/pull/580/commits/152021253721b4d734a08b97c3881726fe520ba0#diff-035a00c5352f69159c65e81dbeffed087f0028e2a9324d7ff62786940f548827R339-R354

rkintzi:
And what about them? I really don't get what is the change that you suggest?

sashabaranov:
@rkintzi ah, sorry, they are still here. I was looking only at the diff from the latest commit! Thank you for the PR! ❤️

rkintzi:
Sure thing! You're welcome!

PR #598, opened 2023-11-18
Add completion-with-tool example

mikeb26:
**Describe the change**
Add a completion-with-tool example showing how to correctly utilize the function calling capability of GPT4

**Describe your solution**
This is based on the existing completion example and uses the same get_current_weather() python function that the OpenAI documentation includes.

**Tests**
Manually run the new example.

**Additional context**
As a user of this go SDK it was not immediately intuitive to me how to correctly utilize the function calling capability of GPT4 (https://platform.openai.com/docs/guides/function-calling). While the aformentioned link provides a helpful example written in python, I initially tripped over how to correctly translate the specification of function arguments when using this go SDK.

### DIFF
--- a/examples/completion-with-tool/main.go
+++ b/examples/completion-with-tool/main.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/sashabaranov/go-openai"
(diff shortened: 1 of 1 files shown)


sashabaranov:
Thank you for the PR! Looks good!

sashabaranov (approved):

PR #603, opened 2023-11-21
Add canary-tts to speech models

pjuhasz:
There is an additional text-to-speech model called "canary-tts", which is not advertised on OpenAI's API documentation page, nevertheless it appears among the allowed models, and it seems to work with exactly the same voices as the main tts-1 models in exactly the same way.

The only important difference is that it is currently not rate limited as aggressively as the tts-1 models, which may be useful for those on the free tier.

It's entirely possible that the mere existence of this model is a mistake on OpenAI's part, but as long as it is available, we might as well use it.

The commit also fixes a typo in one of the existing constants. This is technically a breaking change, but it seemed like the proper thing to fix it.

### DIFF
--- a/speech.go
+++ b/speech.go
@@ -10,8 +10,9 @@ import (
 type SpeechModel string
 
 const (
-	TTSModel1   SpeechModel = "tts-1"
-	TTsModel1HD SpeechModel = "tts-1-hd"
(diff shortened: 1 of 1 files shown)


sashabaranov (approved):
Thank you for the PR!

PR #610, opened 2023-11-29
feat: expose io.Reader of streamReader

WqyJh:
Expose `*bufio.Reader` for custom operation which is harmless and maybe useful for someone.


sashabaranov:
@WqyJh thank you for the PR! The `streamReader` is private, so we'll only use such a method when it would be really needed internally. Let's not add it until then to avoid extra complexity.

Issue #581, opened 2023-11-15
Support IO stream instead of local file path in file apis


Nomango:
Most of the current file-related interfaces directly read local files, such as `CreateFile`, `CreateAssistantFile`, etc.

It is better to pass `io.Reader` directly. Otherwise, if I want to upload JSON data from memory, I must first write it to a local file.

Similar to #246, but with a larger scope of modifications.

PR #622, opened 2023-12-21
feat: CreateFile support io.Reader

eiixy:
Issue: #581


abdullah2993:
would be nice if this one gets merged. It follow the same api as [AudioRequest](https://github.com/sashabaranov/go-openai/blob/eff8dc1118ea82a1b50ee316608e24d83df74d6b/audio.go#L32)

Issue #627, opened 2024-01-01
Azure image endpoint URL is broken


ikarius:
**Describe the bug**

Creating a new image with a text-to-image model, such as Dall-E 2|3 is broken when using Azure platform. 
Using current `Client.CreateImage()` method creates an incorrect Azure endpoint URL, because the model/deployment name is not passed as a parameter  to `Client.fullURL()` method, hence always creating an endpoint URL with `UNKNOWN` as the model name (and causing a 404 from Azure side).

**To Reproduce**

Just try to create an image on Azure platform, as described in the README (url or b64).

**Expected behavior**

The model/deployment name is properly included in the generated Azure endpoint URL.

**Screenshots/Logs**

For instance, using a `dall-e-3` model generates an URL like (the `UNKNOWN` part should be the name of the Azure deployment, i.e. `dall-e-3`) : 
```
url: https://ikariusopenaisweden.openai.azure.com/openai/deployments/UNKNOWN/images/generations?api-version=2023-12-01-preview
2023/12/31 23:30:39 Erreur: error, status code: 404, message: The API deployment for this resource does not exist. If you created the deployment within the last 5 minutes, please wait a moment and try again.
```

**Environment:**
 - go-openai version: v1.17.9
 - Go version: 1.21.5
 - OpenAI API version: [e.g. v1]
 - OS: Manjaro Linux
 
**Additional context**

Updating `CreateImage` to : 
```
// CreateImage - API call to create an image. This is the main endpoint of the DALL-E API.
func (c *Client) CreateImage(ctx context.Context, request ImageRequest) (response ImageResponse, err error) {
	urlSuffix := "/images/generations"
	req, err := c.newRequest(ctx, http.MethodPost, c.fullURL(urlSuffix, request.Model), withBody(request))
	if err != nil {
		return
	}

	err = c.sendRequest(req, &response)
	return
}
```

does the job (passing the model to `fullURL`).

PR #608, opened 2023-11-27
Added support for createImage Azure models

zjy282:
**Describe the change**
Added support for createImage Azure DALL-E 3 models.

PR #628, opened 2024-01-01
fix Azure endpoint URL for image creation

ikarius:
**Describe the change**
Model or deployment name must be included in the generated Azure endpoint URL when using image creation (f.i. with Dall-E).

None

**Describe your solution**
Properly passing deployment/model name when creating the Azure endpoint URL.

**Tests**
URL is now properly generated (just passing the expected/forgotten? parameter).

Issue: #627


sashabaranov:
Thank you for the PR! Duplicates #608

Issue #602, opened 2023-11-21
Embeddings API forcing model to be an int is awkward


jaffee:
**Is your feature request related to a problem? Please describe.**
I'm writing a tool that allows the user to optionally specify the name of the model they'd like to use for various requests. For completion requests this works fine as the model name is accepted as a string. For embeddings however, afaict, there is no way to specify an embedding that isn't already defined within the library, and to take in a string from the user, one would have to create a pointer to an EmbeddingModel and then call UnmarshalText on the user's input to convert it to the int representation, then the library would take it and ultimately convert it back to a string before sending to OpenAI. If the Embeddings API took the model as a string:

1) It would loosen the coupling between the library and OpenAI, because when OpenAI adds a new Embedding model, it will be possible for users to use it without needing to upgrade the library immediately.
2) It would make the embeddings API more consistent with the completions API.
3) It would reduce complexity and lines of code (I think)

**Describe the solution you'd like**
I'd like to make the EmbeddingModel type a string. I think this would be a non-breaking change for most users of the library, since they're already required to refer to openai.AdaEmbeddingV2. However it would technically be a breaking change to the API if someone were serializing the model as an int to store it in a config file or something.

I'm happy to take a stab at making this change, but want to make sure it has a chance of getting accepted before I go through the effort.

**Describe alternatives you've considered**
If this isn't possible I think I just have to fork the library, change it, and vendor it into my project.

I guess if a breaking change is unacceptable, some new API could be added which allowed this... even just adding one more field to EmbeddingRequest called ModelString that's used if it's non-empty. Seems messy though 🤷

juliuslipp:
+1 This was started once already, but never made it:
https://github.com/sashabaranov/go-openai/pull/476

jaffee:
There is now a PR https://github.com/sashabaranov/go-openai/pull/629

PR #629, opened 2024-01-03
[BREAKING_CHANGES] convert EmbeddingModel to string type

jaffee:
This gives the user the ability to pass in models for embeddings that are not already defined in the library. Also more closely matches how the completions API works.

**Describe the change**
Addresses the issue described in #602 essentially that forcing embedding model to be represented as an int is awkward.

The integer representation is never used by openai, only the string https://platform.openai.com/docs/api-reference/embeddings

**Describe your solution**
Gets rid of the extra layer of indirection and uses the string value directly.

**Tests**
This mostly ended up deleting code, and so deleted the associated test as well.

**Additional context**

Issue: #602

### DIFF
--- a/embeddings.go
+++ b/embeddings.go
@@ -13,107 +13,29 @@ var ErrVectorLengthMismatch = errors.New("vector length mismatch")
 
 // EmbeddingModel enumerates the models which can be used
 // to generate Embedding vectors.
-type EmbeddingModel int
-
-// String implements the fmt.Stringer interface.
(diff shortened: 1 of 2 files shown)


sashabaranov:
@jaffee thank you so much for the PR! (and sorry for taking so long to review)

sashabaranov (approved):

jaffee:
thanks @sashabaranov 🙏 . Lovely library, appreciate you.

Issue #636, opened 2024-01-15
Use setters pattern for `fullURL` method


sashabaranov:
Currently client's `fullURL` method expects optional first argument to be an Azure model name:

```go
// fullURL returns full URL for request.
// args[0] is model name, if API type is Azure, model name is required to get deployment name.
func (c *Client) fullURL(suffix string, args ...any) string {
```

Let's make this more explicit by using setters pattern similar to `newRequest` method:

```go
func (c *Client) newRequest(ctx context.Context, method, url string, setters ...requestOption) (*http.Request, error) {
```

PR #637, opened 2024-01-16
Fix Azure embeddings model detection by passing string to `fullURL`

greysteil:
A [recent pull request](https://github.com/sashabaranov/go-openai/pull/629) broke support for the Azure OpenAI embeddings endpoint by [passing a variable of type `EmbeddingModel` to `client.fullURL`](https://github.com/sashabaranov/go-openai/pull/629/files#diff-0fdd78c64708f1e3c576725363d6d7accb8b43590810436765198e7a5522770fR231). This breaks the [Azure deployment name lookup](https://github.com/sashabaranov/go-openai/blob/09f6920ad04666f65dd86ed542e5ebf8bffc93a8/client.go#L223-L229).

This pull request fixes Azure embeddings support by casting the type to `string`.

**Tests**
Added tests for generating Azure embeddings 

Note: I'm not a great Go programmer. If there's a better way to do this I'm all ears!

### DIFF
--- a/embeddings.go
+++ b/embeddings.go
@@ -228,7 +228,7 @@ func (c *Client) CreateEmbeddings(
 	conv EmbeddingRequestConverter,
 ) (res EmbeddingResponse, err error) {
 	baseReq := conv.Convert()
-	req, err := c.newRequest(ctx, http.MethodPost, c.fullURL("/embeddings", baseReq.Model), withBody(baseReq))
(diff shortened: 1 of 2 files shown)


sashabaranov:
@greysteil great catch, thank you!

For context: here's the exact part which breaks

https://github.com/sashabaranov/go-openai/blob/master/client.go#L224-L229

And a minimal code example to reproduce: https://go.dev/play/p/KakKIBplunp

There's an open issue to get rid of `args ...any` there and make this part type-safe https://github.com/sashabaranov/go-openai/issues/636

sashabaranov (approved):

PR #638, opened 2024-01-16
fix(audio): fix audioTextResponse decode

WqyJh:
The audio api passes `audioTextResponse` to `sendRequest`,
```go
	if request.HasJSONResponse() {
		err = c.sendRequest(req, &response)
	} else {
		var textResponse audioTextResponse
		err = c.sendRequest(req, &textResponse)
		response = textResponse.ToAudioResponse()
	}
```

but it cannot be asserted as `*string` type, lead to decoding failure.
```go
func decodeResponse(body io.Reader, v any) error {
	if v == nil {
		return nil
	}

	if result, ok := v.(*string); ok {
		return decodeString(body, result)
	}
	return json.NewDecoder(body).Decode(v)
}
```

Fix this problem.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"github.com/sashabaranov/go-openai/internal/test"
+	"github.com/sashabaranov/go-openai/internal/test/checks"
 )
 
 var errTestRequestBuilderFailed = errors.New("test request builder failed")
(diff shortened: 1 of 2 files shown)


sashabaranov:
@WqyJh thank you for the PR! Could you also please add a test for that?

WqyJh:
Added test.

sashabaranov:
@WqyJh thanks! There are some linter problems now

client_test.go
```diff
@@ -43,37 +44,67 @@ func TestDecodeResponse(t *testing.T) {
...
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			err := decodeResponse(tc.body, tc.value)
-			if (err != nil) != tc.hasError {
-				t.Errorf("Unexpected error: %v", err)
+			if tc.hasError {
+				if err == nil {
+					t.Error("Unexpected nil error")
+				}
+			} else {
```
sashabaranov, line 94: we can reduce the amount of `else` by using early return pattern:

```
if something {
// code
    return
}

// else block
```

WqyJh: Of course! It would be better to use github.com/stretchr/testify/assert for testing, as it eliminates the need for if/else statements for assertions. However, I didn't use it because it doesn't have any external dependencies. Would you like to use it or keep it dependency-free?

sashabaranov: @WqyJh staying dependency free is important for the project, so let's not use the library. We have an internal lib of helpers though: https://github.com/sashabaranov/go-openai/blob/master/internal/test/checks/checks.go

sashabaranov (approved):
Thank you!

PR #645, opened 2024-01-28
Add new struct filed dimensions for embedding API

chenhhA:
**Describe the change**
Add new struct field `dimensions`  that support for shortening embeddings. Release in January 25, 2024

* [New embedding models and API updates](https://openai.com/blog/new-embedding-models-and-api-updates)
* [Embedding API Reference](https://platform.openai.com/docs/api-reference/embeddings)

**Describe your solution**
Add new struct filed

**Tests**
manual

### DIFF
--- a/embeddings.go
+++ b/embeddings.go
@@ -159,7 +159,7 @@ type EmbeddingRequest struct {
 	EncodingFormat EmbeddingEncodingFormat `json:"encoding_format,omitempty"`
 	// Dimensions The number of dimensions the resulting output embeddings should have.
 	// Only supported in text-embedding-3 and later models.
-	Dimensions *int `json:"dimensions,omitempty"`
+	Dimensions int `json:"dimensions,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


embeddings.go
```diff
@@ -157,6 +157,9 @@ type EmbeddingRequest struct {
 	Model          EmbeddingModel          `json:"model"`
 	User           string                  `json:"user"`
 	EncodingFormat EmbeddingEncodingFormat `json:"encoding_format,omitempty"`
+	// Dimensions The number of dimensions the resulting output embeddings should have.
+	// Only supported in text-embedding-3 and later models.
+	Dimensions *int `json:"dimensions,omitempty"`
```
sashabaranov, line 162: Thank you for the PR!

Wouldn't `int` work just as well here? `0` would be ignored and all other values be included

chenhhA: Thanks you sugguestion! I chang it type to `int`

sashabaranov (approved):

Issue #633, opened 2024-01-08
Missing moderation categories


richdawe-cio:
**Describe the bug**

The [moderations endpoint](https://platform.openai.com/docs/api-reference/moderations) returns a number of categories that are not supported by go-openai. I'm not going to list them here, but the entire `harrassment` category is missing, along with some subcategories.

**To Reproduce**

Take a look at the categories listed by:

```
curl -H 'Authorization: Bearer API-KEY-GOES-HERE' -H 'Content-Type: application/json' -d '{ "input": "OHAI" }' https://api.openai.com/v1/moderations
```

and compare it with the list in [https://github.com/sashabaranov/go-openai/blob/master/moderation.go](moderations.go)

**Expected behavior**

I expect all moderation categories to be available in the `openai.ResultCategories` and `openai.ResultCategoryScores` structs.

Suggestion: Perhaps the moderation code could use map types, instead of having hard-coded fields. That way, the code would not have to change everytime OpenAI adds a new moderation type.

**Screenshots/Logs**
If applicable, add screenshots to help explain your problem. For non-graphical issues, please provide any relevant logs or stack traces.

**Environment (please complete the following information):**
 - go-openai version: v1.17.5
 - Go version: go1.20.11 darwin/arm64
 - OpenAI API version: latest (as of 2024-01-08)
 - OS: macOS Ventura

**Additional context**

stillmatic:
```
{
  "id": "modr-something",
  "model": "text-moderation-006",
  "results": [
    {
      "flagged": false,
      "categories": {
        "sexual": false,
        "hate": false,
        "harassment": false,
        "self-harm": false,
        "sexual/minors": false,
        "hate/threatening": false,
        "violence/graphic": false,
        "self-harm/intent": false,
        "self-harm/instructions": false,
        "harassment/threatening": false,
        "violence": false
      },
      "category_scores": {
        "sexual": 0.005986555479466915,
        "hate": 0.00019341870211064816,
        "harassment": 0.00013092363951727748,
        "self-harm": 1.1213144262001151e-6,
        "sexual/minors": 0.00009209589188685641,
        "hate/threatening": 3.4858462640841026e-6,
        "violence/graphic": 4.281196652300423e-6,
        "self-harm/intent": 8.618121114523092e-7,
        "self-harm/instructions": 1.192578224618046e-6,
        "harassment/threatening": 2.3499235339841107e-6,
        "violence": 0.000017707703591440804
      }
    }
  ]
}
```

[...]

PR #662, opened 2024-02-17
[BREAKING_CHANGES] converted ResultCategoryScores struct to map

raphoester:
**Describe the change**
I converted the `ResultCategoriesScores` struct into a `map[string]float32`. 
This change has been discussed in issue #633 .

https://platform.openai.com/docs/guides/moderation/overview

**Describe your solution**
I just converted the struct into a map to allow more categories to be accessible without the need of changing the code everytime

**Tests**
I simply adapted the existing tests to the new data structure with more test cases 

Issue: #633

### DIFF
--- a/moderation.go
+++ b/moderation.go
@@ -20,6 +20,20 @@ const (
 	ModerationText001 = "text-moderation-001"
 )
 
+const (
+	ModerationCategoryHate                  = "hate"
+	ModerationCategoryHateThreatening       = "hate/threatening"
+	ModerationCategoryHarassment            = "harassment"
+	ModerationCategoryHarassmentThreatening = "harassment/threatening"
+	ModerationCategorySelfHarm              = "self-harm"
+	ModerationCategorySelfHarmIntent        = "self-harm/intent"
+	ModerationCategorySelfHarmInstructions  = "self-harm/instructions"
+	ModerationCategorySexual                = "sexual"
+	ModerationCategorySexualMinors          = "sexual/minors"
+	ModerationCategoryViolence              = "violence"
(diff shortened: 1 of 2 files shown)


sashabaranov:
Thank you for the PR!

moderation.go
```diff
@@ -44,25 +58,21 @@ type Result struct {
...
-type ResultCategoryScores struct {
-	Hate            float32 `json:"hate"`
-	HateThreatening float32 `json:"hate/threatening"`
-	SelfHarm        float32 `json:"self-harm"`
-	Sexual          float32 `json:"sexual"`
-	SexualMinors    float32 `json:"sexual/minors"`
-	Violence        float32 `json:"violence"`
-	ViolenceGraphic float32 `json:"violence/graphic"`
-}
+type ResultCategoryScores map[string]float32
```
sashabaranov, line 75: Could we just add additional scores to the scores struct? That would not be completely future proof, but would not introduce any breaking changes and keep categories and scores in sync (struct+struct, instead of struct+map)?

raphoester: We could ! 
We could also make the `ResultCategoryScores` struct a map aswell. 

**Pros of having maps** 
- Future proof 
**Cons of having maps**
- Breaking change 
- Might need to keep the former structs as deprecated
- Structs are easier and safer to use

The decision is up to you

sashabaranov: Let's go with structs!

raphoester: alright, reverted to a struct and just added fields

sashabaranov (approved):

PR #668, opened 2024-02-22
Fix for broken Azure Threads url

coggsflod:
Threads API is broken for Azure OpenAI implementation returning a 404 error

Expected URL = https://your_resource_name.openai.azure.com/openai/threads?api-version=2024-02-15-preview
Current URL = https://your_resource_name.openai.azure.com/openai/deployments/UNKNOWN/threads?api-version=2024-02-15-preview

Link to Microsoft docs: https://github.com/MicrosoftDocs/azure-docs/blob/bb63f74fabd8a70267e1ab2960773f706c8c50c3/articles/ai-services/openai/assistants-reference-threads.md

Fix removes /deployments/UNKNOWN/ by using same approach for /models /assistants API

Added unit tests for Azure Threads API

FYI, this fix also addresses same issue with messages & runs APIs

### DIFF
--- a/client.go
+++ b/client.go
@@ -221,7 +221,7 @@ func (c *Client) fullURL(suffix string, args ...any) string {
 		baseURL = strings.TrimRight(baseURL, "/")
 		// if suffix is /models change to {endpoint}/openai/models?api-version=2022-12-01
 		// https://learn.microsoft.com/en-us/rest/api/cognitiveservices/azureopenaistable/models/list?tabs=HTTP
-		if strings.Contains(suffix, "/models") {
+		if strings.Contains(suffix, "/models") || strings.Contains(suffix, "/assistants") {
 			return fmt.Sprintf("%s/%s%s?api-version=%s", baseURL, azureAPIPrefix, suffix, c.config.APIVersion)
 		}
 		azureDeploymentName := "UNKNOWN"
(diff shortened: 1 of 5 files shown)


client.go
```diff
@@ -221,7 +221,8 @@ func (c *Client) fullURL(suffix string, args ...any) string {
 		baseURL = strings.TrimRight(baseURL, "/")
 		// if suffix is /models change to {endpoint}/openai/models?api-version=2022-12-01
 		// https://learn.microsoft.com/en-us/rest/api/cognitiveservices/azureopenaistable/models/list?tabs=HTTP
-		if strings.Contains(suffix, "/models") || strings.Contains(suffix, "/assistants") {
+		if strings.Contains(suffix, "/models") || strings.Contains(suffix, "/assistants") ||
```
sashabaranov, line 224: Thank you for the PR!

Could we refactor this please?

sashabaranov: That would be cool https://freshman.tech/snippets/go/check-if-slice-contains-element/
But I think it makes sense to postpone to keep more go versions supported

coggsflod: @sashabaranov are you on-board with using the golang.org/x/exp/slices package to support 1.18+ ?

sashabaranov: @coggsflod let's just add a small helper functions without any extra dependencies

coggsflod: @sashabaranov actually all these [options](https://freshman.tech/snippets/go/check-if-slice-contains-element/) perform an exact match check rather than the strings.Contains which checks whether substr is within passed string.

Happy to create a containsSubstr helper if that works?

coggsflod:
@sashabaranov any further comments / issues you'd like to address?

sashabaranov (approved):

PR #669, opened 2024-02-22
fix: when no Assistant Tools are specified, an empty list should be sent

guillaume-dussault:
**Describe the change**
Currently, it is not possible to remove all tools when modifying an assistant because the field will be omitted if empty. With this change, it will now be possible.

https://platform.openai.com/docs/api-reference/assistants/modifyAssistant#assistants-modifyassistant-tools

**Describe your solution**
When creating or modifying an assistant, an empty list of tools will be sent instead of omitting the field.

**Tests**
- Created an assistant with no tools
- Created an assistant with tools
- Updated an assistant with tools to remove all tools
- Updated an assistant with tools to remove some tools
- Updated an assistant without tools to add some tools
- Ran unit tests

### DIFF
--- a/assistant.go
+++ b/assistant.go
@@ -46,7 +46,7 @@ type AssistantRequest struct {
 	Name         *string         `json:"name,omitempty"`
 	Description  *string         `json:"description,omitempty"`
 	Instructions *string         `json:"instructions,omitempty"`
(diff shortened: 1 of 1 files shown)


sashabaranov (approved):
Thank you for the PR!

PR #670, opened 2024-02-28
fix: repair json decoding of moderation response

mheck136:
PR #662 has broken the json decoding of moderation responses by changing the type of score values from float32 to bool. This commit fixes that problem.


sashabaranov (approved):
🤦🏻‍♂️how could've I miss that! Thanks a lot!

sashabaranov:
We need to enable integration tests to prevent that going forward

PR #671, opened 2024-03-01
Added 'wav' and 'pcm' Audio Formats

LinuxSploit:
**Describe the change**
Added "wav" and "pcm" audio formats as per OpenAI API documentation for createSpeech endpoint.  Ref: https://platform.openai.com/docs/api-reference/audio/createSpeech 
Supported formats are mp3, opus, aac, flac, wav, and pcm.

**OpenAI documentation link**
[ Relevant API doc from https://platform.openai.com/docs/api-reference
 ](https://platform.openai.com/docs/api-reference/audio/createSpeech)

### DIFF
--- a/speech.go
+++ b/speech.go
@@ -35,7 +35,6 @@ const (
 	SpeechResponseFormatFlac SpeechResponseFormat = "flac"
 	SpeechResponseFormatWav SpeechResponseFormat = "wav"
 	SpeechResponseFormatPcm SpeechResponseFormat = "pcm"
-
 )
 
 var (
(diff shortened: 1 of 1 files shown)


speech.go
```diff
@@ -33,6 +33,8 @@ const (
 	SpeechResponseFormatOpus SpeechResponseFormat = "opus"
 	SpeechResponseFormatAac  SpeechResponseFormat = "aac"
 	SpeechResponseFormatFlac SpeechResponseFormat = "flac"
+	SpeechResponseFormatWav SpeechResponseFormat = "wav"
+	SpeechResponseFormatPcm SpeechResponseFormat = "pcm"
```
sashabaranov, line 37: @LinuxSploit thank you for the PR! Please run goimports so the file would get accepted by the linter!

PR #681, opened 2024-03-11
Update streamReader Close() method to return error

sunshineplan:
This commit updates the Close() method of the streamReader struct to return an error. This allows callers to handle any errors that occur when closing the stream.


sashabaranov (approved):
Thank you for the PR!

Issue #682, opened 2024-03-14
Add Support for Groq's Hosted Models via Groq's OpenAI Compatibility API 


jackspirou:
**Is your feature request related to a problem? Please describe.**
I want the option to use Go to access Groq hosted models, due to their greater inference speed.

**Describe the solution you'd like**
I would like to suggest this project do two things: 
1.  Use Groq's OpenAI Compatibility for minimal changes
2. Provide a list of Groq models for easy usage
3. Add basic documentation since this more of a side-feature with huge value than a main case. 

**Describe alternatives you've considered**
This could be a new/different library, but I think it makes sense more sense to build on the work done by this project. 

**Additional context**
- Groq's OpenAI Compatibility: https://console.groq.com/docs/openai
> OpenAI Compatibility
Groq's APIs are designed to be compatible with OpenAI's, with the goal of making it easy to leverage Groq in applications you may have already built. 
- Groq's Hosted Models: https://console.groq.com/docs/models
> Supported Models
GroqCloud currently supports the following models:
[LLaMA2-70b](https://console.groq.com/docs/openai#llama270b)
Developer: Meta
Model Name: LLaMA2-70b-chat
Context Window: 4,096 tokens
API String: llama2-70b-4096
[Mixtral-8x7b](https://console.groq.com/docs/openai#mixtral8x7b)
Developer: Mistral
Model Name: Mixtral-8x7b-Instruct-v0.1
Context Window: 32,768 tokens
API String: mixtral-8x7b-32768
[Gemma-7b-it](https://console.groq.com/docs/openai#gemma7bit)
Developer: Google
Model Name: Gemma-7b-it
Context Window: 8,192 tokens
API String: gemma-7b-it

These are chat type models and are directly accessible through the GroqCloud Models API endpoint using their respective API strings.

I'll submit a PR that outlines this change shortly.

PR #683, opened 2024-03-14
PR: Add Support for Groq's Hosted Models via Groq's OpenAI Compatibility API

jackspirou:
- Add Groq model options via Groq's OpenAI API compatibility
- Update documentation accordingly
- Fix example

Fixes https://github.com/sashabaranov/go-openai/issues/682


sashabaranov:
@jackspirou thank you for the PR!

I don't think Groq needs any special treatment, as there are multiple services with OpenAI-compatible API now.

It requires a minimal four-line change to the basic example we have in [README](https://github.com/sashabaranov/go-openai?tab=readme-ov-file#chatgpt-example-usage) to make it work with Groq (or any other service like Ollama for that matter). I'm totally open to expanding our README with such an example, though!

```diff
--- basic.go	2024-03-15 14:40:43
+++ groq.go	2024-03-15 14:40:29
@@ -7,11 +7,14 @@
 )

 func main() {
-	client := openai.NewClient("your token")
+	config := openai.DefaultConfig("token")
+	config.BaseURL = "https://api.groq.com/openai/v1"
+	client := openai.NewClientWithConfig(config)
+
 	resp, err := client.CreateChatCompletion(
 		context.Background(),
 		openai.ChatCompletionRequest{
-			Model: openai.GPT3Dot5Turbo,
+			Model: "mixtral-8x7b-32768",
 			Messages: []openai.ChatCompletionMessage{
 				{
 					Role:    openai.ChatMessageRoleUser,```

jackspirou:
@sashabaranov - Thanks for the response and thanks for the example!  I didn't understand the forest for the trees here.  I agree with your assessment above, given the example you shared. 

I'll update this PR to take a stab at a simple readme update change per your advice. 

One side-note - it might be nice to update the pattern of: 
`openai.NewClient(authToken string)` to something like
`openai.NewClient(authToken string, optons ...func(*Client))`.  

I think it might have provided some hints for me to see the forrest for the trees in this case.  Maybe it's something you've already considered.

donvito:
was this released?

Issue #684, opened 2024-03-14
Recent PR breaks requests to modify Assistant (breaks our production deployment)


qhenkart:
**Describe the bug**
A breaking change / bug was introduced with https://github.com/sashabaranov/go-openai/pull/669 when modifying an Assistant, we now regularly get this error `Invalid type for 'tools': expected an array of tools, but got null instead.` That's because an uninitialized Slice in Go equates to nil even if it's not a pointer. 

The Linked PR suggests that an empty list will be supplied by default, however null is being sent instead which is not allowed in the API.

<img width="691" alt="image" src="https://github.com/sashabaranov/go-openai/assets/5798752/4be25378-13b0-409d-9096-d662bc4c5f0e">

**To Reproduce**
1. create an assistant with instructions and tools
2. modify the assistant without any tools specified
```
req := openai.AssistantRequest{Instructions: &instructions}
	if _, err := h.OpenAIClient.ModifyAssistant(ctx, assistantID, req); err != nil {
		panic(err)
	}
// panic " Invalid type for 'tools': expected an array of tools, but got null instead."
```

**Expected behavior**
The API clearly demonstrates that the Tools payload parameter is an optional field. The linked PR breaks with the API by requiring an empty array of tools or all of the existing tools to avoid deleting them

**Screenshots/Logs**
screenshot of API above

**Environment (please complete the following information):**
 - go-openai version: [e.g. v1.20.3]
 - Go version: [e.g. 1.21]
 - OpenAI API version: [latest]
 - OS: [Linux]

qhenkart:
If the linked PR worked, it would automatically delete all tools anytime instructions are updated which is not how the API works, and would have been a much more disastrous breaking change

More repro to show how the API works:

original Assistant
{
   instructions:  "original instructions",
   tools: [{...tool1}, {...tool2}]
}

```
curl --request POST \
  --url https://api.openai.com/v1/assistants/{assistantId} \
  --header 'Content-Type: application/json' \
  --header 'authorization: bearer {token} \
  --header 'openai-beta: assistants=v1' \
  --data '{
	"instructions": "new instructions",
}'

```
returns
```
{
   instructions:  "new instructions",
   tools: [{...tool1}, {...tool2}]
}
```

delete tools
```
curl --request POST \
  --url https://api.openai.com/v1/assistants/{assistantId} \
  --header 'Content-Type: application/json' \
  --header 'authorization: bearer {token} \
  --header 'openai-beta: assistants=v1' \
[...]

PR #685, opened 2024-03-14
Fix broken implementation AssistantModify implementation

qhenkart:
This resolves https://github.com/sashabaranov/go-openai/issues/684 
both reverts the changes in https://github.com/sashabaranov/go-openai/pull/669
and modifies the code to actually match the intention of the https://github.com/sashabaranov/go-openai/pull/669

Currently the SDK does not match the API and requires all of the existing tools to be provided in every request to avoid an API error or deleting all of the tools. Please see the linked issue for repro steps and use cases including step by step cURL requests to demonstrate the existing functionality of the API

As specified in the inline documentation
A custom marshaller was required here to handle the various use cases of the API request with respect to GO-isms and to avoid breaking changes:

If Tools is nil (or undefined), the field is omitted from the JSON. > this reflects the functionality of the API and reverts the breaking change introduced in https://github.com/sashabaranov/go-openai/pull/669

If Tools is an empty slice, it's included in the JSON as an empty array ([]). > this matches the intention of https://github.com/sashabaranov/go-openai/pull/669 by allowing the ability to remove all Tools, something that was broken previously and the linked PR attempted to resolve
If Tools is populated, it's included in the JSON with the elements > and thusly replaces the Tools. This functionality existed already and simply is maintained in this PR

This is important to note because in Golang, omitempty will omit both a nil (uninitialized) slice as well as an empty slice. To provide support to allow both a non-existant slice, and an empty slice, and a populated slice and never a nil slice (see the linked issue to see the API error if it receives nil). A custom Marshaller and aliast was required

Additionally I wrapped the tests block into Run blocks to be able to isolate tests and remove variable shadowing. I also added explicit unit tests for this functionality, and the run blocks were necessary to be able to properly isolate them for testing and development.

### DIFF
--- a/assistant.go
+++ b/assistant.go
@@ -2,6 +2,7 @@ package openai
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -21,7 +22,7 @@ type Assistant struct {
 	Description  *string         `json:"description,omitempty"`
 	Model        string          `json:"model"`
 	Instructions *string         `json:"instructions,omitempty"`
-	Tools        []AssistantTool `json:"tools,omitempty"`
+	Tools        []AssistantTool `json:"tools"`
 	FileIDs      []string        `json:"file_ids,omitempty"`
 	Metadata     map[string]any  `json:"metadata,omitempty"`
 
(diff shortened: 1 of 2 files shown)


sashabaranov (approved):
Thank you for such a thorough PR!

We might apply a similar approach to a long-standing temperature issue https://github.com/sashabaranov/go-openai/issues/9

PR #691, opened 2024-03-22
Allow custom Voice and Speech Models

adrianliechti:
this pull request removes the client-side logic to verify a valid speech and voice model.
the motivation is to support custom models on openai compatible backends

thanks for considering

### DIFF
--- a/speech.go
+++ b/speech.go
@@ -2,7 +2,6 @@ package openai
 
 import (
 	"context"
-	"errors"
 	"io"
 	"net/http"
 )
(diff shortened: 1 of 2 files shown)


adrianliechti:
@sashabaranov thanks a ton for coming back on this!
my own one adapts tools like mimic3, coqui.ai, whisper.cpp compatible ones into an openai-compatible api (also multiple ones). here a link: https://github.com/adrianliechti/llama. It is in use as part of a commercial platform.
and since I was using your lovely SDK for both tests but also integration, I allowed me to create this PR.

I think it is in line with this project's recent changes on embedding/models - so I hope over time a broader audience benefits (or at least needs less urgent maintenance once openai releases new whisper or tts models)

adrianliechti:
@sashabaranov
Had you already found the time to consider this change?

sashabaranov (approved):
Sorry for the delay, and thank you so much for the PR! ❤️

adrianliechti:
you made my day! cheers to you

PR #694, opened 2024-03-27
feat: get header from sendRequestRaw

WqyJh:
Add response header to `CreateSpeech` and `GetFileContent`.

Replace `io.ReadCloser` with `RawResponse`, which contains an `io.ReadCloser` and a `httpHeader`. The `RawResponse` can be treated as `io.ReadCloser` therefore keeping backwards compatibility.

### DIFF
--- a/client.go
+++ b/client.go
@@ -38,6 +38,12 @@ func (h *httpHeader) GetRateLimitHeaders() RateLimitHeaders {
 	return newRateLimitHeaders(h.Header())
 }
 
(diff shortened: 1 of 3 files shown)


WqyJh:
Hi @sashabaranov, would you please checkout this PR?

sashabaranov:
@WqyJh sorry for taking so long and thank you for the PR! Looks good to me!

sashabaranov (approved):

PR #698, opened 2024-03-27
feat: support for Assistant stream mode and implemented stream event callbacks

push-edp:
This update enhances the existing functionality by adding support for Assistant stream mode and stream event callbacks without introducing any breaking changes or modifications to the existing API endpoints.

The implemented features are in line with the OpenAI Assistant streaming API, as documented here: [OpenAI Assistant Streaming API](https://platform.openai.com/docs/api-reference/assistants-streaming)

### DIFF
--- a/assistant_stream.go
+++ b/assistant_stream.go
@@ -1,0 +1,79 @@
+package openai
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+)
+
+/*
+	type AssistantThreadRunStreamDelta struct {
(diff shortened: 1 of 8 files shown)


sashabaranov:
@push-edp thank you for the PR! Could you please add tests so coverage check would pass?

push-edp:
> @push-edp thank you for the PR! Could you please add tests so coverage check would pass?

Sure, I'll add the tests shortly.

Issue #692, opened 2024-03-24
ModifyMessage failed with error code 400, message: Unknown parameter


Ahmadhegavf:
**Bug Description:**
The `ModifyMessage` function from the `go-openai` library expects a parameter `metadata` of type `map[string]any`, but the API it interfaces with, specifically the [Modify Message endpoint](https://platform.openai.com/docs/api-reference/messages/modifyMessage), requires the metadata to be wrapped in a JSON object with a field named `"metadata"`. This discrepancy causes the API call to fail due to an unknown parameter.

**To Reproduce:**
Call the `ModifyMessage` function with a non-empty `metadata` parameter.

**Expected Behavior:**
The method should serialize the metadata into a JSON object with a field named `"metadata"` and send it to the API, aligning with the API's expectations.

**Environment:**
- `go-openai` version: v1.120.3
- Go version: go1.21.3
- OpenAI API version: v1
- OS: Ubuntu 20.04

qhenkart:
I just ran into this as well and you was going to make an issue. I can confirm this report

PR #699, opened 2024-03-29
[BREAKING_CHANGES] Fix update message payload

qhenkart:
**Describe the change**
Resolves https://github.com/sashabaranov/go-openai/issues/692
The ModifyMessage endpoint improperly wraps the payload. Including the expected map[string]any object into the method will result in an API error as the API requires it to be wrapped in "metadata".

This is a breaking change because to have successfully used this endpoint, developers would have run into this API error, debugged it and realized that the payload isn't being wrapped by the underlying code as expected by the API. 

https://platform.openai.com/docs/api-reference/messages/modifyMessage

**Describe your solution**
This simply wraps the payload as expected by the API and as assumed by the interface. The interface accepts a metadata object, either the interface needs to change to accept a generic payload object (which would be equally misleading). Or we update the underlying code to properly wrap the payload as expected

**Tests**
Unit tests

**Additional context**
We will need to inform that anyone who figured out how to use this endpoint will need to update their code to remove the manual wrapping of the payload.

Issue: #692

Alternatively I can avoid the breaking change by only wrapping the payload if it's not already wrapped. I think this will create complexity down the road, but it is an option

### DIFF
--- a/assistant.go
+++ b/assistant.go
@@ -2,6 +2,7 @@ package openai
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -21,7 +22,7 @@ type Assistant struct {
 	Description  *string         `json:"description,omitempty"`
 	Model        string          `json:"model"`
 	Instructions *string         `json:"instructions,omitempty"`
-	Tools        []AssistantTool `json:"tools,omitempty"`
+	Tools        []AssistantTool `json:"tools"`
 	FileIDs      []string        `json:"file_ids,omitempty"`
 	Metadata     map[string]any  `json:"metadata,omitempty"`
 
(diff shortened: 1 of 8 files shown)


qhenkart:
I'm not sure why the Linter is breaking randomly. Did something change in the underlying build process? It doesn't look like it's due to any changes in this PR

```
golangci-lint run ./...
chat_test.go:330:1: cognitive complexity 30 of func `TestMultipartChatMessageSerialization` is high (> 20) (gocognit)
func TestMultipartChatMessageSerialization(t *testing.T) {
^
fine_tunes.go:118:2: directive `//nolint:goconst // Decreases readability` is unused for linter "goconst" (nolintlint)
	//nolint:goconst // Decreases readability
	^
messages.go:21:49: directive `//nolint:revive //backwards-compatibility` is unused for linter "revive" (nolintlint)
	FileIds     []string         `json:"file_ids"` //nolint:revive //backwards-compatibility
	                                               ^
messages.go:57:54: directive `//nolint:revive // backwards-compatibility` is unused for linter "revive" (nolintlint)
	FileIds  []string       `json:"file_ids,omitempty"` //nolint:revive // backwards-compatibility
	                                                    ^
run.go:27:54: directive `//nolint:revive // backwards-compatibility` is unused for linter "revive" (nolintlint)
	FileIDS        []string           `json:"file_ids"` //nolint:revive // backwards-compatibility
```

qhenkart:
Also worth noting. metadata should be a map[string]string. Non string types are not accepted and can also lead to an error. Would you like me to include that in this PR? It will lead to more breaking changes, but also enforce the requirements of the API which are currently unenforced and lead to error

sashabaranov:
@qhenkart thank you for the PR! The master branch contains the fix for linter, you can merge/rebase it here so the sanity check would pass!

qhenkart:
@sashabaranov Okay thanks, this PR is ready for your final review

sashabaranov (approved):

PR #702, opened 2024-04-09
Add prompt filter results to content filters

jchaykow:
**Describe the change**
Adding field for `prompt_filter_results` which is in the open content filters to assess the content of the user input. Current implementation in go-openai will only return content filter information on the model output using `ChatCompletionStreamResponse`. This is for AzureOpenAI spec.

Relevant API doc: https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter
[See Bottom under Annotations]

**Describe your solution**
Change struct


sashabaranov (approved):
Thank you!

Issue #717, opened 2024-04-18
Assistants version migration `v1` ➡️ `v2`


danai-antoniou:
OpenAI released assistants `v2`, and the corresponding migration instructions https://platform.openai.com/docs/assistants/migration, which say:
```
Simply pass the `v2` beta version header
```

However, at the moment the `v1` version of assistants is hardcoded within every assistant method 😬 

Edit: [PR open](https://github.com/sashabaranov/go-openai/pull/719)

Roockeeeet:
# Jarvis DApp

Jarvis is a decentralized application (DApp) built on the Ethereum blockchain, partnered with OpenAI.

## Smart Contracts

- `MyContract.sol`: Main smart contract for the Jarvis DApp.

## Frontend

- `index.html`: HTML file for the frontend.
- `styles.css`: CSS file for styling the frontend.
- `script.js`: JavaScript file for frontend logic.

## Testing

- Automated tests for smart contracts and frontend components.

## Deployment

- Deployment scripts and instructions for deploying the Jarvis DApp.

## Usage

- Instructions for using the Jarvis DApp.

## Contributors

- List of contributors to the Jarvis DApp project.

## License

- License information for the Jarvis DApp.

Created with Ethereum. Partnered by [OpenAI](https://platform.openai.io).

Roockeeeet:
from openai import OpenAI
client = OpenAI()

response = client.chat.completions.create(
  model="gpt-3.5-turbo",
  messages=[
    {
      "role": "system",
[...]

PR #719, opened 2024-04-18
Remove hardcoded assistants version

danai-antoniou:
**Describe the change**
OpenAI released assistants v2, however the code currently hardcodes `v1` in every assistant and thread method.

https://platform.openai.com/docs/assistants/migration

**Describe your solution**
I've added an `AssistantVersion` parameter in the client config, which defaults to `v1`, so there will be no breaking change for anyone.
Users will be able to use `v2` and any future versions by passing the version in the client config.

**Tests**
I've been using this in our fork of `go-openai`.

Issue: #717

### DIFF
--- a/assistant.go
+++ b/assistant.go
@@ -11,7 +11,6 @@ import (
 const (
 	assistantsSuffix      = "/assistants"
 	assistantsFilesSuffix = "/files"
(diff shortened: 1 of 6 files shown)


sashabaranov (approved):
Thank you for the PR!

PR #725, opened 2024-04-24
feat: make RequestBuilder configurable

WqyJh:
`RequestBuilder` is an interface, therefore, it's designed to be extensible.

```go
type RequestBuilder interface {
	Build(ctx context.Context, method, url string, body any, header http.Header) (*http.Request, error)
}
```

However, currently it's a private field of `Client`, which make it impossible to customize.

```go
type Client struct {
	config ClientConfig

	requestBuilder    utils.RequestBuilder
	createFormBuilder func(io.Writer) utils.FormBuilder
}
```

This PR moves `requestBuilder` to `ClientConfig`, making it configurable, so that users can implement their own `RequestBuilder`, which is very useful in situation of adding custom request headers and injecting tracing attributes.

### DIFF
--- a/client.go
+++ b/client.go
@@ -16,7 +16,6 @@ import (
 type Client struct {
 	config ClientConfig
 
-	requestBuilder    utils.RequestBuilder
 	createFormBuilder func(io.Writer) utils.FormBuilder
 }
 
@@ -52,9 +51,11 @@ func NewClient(authToken string) *Client {
 
 // NewClientWithConfig creates new OpenAI API client for specified config.
 func NewClientWithConfig(config ClientConfig) *Client {
+	if config.RequestBuilder == nil {
+		config.RequestBuilder = utils.NewRequestBuilder()
+	}
 	return &Client{
-		config:         config,
-		requestBuilder: utils.NewRequestBuilder(),
+		config: config,
 		createFormBuilder: func(body io.Writer) utils.FormBuilder {
 			return utils.NewFormBuilder(body)
 		},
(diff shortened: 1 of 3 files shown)


sashabaranov:
@WqyJh it is possible to modify requests with custom `Transport` in `http.Client`. We have a basic proxy example in our README, and here's an [example](https://uptrace.dev/get/instrument/opentelemetry-net-http.html#instrumenting-http-client) for telemetry (+ entire transport [implementation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/transport.go#L113))

On another note, I think the biggest problem with custom request builders is that users would need to substitute full request builder logic every time:

https://github.com/sashabaranov/go-openai/blob/c84ab5f6ae8da3a78826ed2c8dc4c5cf93e30589/internal/request_builder.go#L31-L50

It might make sense to introduce instrumentation just before request is [sent](https://github.com/sashabaranov/go-openai/blob/c84ab5f6ae8da3a78826ed2c8dc4c5cf93e30589/client.go#L125) similar to gRPC interceptors [concept](https://github.com/grpc-ecosystem/go-grpc-middleware?tab=readme-ov-file#go-grpc-middleware). That might have better ergonomics than implementing custom HTTP transport.

WqyJh:
Finally got time to try this solution. It's easy and works just the same. Thank you for your great idea!

Here's a short example:
```go
var _ http.RoundTripper = (*Transport)(nil)

type Transport struct {
	rt http.RoundTripper
}

func NewTransport(rt http.RoundTripper) *Transport {
	return &Transport{rt: rt}
}

func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
	req.Header.Set("X-Custom", "true")

	return t.rt.RoundTrip(req)
}
```

PR #731, opened 2024-04-29
feat: Assistant stream mode and event callback

CallOrRet:
This update enhances the existing functionality by adding support for Assistant stream mode and stream event callbacks without introducing any breaking changes or modifications to the existing API endpoints.

The implemented features are in line with the OpenAI Assistant streaming API, as documented here: [OpenAI Assistant Streaming API](https://platform.openai.com/docs/api-reference/assistants-streaming)

```
Stream.On("event1", callback1)
Stream.On("event2", callback2)
Stream.Run()
```

### DIFF
--- a/assistant_stream.go
+++ b/assistant_stream.go
@@ -1,0 +1,82 @@
+package openai
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+var (
+	ErrStreamEventEmptyTopic    = errors.New("stream event gets a empty topic")
+	ErrStreamEventCallbackPanic = errors.New("occurs a runtime panic during the stream event callback")
+)
+
+type AssistantThreadRunStreamResponse struct {
+	ID     string       `json:"id"`
+	Object string       `json:"object"`
+	Delta  MessageDelta `json:"delta,omitempty"`
+}
+
+type AssistantThreadRunStream struct {
+	*streamReader[AssistantThreadRunStreamResponse]
+}
+
+func (c *Client) CreateAssistantThreadRunStream(
+	ctx context.Context,
+	threadID string,
+	request RunRequest,
+) (stream *AssistantThreadRunStream, err error) {
+	request.Stream = true
+	urlSuffix := fmt.Sprintf("/threads/%s/runs", threadID)
+	req, err := c.newRequest(
+		ctx,
+		http.MethodPost,
+		c.fullURL(urlSuffix),
+		withBody(request),
+		withBetaAssistantVersion(c.config.AssistantVersion),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := sendRequestStream[AssistantThreadRunStreamResponse](c, req)
+	if err != nil {
+		return nil, err
+	}
+
+	stream = &AssistantThreadRunStream{
+		streamReader: resp,
+	}
+	return
+}
+
+func (c *Client) CreateAssistantThreadRunToolStream(
+	ctx context.Context,
+	threadID string,
+	runID string,
+	request SubmitToolOutputsRequest,
+) (stream *AssistantThreadRunStream, err error) {
+	request.Stream = true
+	urlSuffix := fmt.Sprintf("/threads/%s/runs/%s/submit_tool_outputs", threadID, runID)
+	req, err := c.newRequest(
+		ctx,
+		http.MethodPost,
+		c.fullURL(urlSuffix),
+		withBody(request),
+		withBetaAssistantVersion(c.config.AssistantVersion),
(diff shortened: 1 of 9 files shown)


sashabaranov:
Thank you for the PR! Sorry for a long review process :D

assistant_stream.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+var (
+	ErrStreamEventEmptyTopic    = errors.New("stream event gets a empty topic")
```
sashabaranov, line 11:
```suggestion
	ErrStreamEventEmptyTopic    = errors.New("stream event gets an empty topic")
```

stream_reader.go
```diff
@@ -28,9 +30,56 @@ type streamReader[T streamable] struct {
...
+	stream.handlers[topic] = handler
+	return nil
+}
+
+func (stream *streamReader[T]) Run() error {
+	if stream.handlerCtx == nil {
+		stream.event = "message" // default event for chat completion stream
+		ctx, cancel := context.WithCancel(context.Background())
+		stream.handlerCtx = ctx
+		go func() {
```
sashabaranov, line 60: Sorry, I don't think we're going to merge this as starting goroutines don't fit into the design of this library.

Please refer to the following comments:
- https://github.com/sashabaranov/go-openai/pull/54#discussion_r1083519779
- https://github.com/sashabaranov/go-openai/issues/49#issuecomment-1373423407

sashabaranov: I'm happy to have this in the `examples/` folder, though!

CallOrRet: No problem. What if I revise the parts where I used goroutines? Would that work better with the design of the library?

sashabaranov: @CallOrRet I'm totally open to that!

assistant_stream.go
```diff
@@ -0,0 +1,82 @@
...
+var (
+	ErrStreamEventEmptyTopic    = errors.New("stream event gets an empty topic")
+	ErrStreamEventCallbackPanic = errors.New("occurs a runtime panic during the stream event callback")
+)
+
+type AssistantThreadRunStreamResponse struct {
+	ID     string       `json:"id"`
+	Object string       `json:"object"`
+	Delta  MessageDelta `json:"delta,omitempty"`
+}
```
sauravexodus, line 19: This is a very trimmed down version of the response model.

assistant_stream.go
```diff
@@ -0,0 +1,82 @@
...
+
+	resp, err := sendRequestStream[AssistantThreadRunStreamResponse](c, req)
+	if err != nil {
+		return nil, err
+	}
+
+	stream = &AssistantThreadRunStream{
+		streamReader: resp,
+	}
+	return
```
sauravexodus, line 51:
```suggestion
	return stream, nil
```

assistant_stream.go
```diff
@@ -0,0 +1,82 @@
...
+	urlSuffix := fmt.Sprintf("/threads/%s/runs/%s/submit_tool_outputs", threadID, runID)
+	req, err := c.newRequest(
+		ctx,
+		http.MethodPost,
+		c.fullURL(urlSuffix),
+		withBody(request),
+		withBetaAssistantVersion(c.config.AssistantVersion),
+	)
+	if err != nil {
+		return
```
sauravexodus, line 70:
```suggestion
		return nil, err
```

assistant_stream.go
```diff
@@ -0,0 +1,82 @@
...
+
+	resp, err := sendRequestStream[AssistantThreadRunStreamResponse](c, req)
+	if err != nil {
+		return nil, err
+	}
+
+	stream = &AssistantThreadRunStream{
+		streamReader: resp,
+	}
+	return
```
sauravexodus, line 81:
```suggestion
	return stream, nil
```

sauravexodus (commented):
There are some commented code in this PR as well. But the bigger concern is the incomplete response model.

assistant_stream.go
sauravexodus: The complete response has a lot many parameters which would be important to run function calling etc. https://platform.openai.com/docs/api-reference/runs/createRun

run.go
```diff
@@ -137,6 +138,7 @@ type RunList struct {
 
 type SubmitToolOutputsRequest struct {
 	ToolOutputs []ToolOutput `json:"tool_outputs"`
+	Stream      bool
```
sauravexodus, line 141:
```suggestion
	Stream      bool              `json:"stream,omitempty"`
```

PR #732, opened 2024-05-02
Add preliminary MistralAI support

iMilnb:
This small patch provides the ability to use another API endpoint declared in an environment variable, and adds Mistral models.

Here is the python example from MistralAI: https://github.com/mistralai/client-python/blob/0d55818364fd2a7ca67401dd550632f8ca70e5af/examples/chatbot_with_streaming.py#L15

Successfully tested with MistralAI.

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"strings"
 
 	utils "github.com/sashabaranov/go-openai/internal"
@@ -224,6 +225,9 @@ func decodeString(body io.Reader, output *string) error {
 // fullURL returns full URL for request.
 // args[0] is model name, if API type is Azure, model name is required to get deployment name.
 func (c *Client) fullURL(suffix string, args ...any) string {
+	if val := os.Getenv("OPENAI_BASEURL"); val != "" {
+		baseURL = val
+	}
 	// /openai/deployments/{model}/chat/completions?api-version={api_version}
 	if c.config.APIType == APITypeAzure || c.config.APIType == APITypeAzureAD {
 		baseURL := c.config.BaseURL
--- a/client.go
+++ b/client.go
@@ -226,7 +226,7 @@ func decodeString(body io.Reader, output *string) error {
 // args[0] is model name, if API type is Azure, model name is required to get deployment name.
 func (c *Client) fullURL(suffix string, args ...any) string {
 	if val := os.Getenv("OPENAI_BASEURL"); val != "" {
-		baseURL = val
+		c.config.baseURL = val
 	}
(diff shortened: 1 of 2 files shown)


kmesiab:
I think this PR signals a mature direction to start thinking about moving this project.

iMilnb:
> I think this PR signals a mature direction to start thinking about moving this project.

Actually, I think the model declaration needs to be revamped. `go-openai` would certainly work with [ollama](https://ollama.com/blog/openai-compatibility) just like _Mistral_, but adding models in the `completion.go` file is not sustainable. The `const` list should be extendable via an environment variable or a config file.  
I can work on this it you agree with this idea.

sashabaranov:
@iMilnb Thank you for the PR! We're definitely not merging that, as implicit dependency on the environment variable is not what we want. (And I'm getting strong xz vibes from this PR tbh).

Regarding the list of consts: there is no need to rely on it, the `Model` [parameter](https://pkg.go.dev/github.com/sashabaranov/go-openai#ChatCompletionRequest) is just a string, and you can pass any model name you want.

client.go
```diff
@@ -224,6 +225,9 @@ func decodeString(body io.Reader, output *string) error {
 // fullURL returns full URL for request.
 // args[0] is model name, if API type is Azure, model name is required to get deployment name.
 func (c *Client) fullURL(suffix string, args ...any) string {
+	if val := os.Getenv("OPENAI_BASEURL"); val != "" {
+		c.config.BaseURL = val
```
sashabaranov, line 229: The `fullURL` method is getting called during the usage of the client (days, weeks), so this change allows you to quietly change the base url using an environment variable from anywhere if you could just set an environment variable.

iMilnb: I didn't realize that, my bad. I wanted to provide a non disruptive and quick way of supporting another API endpoint.  
Is alternative API endpoint support planned somehow or is this project only focusing at OpenAI?

sashabaranov: You probably want something like that:

```go
config := openai.DefaultConfig("token")
config.BaseURL = "your url" // probably "https://api.mistral.ai" 

response, err := client.CreateChatCompletion(
	context.Background(),
	openai.ChatCompletionRequest{
		Model: "mistral-small",
			{
				Content: "Hello!",
			},
		},
	},
)


```

iMilnb: Ok I totally missed the `NewClientWithConfig()` function, thanks for clarifying.  
As per the `xz` vibes, I use `go-openai` in this little project of mine: https://gitlab.com/iMil/cbg  
Again, sorry for the misunderstanding and keep up the good work.

kevin-mesiab:
> Regarding the list of consts: there is no need to rely on it.

Just trying to understand the design philosophy here.  Is the goal to prevent the constants list from being polluted with non openai models?

PR #744, opened 2024-05-13
enhancement: Add new GPT4-o and alias to completion enums

kmesiab:
This PR introduces two new enums for OpenAI's new model release: `GPT4-o`

https://platform.openai.com/docs/models/gpt-4o


sashabaranov (approved):
Thank you for a swift update!

Issue #724, opened 2024-04-22
support the new batch api


yingzhuivy:
Your issue may already be reported!
Please search on the [issue tracker](https://github.com/sashabaranov/go-openai/issues) before creating one.

**Is your feature request related to a problem? Please describe.**
OpenAI launched batch API: https://community.openai.com/t/batchapi-is-now-available/718416
could be very useful and it would be good if the library supports it
https://platform.openai.com/docs/api-reference/batch/create

**Describe the solution you'd like**
support of the batch api

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

ricboer0:
I too think this is very much needed and extremely useful!!!

PR #746, opened 2024-05-17
feat: support batches api

eiixy:
support batches api #724


animalnots:
can this be looked into and merged, since it's quite a useful cost effective feature

sashabaranov:
@eiixy thank you so much for the PR and sorry for such a long review!

batch.go
```diff
@@ -0,0 +1,273 @@
...
+
+type CreateBatchWithUploadFileRequest struct {
+	Endpoint         BatchEndpoint  `json:"endpoint"`
+	CompletionWindow string         `json:"completion_window"`
+	Metadata         map[string]any `json:"metadata"`
+	UploadBatchFileRequest
+}
+
+// CreateBatchWithUploadFile — API call to Create batch with upload file.
+func (c *Client) CreateBatchWithUploadFile(
```
sashabaranov, line 193: I'm not sure we need this method. Do you think it would be used in >90% of cases when people use batches?

An alternative would be to have a README example of how to upload file and create a batch

batch.go
```diff
@@ -0,0 +1,273 @@
...
+	httpHeader
+	Object  string  `json:"object"`
+	Data    []Batch `json:"data"`
+	FirstID string  `json:"first_id"`
+	LastID  string  `json:"last_id"`
+	HasMore bool    `json:"has_more"`
+}
+
+// ListBatch API call to List batch.
+func (c *Client) ListBatch(ctx context.Context, after *string, limit *int) (response ListBatchResponse, err error) {
```
sashabaranov, line 252: Should we use the [Pagination](https://github.com/sashabaranov/go-openai/blob/211cb49fc22766f4174fef15301c4d39aef609d3/run.go#L210) struct here?

batch.go
```diff
@@ -0,0 +1,273 @@
...
+	"net/url"
+)
+
+const batchesSuffix = "/batches"
+
+type BatchEndpoint string
+
+const BatchEndpointChatCompletions BatchEndpoint = "/v1/chat/completions"
+const BatchEndpointCompletions BatchEndpoint = "/v1/completions"
+const BatchEndpointEmbeddings BatchEndpoint = "/v1/embeddings"
```
sashabaranov, line 19: Possibly:

```go
type BatchEndpoint string

const (
	BatchEndpointChatCompletions BatchEndpoint = "/v1/chat/completions"
	BatchEndpointCompletions     BatchEndpoint = "/v1/completions"
	BatchEndpointEmbeddings      BatchEndpoint = "/v1/embeddings"
)
```

batch.go
eiixy: The request parameters provided by openai are inconsistent with the Pagination structure
https://platform.openai.com/docs/api-reference/batch/list

batch.go
eiixy: ok~

batch.go
eiixy: I think providing this method would reduce complexity, and I have provided an example: https://github.com/eiixy/monorepo/blob/master/example/openai/main.go

sashabaranov (approved):

PR #752, opened 2024-05-23
Support Vector Store API

kodernubie:
implement vector strore api (including vector file + file batch)

https://platform.openai.com/docs/api-reference/vector-stores

### DIFF
--- a/vector_store.go
+++ b/vector_store.go
@@ -104,11 +104,8 @@ type VectorStoreFileBatchRequest struct {
 
 // CreateVectorStore creates a new vector store.
 func (c *Client) CreateVectorStore(ctx context.Context, request VectorStoreRequest) (response VectorStore, err error) {
-	req, err := c.newRequest(ctx, http.MethodPost, c.fullURL(vectorStoresSuffix), withBody(request),
+	req, _ := c.newRequest(ctx, http.MethodPost, c.fullURL(vectorStoresSuffix), withBody(request),
 		withBetaAssistantVersion(c.config.AssistantVersion))
-	if err != nil {
-		return
-	}
 
 	err = c.sendRequest(req, &response)
 	return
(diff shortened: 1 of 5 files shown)


sashabaranov:
@kodernubie thank you for the PR!

vector_store.go
```diff
@@ -0,0 +1,349 @@
...
+	return
+}
+
+// ListVectorStores Lists the currently available vector store.
+func (c *Client) ListVectorStores(
+	ctx context.Context,
+	limit *int,
+	order *string,
+	after *string,
+	before *string,
```
sashabaranov, line 160: Could we please use [Pagination](https://github.com/sashabaranov/go-openai/blob/211cb49fc22766f4174fef15301c4d39aef609d3/run.go#L210) struct here?

vector_store.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// ListVectorStoreFiles Lists the currently available files for a vector store.
+func (c *Client) ListVectorStoreFiles(
+	ctx context.Context,
+	vectorStoreID string,
+	limit *int,
+	order *string,
+	after *string,
+	before *string,
```
sashabaranov, line 240: Could we please use the [Pagination](https://github.com/sashabaranov/go-openai/blob/211cb49fc22766f4174fef15301c4d39aef609d3/run.go#L210) struct here?

vector_store.go
```diff
@@ -0,0 +1,349 @@
...
+
+// ListVectorStoreFiles Lists the currently available files for a vector store.
+func (c *Client) ListVectorStoreFilesInBatch(
+	ctx context.Context,
+	vectorStoreID string,
+	batchID string,
+	limit *int,
+	order *string,
+	after *string,
+	before *string,
```
sashabaranov, line 321: Could we please use the [Pagination](https://github.com/sashabaranov/go-openai/blob/211cb49fc22766f4174fef15301c4d39aef609d3/run.go#L210) struct here?

vector_store_test.go
```diff
@@ -0,0 +1,339 @@
...
+					Object:        "vector_store.file_batch",
+					CreatedAt:     1234567890,
+					VectorStoreID: vectorStoreID,
+					Status:        "completed",
+					FileCounts: openai.VectorStoreFileCount{
+						InProgress: 0,
+						Completed:  1,
+						Failed:     0,
+						Cancelled:  0,
+						Total:      0,
```
sashabaranov, line 139: There's no need to specify 0 here and in the similar cases as 0 is the default value https://go.dev/tour/basics/12

vector_store.go
sashabaranov: We might as well put pagination into a separate file and provide a `NoPagination = Pagination{}` const

sashabaranov (changes requested):
Please see earlier comments

Issue #753, opened 2024-05-27
Update RunRequest to match OpenAI parameters


TheAdamSmith:
Your issue may already be reported!
Please search on the [issue tracker](https://github.com/sashabaranov/go-openai/issues) before creating one.

**Is your feature request related to a problem? Please describe.**
I would like to be able to add values to the `top_p`, `tool_choice`, and `response_format` parameters of a run request
https://platform.openai.com/docs/api-reference/runs/createRun

**Describe the solution you'd like**
Add these fields to the `RunRequest` struct. Ignore the `stream` param for now since that requires an assistant streaming implementation. The others should be safe to add.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

Disorrder:
waiting for stream feature, any plans to implement it soon?

PR #754, opened 2024-05-27
Add tool_choice, response_format, and top_p to assistant RunRequest

TheAdamSmith:
**Describe the change**
Adds the `tool_choice`, `response_format`, and `top_p` params to the `RunRequest` object to match AssistantsV2 
https://platform.openai.com/docs/assistants/whats-new
These all don't require any additional implementation. Left out the stream option since that is not yet supported. 

https://platform.openai.com/docs/api-reference/runs/createRun

**Describe your solution**
This just adds the fields to the struct so that they can be passed in the request body. These are just extra params that are ignored when passed to Assistants V1. So they should be safe to add.

**Tests**
Unit tests run successfully. 
Getting the following error when running the integration tests
```
    api_integration_test.go:86: error, status code: 404, message: The model `ada` has been deprecated, learn more here: https://platform.openai.com/docs/deprecations [
CreateCompletionStream returned error]
--- FAIL: TestAPI (2.53s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6a7cf4]
```
But this seems unrelated to my change since these tests are not invoking the assistants api. LMK if this test failure is a user error.

I did test making a request with both Assistants V1 and V2 and both worked.

Issue: #753

### DIFF
--- a/run.go
+++ b/run.go
@@ -92,6 +92,7 @@ type RunRequest struct {
 	// Sampling temperature between 0 and 2. Higher values like 0.8 are  more random.
 	// lower values are more focused and deterministic.
 	Temperature *float32 `json:"temperature,omitempty"`
+	TopP        *float32 `json:"top_p,omitempty"`
 
 	// The maximum number of prompt tokens that may be used over the course of the run.
 	// If the run exceeds the number of prompt tokens specified, the run will end with status 'complete'.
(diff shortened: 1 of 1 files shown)


sashabaranov (approved):
Thank you for the PR!

Integration tests are definitely an area for improvement for us — we should fix them and start running on every merge to the master branch

TheAdamSmith:
> Thank you for the PR!
> 
> Integration tests are definitely an area for improvement for us — we should fix them and start running on every merge to the master branch

Yeah of course! I created an issue for the integration tests and will see if I can get to it this week

Issue #761, opened 2024-06-03
Integration Tests Panic


TheAdamSmith:
**Describe the bug**
Integration tests do not run and exit with a panic
```
    api_integration_test.go:86: error, status code: 404, message: The model `ada` has been deprecated, learn more here: https://platform.openai.com/docs/deprecations [
CreateCompletionStream returned error]
--- FAIL: TestAPI (2.53s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6a7cf4]
```
based on the error it is because the models the integration tests are using are deprecated. So we should be able to just upgrade them to the closest compatible model.

**To Reproduce**
Run the integration tests as detailed in the readme

**Expected behavior**
Test should run successfully

PR #762, opened 2024-06-05
Fix integration test

TheAdamSmith:
**Describe the change**
Updates the integration tests to the updated models so that the tests will work. Updated models based off of the Openai recommended upgrades https://platform.openai.com/docs/deprecations

https://platform.openai.com/docs/deprecations

**Describe your solution**
Refactored the completion streaming test into its own class and added a `NoErrorF` check. This fixes the nil pointer dereference error when this test fails and allows the other tests to keep running.
Updated the models the tests were using 

**Tests**
image

**Additional context**
https://github.com/sashabaranov/go-openai/issues/761

Issue: #761


embeddings.go
```diff
@@ -16,7 +16,7 @@ var ErrVectorLengthMismatch = errors.New("vector length mismatch")
 type EmbeddingModel string
 
 const (
-	// Deprecated: The following block will be shut down on January 04, 2024. Use text-embedding-ada-002 instead.
+	// Deprecated: The following block is shut down. Use text-embedding-ada-002 instead.
```
TheAdamSmith, line 19: IMO these should just be removed now that they are shutdown, but wasn't sure that would cause issues

sashabaranov (approved):
Thank you so much! We should enable integration tests in the CI on master merge

PR #766, opened 2024-06-10
Support Project ID

### LINKED ISSUE #757: ClientConfig should support a ProjectID
Recently OpenAI has added the support for Projects, within an Org; and Assistants are now project-scoped.
When creating a new Client, [`ClientConfig`](https://github.com/sashabaranov/go-openai/blob/config.go#L30-L42) should support a `ProjectID` field, similar to what the [Python API](https://platform.openai.com/docs/api-reference/authentication) does:

```python
from openai import OpenAI

client = OpenAI(
  organization='org-xG5HLfr7CRdMSiSBlycUpcIw',
  project='$PROJECT_ID',
)
```

I suspect there may be a workaround currently by adding a custom header:
```
  -H "OpenAI-Project: $PROJECT_ID"
```
but this is not documented.

**Is your feature request related to a problem? Please describe.**

Assistants (and by induction, Threads and Messages) are Project-scoped in the OpenAI API, so not having this limits us to using only the `DefaultProject` which causes issues of isolation, etc.

**Describe the solution you'd like**
Add a `ProjectID` to the `ClientConfig` struct

**Describe alternatives you've considered**
Custom `OpenAI-Project` header

**Additional context**
I'd be happy to contribute the code to implement this: just say the word!

:heart: `go-openai`!!!

pkosiec:
**Describe the change**
Please provide a clear and concise description of the changes you're proposing. Explain what problem it solves or what feature it adds.

https://platform.openai.com/docs/api-reference/organizations-and-projects-optional

**Describe your solution**
I've added an optional header, same as the Organization ID.

As @massenz mentioned [here](https://github.com/sashabaranov/go-openai/issues/757#issuecomment-2134518097), while the Project ID it is optional, it would be great to match the official OpenAI client and have a double check for the project scope.

**Tests**
I've tested the changes manually.

Resolves https://github.com/sashabaranov/go-openai/issues/757

### DIFF
--- a/client.go
+++ b/client.go
@@ -191,6 +191,9 @@ func (c *Client) setCommonHeaders(req *http.Request) {
 	if c.config.OrgID != "" {
 		req.Header.Set("OpenAI-Organization", c.config.OrgID)
 	}
+	if c.config.ProjectID != "" {
+		req.Header.Set("OpenAI-Project", c.config.ProjectID)
+	}
 }
 
 func isFailureStatusCode(resp *http.Response) bool {
--- a/config.go
+++ b/config.go
@@ -32,6 +32,7 @@ type ClientConfig struct {
 
 	BaseURL              string
 	OrgID                string
+	ProjectID            string
 	APIType              APIType
 	APIVersion           string // required when APIType is APITypeAzure or APITypeAzureAD
 	AssistantVersion     string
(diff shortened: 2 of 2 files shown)


massenz:
Thanks for the patch!
The change looks good to me ✅

PR #772, opened 2024-06-16
feat: provide vector store

Pengguancheng:
A similar PR may already be submitted!

Just want to finish this feature, thanks

implement vector strore api (including vector file + file batch)

https://platform.openai.com/docs/api-reference/vector-stores

### DIFF
--- a/vector_store.go
+++ b/vector_store.go
@@ -1,0 +1,349 @@
+package openai
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+const (
+	vectorStoresSuffix            = "/vector_stores"
+	vectorStoresFilesSuffix       = "/files"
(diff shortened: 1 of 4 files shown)


sashabaranov:
Thank you for updating https://github.com/sashabaranov/go-openai/pull/752 — looks good to me!

vector_store.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// ListVectorStores Lists the currently available vector store.
+func (c *Client) ListVectorStores(
+	ctx context.Context,
+	pagination Pagination,
+) (response VectorStoresList, err error) {
+	urlValues := url.Values{}
+
+	if pagination.After != nil {
```
sashabaranov, line 166: If you'll have time — it would be great to have a reusable func for this logic, there are 4 cases where we use it in this repo!

sashabaranov (approved):

PR #787, opened 2024-07-10
feat: ParallelToolCalls to ChatCompletionRequest with helper functions

LGXerxes:
**Describe the change**
Openai has by default a `parallel_tool_calls` set to true, which allows the llm to make multiple tool calls in one response.

This can be opted out of with setting it to false, and ensures the llm only responds with one tool call at a time.

https://platform.openai.com/docs/guides/function-calling/parallel-function-calling

**Describe your solution**
We only need to set the field to false, and true for being certain it is true, we can not just use a `bool` type.
As that would require a custom marshaller for not including it when not set.

I think a `*bool` works oke here, it is a bit annoying to set a `*bool` value. Which is why the helper functions have been given: `ParallelOptionFalse()` and `ParallelOptionTrue()`.

Intended use:
```go
request := openai.ChatCompletionRequest{}
request.Model = completeChatModel
	
request.ParallelToolCalls = openai.ParallelOptionFalse()	
```

This seemed the best option without resorting to custom types with custom marshallers, or creating a custom marshal for `ChatCompletionRequest`

**Tests**
Currently using a version with `bool` type and no `omitempty` which has stopped gpt responding with multiple function calls.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -218,6 +218,18 @@ type ChatCompletionRequest struct {
 	ToolChoice any `json:"tool_choice,omitempty"`
 	// Options for streaming response. Only set this when you set stream: true.
 	StreamOptions *StreamOptions `json:"stream_options,omitempty"`
+	// Disable the default behavior of parallel tool calls.
+	ParallelToolCalls *bool `json:"parallel_tool_calls,omitempty"`
+}
+
+func ParallelOptionFalse() *bool {
+	b := false
+	return &b
+}
+
+func ParallelOptionTrue() *bool {
+	b := true
+	return &b
 }
 
 type StreamOptions struct {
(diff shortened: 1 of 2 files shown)


sashabaranov:
Hey @LGXerxes thank you for the PR!

chat.go
```diff
@@ -218,6 +218,18 @@ type ChatCompletionRequest struct {
 	ToolChoice any `json:"tool_choice,omitempty"`
 	// Options for streaming response. Only set this when you set stream: true.
 	StreamOptions *StreamOptions `json:"stream_options,omitempty"`
+	// Disable the default behavior of parallel tool calls.
+	ParallelToolCalls *bool `json:"parallel_tool_calls,omitempty"`
```
sashabaranov, line 222: I guess we can make this an `any`, comment that the boolean is expected and don't bother with `ParallelOptionTrue`

LGXerxes: Yes! makes much more sense.

have modified it to be any

sashabaranov (approved):

Issue #735, opened 2024-05-06
Support Thread and Message for Assistant v2


ZhihanWei:
**Is your feature request related to a problem? Please describe.**
OpenAI has released their new version of assistant([v2](https://platform.openai.com/docs/assistants/whats-new)) and the old version of assistant(v1) will be deprecated by the end of 2024. There are some incompatibilities between the two version in terms of API request body. 
For example, `MessageRequest` in v1
```
type MessageRequest struct {
	Role     string         `json:"role"`
	Content  string         `json:"content"`
	FileIds  []string       `json:"file_ids,omitempty"` //nolint:revive // backwards-compatibility
	Metadata map[string]any `json:"metadata,omitempty"`
}
```
`MessageRequest` in v2, which introduce `attachments`
```
type MessageRequest struct {
	Role        string         `json:"role"`
	Content     string         `json:"content"`
	Attachments []Attachment   `json:"attachments,omitempty"`
	Metadata    map[string]any `json:"metadata,omitempty"`
}
```

We let the user configure the assistant version in the client configuration but the new assistant version will cause trouble at runtime when making calls to Thread and Message.

**Describe the solution you'd like**
Support both assistant v1 and v2. 

**Additional context**
There are different implementations I can think of that could make the enhancement, but most likely the compatibility will be checked at runtime. Would be good to see an implementation to support both assistant versions and check errors at compile time while maintaining the configured `assistantVersion`.

qhenkart:
All of the structs have changed. The run steps have also changed format. The existing structs do not work with V2

Trojan295:
There are more changes, because the `Content` can be now a string, but also an array of objects. https://platform.openai.com/docs/api-reference/messages/createMessage

ZhihanWei:
In the new version, seems like `assistant v1` cannot use`gpt-4o`.

nadeemc:
A workaround for now is to use a custom config to set the version:

```go
config := openai.DefaultConfig(key)
config.AssistantVersion = "v2"

return openai.NewClientWithConfig(config)
```

dezchai:
Was working on a pull request for this but don't know how to handle content being either a string or array in Go.

dezchai:
@sashabaranov thoughts?

sashabaranov:
https://github.com/sashabaranov/go-openai/releases/tag/v1.26.0 contains assistant API v2 by default

dezchai:
[...]

PR #792, opened 2024-07-15
fix: Updated ThreadMessage struct with latest fields based on OpenAI docs

ecoralic:
**Describe the change**
ThreadMessage has deprecated fields, and new fields are missing, so requests with files cannot be made.

https://platform.openai.com/docs/api-reference/threads/createThread

**Describe your solution**
Updated ThreadMessage based on OpenAI docs.

### DIFF
--- a/thread.go
+++ b/thread.go
@@ -83,14 +83,24 @@ type ModifyThreadRequest struct {
 type ThreadMessageRole string
 
 const (
-	ThreadMessageRoleUser ThreadMessageRole = "user"
+	ThreadMessageRoleAssistant ThreadMessageRole = "assistant"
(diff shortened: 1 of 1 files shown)


MontiMarco92:
Hi @ecoralic. Thanks for this. I think that `FileIDs` should still remain for backwards compatibility with V1, what do you think? Besides, now `Content` can be a string or an array of objects. Check this related [issue](https://github.com/sashabaranov/go-openai/issues/735)

sashabaranov (approved):
Thank you!

beastmaximus:
What about `MessageRequest` struct? Should it also contain `Attachments` field just like `ThreadMessage` does?

PR #793, opened 2024-07-16
add finetuning hyperparams

janukobytsch:
Adds `learning_rate_multiplier` and `batch_size` parameters.

Supported by both OpenAI and Azure, see:
* https://platform.openai.com/docs/api-reference/fine-tuning/create
* https://learn.microsoft.com/en-us/rest/api/azureopenai/fine-tuning/create?view=rest-azureopenai-2024-03-01-preview&tabs=HTTP#finetuninghyperparameters

### DIFF
--- a/fine_tuning_job.go
+++ b/fine_tuning_job.go
@@ -26,7 +26,9 @@ type FineTuningJob struct {
 }
 
 type Hyperparameters struct {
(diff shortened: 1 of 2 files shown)


sashabaranov (approved):
Thank you!

Issue #794, opened 2024-07-18
Azure openai assistans list interface parameter splicing error


WhiteNightMo:
**Environment:**
 - go-openai version: v1.24.0
 - Go version: 1.19.10

**Screenshots/Logs**
![image](https://github.com/user-attachments/assets/39c6faee-8e9d-4e28-9531-adab20fda420)

**Additional context**
If the parameter contains paging information, the generated url will contain two question marks.
The output I printed: 
```
c.fullURL(urlSuffix): https://xxx.openai.azure.com/openai/assistants?limit=10?api-version=2023-05-15
```

WhiteNightMo:
But the `baseURL` comes from the config file, and it's really the `suffix` that should be of concern.
```golang
baseURL := c.config.BaseURL
baseURL = strings.TrimRight(baseURL, "/")
parseURL, _ := url.Parse(baseURL)
```

Here's what I printed and it didn't change anything.
![image](https://github.com/user-attachments/assets/d0aa0baf-d579-486b-812e-609b39ca02f6)

WhiteNightMo:
@eiixy @sashabaranov

eiixy:
@WhiteNightMo Could you please check if the unit tests cover your use case? #817

WhiteNightMo:
> @WhiteNightMo Could you please check if the unit tests cover your use case? #817

Thanks for the fix. For me, the [`/audio/speech`](https://platform.openai.com/docs/api-reference/audio/createSpeech) interface was missing from the judgment here. I'm not sure if there are other interfaces missing here.
```golang
if !containsSubstr([]string{
	"/completions",
	"/embeddings",
	"/chat/completions",
	"/audio/transcriptions",
	"/audio/translations",
	"/images/generations",
}, parseSuffix.Path) {
	return fmt.Sprintf("%s/%s%s?%s", baseURL, azureAPIPrefix, parseSuffix.Path, query.Encode())
}
```

eiixy:
> > @WhiteNightMo Could you please check if the unit tests cover your use case? #817
> 
> Thanks for the fix. For me, the [`/audio/speech`](https://platform.openai.com/docs/api-reference/audio/createSpeech) interface was missing from the judgment here. I'm not sure if there are other interfaces missing here.
> 
> ```go
> if !containsSubstr([]string{
> 	"/completions",
> 	"/embeddings",
> 	"/chat/completions",
> 	"/audio/transcriptions",
> 	"/audio/translations",
[...]

PR #797, opened 2024-07-19
fix: Azure openai parameter splicing error bug 

eiixy:
fix: #794

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"strings"
 
 	utils "github.com/sashabaranov/go-openai/internal"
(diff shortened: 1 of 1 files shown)


sashabaranov (approved):
❤️ thank you!

client.go
```diff
@@ -228,10 +229,13 @@ func (c *Client) fullURL(suffix string, args ...any) string {
 	if c.config.APIType == APITypeAzure || c.config.APIType == APITypeAzureAD {
 		baseURL := c.config.BaseURL
 		baseURL = strings.TrimRight(baseURL, "/")
+		parseURL, _ := url.Parse(baseURL)
```
sashabaranov, line 232: @eiixy btw, if you'd have any interest – it would be great to refactor this method https://github.com/sashabaranov/go-openai/issues/636

Issue #788, opened 2024-07-10
Completion request should support seed


AjinkyaZ:
ChatCompletionRequest supports a seed value as implemented in #525
https://github.com/sashabaranov/go-openai/blob/03851d20327b7df5358ff9fb0ac96f476be1875a/chat.go#L186-L221

However, the CompletionRequest does not:
https://github.com/sashabaranov/go-openai/blob/03851d20327b7df5358ff9fb0ac96f476be1875a/completion.go#L134-L154

while OpenAI docs do indicate support for it:
OpenAI docs for `/completions` https://platform.openai.com/docs/api-reference/completions/create#completions-create-seed

PR #800, opened 2024-07-19
feat: Completion request support seed

eiixy:
feat: #788


sashabaranov (changes requested):
@eiixy Thank you for the PR! Could we please keep the order of fields?

eiixy:
> 感谢您的公关！我们能否请保持字段的顺序？

ok~

> @eiixy Thank you for the PR! Could we please keep the order of fields?

I think it would be better to keep the field order consistent with the OpenAI documentation to compare the differences.

sashabaranov (approved):

Issue #771, opened 2024-06-13
Integration tests for Assistants API


sashabaranov:
After the recent PRs #762 and #769 we started to run integration tests on every merge to `master` branch.

At the same time, integration tests do not cover many API calls the library supports: https://github.com/sashabaranov/go-openai/blob/master/api_integration_test.go

Notably, integration tests do not call popular Assistants API.

PR #808, opened 2024-07-24
Add batch api example

eiixy:


sashabaranov:
Thank you for the PR! Could you please make it an integration test as well? The library would really benefit from more integration tests (see #771)

For more details: https://github.com/sashabaranov/go-openai/blob/master/api_integration_test.go

eiixy:
I noticed that when people use the `CreateBatchWithUploadFile` method, they often overlook the `AddChatCompletion`, `AddCompletion`, and `AddEmbedding` helper methods in the `CreateBatchWithUploadFileRequest`. Therefore, I added the `CreateBatchWithChatCompletions` and `CreateBatchWithEmbeddings` methods to clarify their usage.

PR #813, opened 2024-08-07
feat: Support Structured Outputs

eiixy:
Support Structured Outputs 
https://platform.openai.com/docs/guides/structured-outputs/introduction

### DIFF
--- a/api_integration_test.go
+++ b/api_integration_test.go
@@ -4,6 +4,7 @@ package openai_test
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"io"
 	"os"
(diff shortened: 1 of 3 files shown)


sashabaranov:
Thank you for the PR!

Could we please add some tests here? Both integration and unit tests. I believe we also need to work on `ChatCompletionResponse` for this

h0rv:
Would love to add this to https://github.com/instructor-ai/instructor-go to streamline parsing the responses into Go structs.

eiixy:
> Thank you for the PR!
> 
> Could we please add some tests here? Both integration and unit tests. I believe we also need to update the `ChatCompletionResponse` to make this work

Thanks for the feedback! I've added the requested unit tests. Please review and let me know if anything else is needed.

api_integration_test.go
```diff
@@ -178,3 +179,63 @@ func TestAPIError(t *testing.T) {
...
+						AdditionalProperties: false,
+					},
+					Strict: true,
+				},
+			},
+		},
+	)
+	checks.NoError(t, err, "CreateChatCompletion (use json_schema response) returned error")
+	var result = make(map[string]string)
+	err = json.Unmarshal([]byte(resp.Choices[0].Message.Content), &result)
```
sashabaranov, line 234: It would be super cool if we could do that automatically based on JSON schema in the future!

sashabaranov (approved):
Thank you!!

api_integration_test.go
sashabaranov: I believe it is not possible to construct a Go type based on json schema description in the current form.

What we could do is to use struct tags like https://github.com/invopop/jsonschema is doing, and than be able to both generate json schema from Go struct and also be able to automatically unmarshal structured responses

sashabaranov: Added this to our v2 roadmap as might require breaking changes https://github.com/sashabaranov/go-openai/discussions/801

eiixy: Implement it like this?
```
package jsonschema

func Unmarshal(schema Definition, data []byte,v any) error {
	// TODO
}
```

sashabaranov: @eiixy Ye, pretty much! The only problem is that you need a definition and Go type. In this test the definition is

```go
openai.ChatCompletionResponseFormatJSONSchema{
	Name: "cases",
	Schema: jsonschema.Definition{
		Type: jsonschema.Object,
		Properties: map[string]jsonschema.Definition{
			"PascalCase": jsonschema.Definition{Type: jsonschema.String},
			"CamelCase":  jsonschema.Definition{Type: jsonschema.String},
			"KebabCase":  jsonschema.Definition{Type: jsonschema.String},
			"SnakeCase":  jsonschema.Definition{Type: jsonschema.String},
		},
		Required:             []string{"PascalCase", "CamelCase", "KebabCase", "SnakeCase"},
		AdditionalProperties: false,
	},
	Strict: true,
},
```

and Go type is `map[string]string`.

Ideally, we would like to
```go
type MyStructuredResponse struct {
	PascalCase struct
	CamelCase struct
	KebabCase struct
	SnakeCase struct
}
```

So that could be automatically used as a schema definition (or converted to it) and unmarshalling of structured response. Potentially we can have a function that creates `jsonschema.Definition` from a given struct

gspeicher: Thanks for putting this together so quickly. I already have a plaintext JSON schema that I have been sending to the API in the system prompt, and already have a Go struct to hold the unmarshalled response. I was hoping to reuse both with the new Structured Outputs feature without having to manually convert the schema to a `jsonschema.Definition` but I don't see any way to do that.

Could you possibly change the accepted schema type from `jsonschema.Definition` to `json.Marshaler` from the standard Go `encoding/json` package? The `jsonschema.Definition` godoc itself states "It is fairly limited, and you may have better luck using a third-party library." but by defining `Schema` as a `jsonschema.Definition` we are precluded from using any third-party library to construct the schema. This would be a backward compatible change since `jsonschema.Definition` already implements `MarshalJSON`.

Thanks for your consideration.

eiixy: @sashabaranov Could you review the feasibility of this implementation plan? #819

PR #819, opened 2024-08-07
Improve handling of JSON Schema in OpenAI API Response Context

eiixy:
* add: jsonschema.Unmarshal
* add: jsonschema.Validate
* add: jsonschema.GenerateSchemaForType
* update: openai.ChatCompletionResponseFormatJSONSchema


jsonschema/json.go
```diff
@@ -53,3 +58,90 @@ func (d Definition) MarshalJSON() ([]byte, error) {
...
+			if s := field.Tag.Get("required"); s != "" {
+				required, _ = strconv.ParseBool(s)
+			}
+			if required {
+				requiredFields = append(requiredFields, jsonTag)
+			}
+		}
+		d.Required = requiredFields
+		d.Properties = properties
+	default:
```
sashabaranov, line 144: I suggest we either return an error or panic here

api_integration_test.go
```diff
@@ -190,6 +189,13 @@ func TestChatCompletionResponseFormat_JSONSchema(t *testing.T) {
 	c := openai.NewClient(apiToken)
 	ctx := context.Background()
 
+	type MyStructuredResponse struct {
+		PascalCase string `json:"pascal_case" required:"true" description:"PascalCase"`
+		CamelCase  string `json:"camel_case" required:"true" description:"CamelCase"`
+		KebabCase  string `json:"kebab_case" required:"true" description:"KebabCase"`
+		SnakeCase  string `json:"snake_case" required:"true" description:"SnakeCase"`
+	}
+	schema := jsonschema.Warp(MyStructuredResponse{})
```
sashabaranov, line 198: Great stuff, I guess we should put a proper README example on how to use this. That should be one of the top examples for sure.

jsonschema/json.go
```diff
@@ -53,3 +57,84 @@ func (d Definition) MarshalJSON() ([]byte, error) {
...
+	}
+	return &v, nil
+}
+
+func (r SchemaWrapper[T]) String() string {
+	bytes, _ := json.MarshalIndent(r.schema, "", "  ")
+	return string(bytes)
+}
+
+func Warp[T any](v T) SchemaWrapper[T] {
```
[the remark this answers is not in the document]
sashabaranov: Or even better: `GenerateSchemaForType`

jsonschema/json.go
```diff
@@ -53,3 +58,90 @@ func (d Definition) MarshalJSON() ([]byte, error) {
 		Alias: (Alias)(d),
 	})
 }
+
+type SchemaWrapper[T any] struct {
```
sashabaranov, line 62: Pardon my ignorance: do we even need this type?

jsonschema/json.go
eiixy: Thank you for the suggestion!

gspeicher:
If this PR is not going to be finished soon, would it be possible to lift the small change to `openai.ChatCompletionResponseFormatJSONSchema` promoting `Schema` from `jsonschema.Definition` to `json.Marshaler` as a separate PR? The requirement to use `jsonschema` is currently a roadblock for myself and others who use this package, as evidenced by PR #824.

Also, I feel that using `json.Marshaler` is both cleaner and more in the spirit of Go than adding a new `SchemaRaw` field side-by-side with `Schema` as proposed by that PR. If you already have a schema in string form then you can simply wrap it in a custom type that satisfies `json.Marshaler` like so:

```
// MySchema is a custom type that wraps a string
type MySchema string

// MarshalJSON implements the json.Marshaler interface for MySchema
func (m MySchema) MarshalJSON() ([]byte, error) {
	return []byte(m), nil
}
```

api_integration_test.go
eiixy: I have added an example use case from OpenAI documentation to the README. For reference, please see: https://platform.openai.com/docs/guides/structured-outputs/examples

gspeicher:
@eiixy thanks for pulling this together quickly. My feedback is centered around `GenerateSchemaForType`.

1. The name feels a little misleading to me, because it implies that you would pass the type corresponding to the schema you want to generate, but instead it expects an object of that type. Seems like either the function name or the parameter type should change so they agree. I defer to @sashabaranov since the name was his suggestion and it's his package.
2. Do we still need `SchemaWrapper` and its `Wrap` method if we have `GenerateSchemaForType`? It's not clear to me when you would use `SchemaWrapper` and it feels like an unnecessarily abstraction of the native JSON package.

eiixy:
> @eiixy thanks for pulling this together quickly. My feedback is centered around `GenerateSchemaForType`.
> 
> 1. The name feels a little misleading to me, because it implies that you would pass the type corresponding to the schema you want to generate, but instead it expects an object of that type. Seems like either the function name or the parameter type should change so they agree. I defer to @sashabaranov since the name was his suggestion and it's his package.
> 2. Do we still need `SchemaWrapper` and its `Wrap` method if we have `GenerateSchemaForType`? It's not clear to me when you would use `SchemaWrapper` and it feels like an unnecessarily abstraction of the native JSON package.


`jsonschema.Wrap` is indeed not necessary; it's just an alternative way of doing it. You can refer to the example below.
```go
type Result struct {
	Steps []struct {
		Explanation string `json:"explanation"`
		Output      string `json:"output"`
	} `json:"steps"`
	FinalAnswer string `json:"final_answer"`
}
sw, err := jsonschema.Wrap(Result{})
if err != nil {
	log.Fatalf("JSONSchema Wrap error: %v", err)
}
resp, err := client.CreateChatCompletion(ctx, openai.ChatCompletionRequest{
	Model: openai.GPT4oMini,
		{
			Role:    openai.ChatMessageRoleSystem,
			Content: "You are a helpful math tutor. Guide the user through the solution step by step.",
		},
		{
			Content: "how can I solve 8x + 7 = -23",
		},
	},
	ResponseFormat: &openai.ChatCompletionResponseFormat{
		Type: openai.ChatCompletionResponseFormatTypeJSONSchema,
		JSONSchema: &openai.ChatCompletionResponseFormatJSONSchema{
			Name:   "math_reasoning",
			Schema: sw,
			Strict: true,
		},
	},
})
if err != nil {
	log.Fatalf("CreateChatCompletion error: %v", err)
}
result, err := sw.Unmarshal(resp.Choices[0].Message.Content)
if err != nil {
	log.Fatalf("Unmarshal error: %v", err)
}
```

gspeicher:
Taking your example and converting it to use `GenerateSchemaForType`, it seems like the only difference would be these two lines:
```
sw, err := jsonschema.Wrap(Result{})
// check for errors and call the API
result, err := sw.Unmarshal(resp.Choices[0].Message.Content)
```
would become:
```
var result Result
sw, err := jsonschema.GenerateSchemaForType(result)
// check for errors and call the API
err := json.Unmarshal([]byte(resp.Choices[0].Message.Content), &result)
```
Again, this is not my library so I defer to @sashabaranov but I personally don't see the point in including that alternative since the standard `encoding/json` library is already simple and ubiquitous, not to mention the complexity and cognitive load are effectively the same between the two options. Overall, adding this as an alternative actually adds unnecessary complexity to the package.

Having said that, I'm chomping at the bit to see this PR get merged and start using Structured Output so I'm not going to die on this hill.

eiixy:
> Taking your example and converting it to use `GenerateSchemaForType`, it seems like the only difference would be these two lines:
> 
> ```
> sw, err := jsonschema.Wrap(Result{})
> // check for errors and call the API
> result, err := sw.Unmarshal(resp.Choices[0].Message.Content)
> ```
> 
> would become:
> 
> ```
> var result Result
> sw, err := jsonschema.GenerateSchemaForType(result)
> // check for errors and call the API
> err := json.Unmarshal([]byte(resp.Choices[0].Message.Content), &result)
> ```
> 
> Again, this is not my library so I defer to @sashabaranov but I personally don't see the point in including that alternative since the standard `encoding/json` library is already simple and ubiquitous, not to mention the complexity and cognitive load are effectively the same between the two options. Overall, adding this as an alternative actually adds unnecessary complexity to the package.
> 
> Having said that, I'm chomping at the bit to see this PR get merged and start using Structured Output so I'm not going to die on this hill.

I've removed the `jsonschema.SchemaWrapper`. Thanks for your suggestion!

jsonschema/json.go
```diff
@@ -53,3 +59,96 @@ func (d Definition) MarshalJSON() ([]byte, error) {
 		Alias: (Alias)(d),
 	})
 }
+
+func (d Definition) Unmarshal(content string, v any) error {
```
sashabaranov, line 63:
```suggestion
func (d *Definition) Unmarshal(content string, v any) error {
```

jsonschema/validate.go
```diff
@@ -0,0 +1,89 @@
+package jsonschema
+
+import (
+	"encoding/json"
+	"errors"
+)
+
+func Unmarshal(schema Definition, content []byte, v any) error {
```
sashabaranov, line 8:
```suggestion
func VerifySchemaAndUnmarshal(schema Definition, content []byte, v any) error {
```

It would be great if we could separate all validation stuff into a separate PR — I'm struggling to review this due to the large size.

api_integration_test.go
```diff
@@ -213,29 +223,16 @@ func TestChatCompletionResponseFormat_JSONSchema(t *testing.T) {
...
-						Properties: map[string]jsonschema.Definition{
-							"PascalCase": jsonschema.Definition{Type: jsonschema.String},
-							"CamelCase":  jsonschema.Definition{Type: jsonschema.String},
-							"KebabCase":  jsonschema.Definition{Type: jsonschema.String},
-							"SnakeCase":  jsonschema.Definition{Type: jsonschema.String},
-						},
-						Required:             []string{"PascalCase", "CamelCase", "KebabCase", "SnakeCase"},
-						AdditionalProperties: false,
-					},
+					//Schema: sw,
```
sashabaranov, line 226: let's remove this line

jsonschema/validate_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"testing"
+
+	"github.com/sashabaranov/go-openai/jsonschema"
+)
+
+func Test_Validate(t *testing.T) {
+	type args struct {
+		data   interface{}
```
sashabaranov, line 11:
```suggestion
		data   any
```

jsonschema/validate.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return err
+	}
+	if !Validate(schema, data) {
+		return errors.New("data validation failed against the provided schema")
+	}
+	return json.Unmarshal(content, &v)
+}
+
+func Validate(schema Definition, data interface{}) bool {
```
sashabaranov, line 20:
```suggestion
func Validate(schema Definition, data any) bool {
```

jsonschema/validate.go
```diff
@@ -0,0 +1,89 @@
...
+			return false
+		} else if !exists && contains(schema.Required, key) {
+			return false
+		}
+	}
+	return true
+}
+
+func validateArray(schema Definition, data any) bool {
+	dataArray, ok := data.([]interface{})
```
sashabaranov, line 70:
```suggestion
	dataArray, ok := data.([]any)
```

jsonschema/validate.go
```diff
@@ -0,0 +1,89 @@
...
+	switch schema.Type {
+	case Object:
+		return validateObject(schema, data)
+	case Array:
+		return validateArray(schema, data)
+	case String:
+		_, ok := data.(string)
+		return ok
+	case Number: // float64 and int
+		_, ok := data.(float64)
```
sashabaranov, line 30: should we also support float32 here?

sashabaranov (approved):
Great work, thank you!!

AyushSawant18588:
I’m building a service-to-service integration that needs to bind incoming chat-completion requests directly into the Go client’s ChatCompletionRequest struct. But when doing this get the following error:
```
json: cannot unmarshal object into Go struct field ChatCompletionResponseFormatJSONSchema.response_format.json_schema.schema of type json.Marshaler"
```
This is happening cause Schema is of type json.Marshaler which does not implement an unmarshal function.
I’d like to know: is there a valid JSON request body I can send today that will bind cleanly against the existing struct definition? thanks!

Issue #814, opened 2024-08-07
Support for Structured Outputs


hari4698:
I was wondering if there will support in the future for [Structured Outputs](https://platform.openai.com/docs/guides/structured-outputs). Open AI released this feature on August 6th.

### Python example from the Open API docs: 

```python
from pydantic import BaseModel
from openai import OpenAI

client = OpenAI()

class CalendarEvent(BaseModel):
    name: str
    date: str
    participants: list[str]

completion = client.beta.chat.completions.parse(
    model="gpt-4o-2024-08-06",
    messages=[
        {"role": "system", "content": "Extract the event information."},
        {"role": "user", "content": "Alice and Bob are going to a science fair on Friday."},
    ],
    response_format=CalendarEvent,
)

event = completion.choices[0].message.parsed
```

h0rv:
I am looking to add support for the `strict` json schema too.

[instructor-go](https://github.com/instructor-ai/instructor-go) currently gives a similar experience as the python API provides, and in place for Pydantic it use Go struct tags and jsonschema.

modeledge:
Great news. 

I'm not sure how you will approach implementing the actual json schema definition within the request but consider allowing for different implementations of the actual json schema creation. I started with field tags, as in [instructor-go](https://github.com/instructor-ai/instructor-go), but have since moved away from that approach and think field tags are too limiting.

h0rv:
For instructor, JSON tags has been sufficient.

I have been struggling to convert the json schema to the currently defined struct here.

I think it would be great to optionally pass a raw schema to `response_format`.

For example, the current struct does not support defs, which are crucial, especially with complex data types. The OpenAI API supports this: https://platform.openai.com/docs/guides/structured-outputs/supported-schemas.

monstercameron:
This enhancement would be great!

rjcorwin:
Looks like this issue is resolved? https://github.com/sashabaranov/go-openai/pull/813

h0rv:
My issue above is resolved: for those wanting to use their own schema string, and not use the `go-openai/jsonschema`, you can do the following:

```go
JSONSchema: &openai.ChatCompletionResponseFormatJSONSchema{
    Name:        "my-name",
    Schema:      json.RawMessage(schemaJSON),
    Strict:      true,
}
```

hari4698:
Support added for Structured Output. #813

PR #824, opened 2024-08-10
add `SchemaRaw` option for strict json response format

h0rv:
**Describe the change**

Added the ability to pass a raw JSON Schema (`[]byte`) to ResponseFormat for strict mode (structured output).

This is done to be more dynamic with the JSON schema, as there is many features that are there and may be added to the API that would be out of scope for this client to support. For example: https://platform.openai.com/docs/guides/structured-outputs/definitions-are-supported.

These JSON spec features are provided by other Go libraries, and would not make sense to add to this client (unless wanting to add a json schema package it as a dependency).

https://platform.openai.com/docs/guides/structured-outputs/definitions-are-supported

**Describe your solution**

I added a solution to pass in a raw JSON schema and overrode the marshalling method to be compatibile with the existing way and support this way while blocking it if the struct and raw schema are both defined.

**Tests**

I added an integration test to pass the raw JSON schema throught the new field.

Issue: https://github.com/sashabaranov/go-openai/issues/814

### DIFF
--- a/api_integration_test.go
+++ b/api_integration_test.go
@@ -239,3 +239,65 @@ func TestChatCompletionResponseFormat_JSONSchema(t *testing.T) {
 		}
 	}
 }
+
+func TestChatCompletionResponseFormat_JSONSchemaRaw(t *testing.T) {
+	apiToken := os.Getenv("OPENAI_TOKEN")
+	if apiToken == "" {
+		t.Skip("Skipping testing against production OpenAI API. Set OPENAI_TOKEN environment variable to enable it.")
+	}
+
+	var err error
+	c := openai.NewClient(apiToken)
+	ctx := context.Background()
+
+	schema := []byte(`{
+		"type": "object",
+		"properties": {
+			"CamelCase": {"type": "string"},
+			"KebabCase": {"type": "string"},
+			"PascalCase": {"type": "string"},
+			"SnakeCase": {"type": "string"}
+		},
+		"required": ["PascalCase", "CamelCase", "KebabCase", "SnakeCase"],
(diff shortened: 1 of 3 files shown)


monstercameron:
Good on you for submitting this change, this is exactly what I need to proceed with my POC.

AlexanderHott:
+1

sashabaranov:
Hey all, we've just merged #819 with Schema improvements — please let me know if it works for you!

h0rv:
@sashabaranov I am trying to test with `1.29.0` but am seeing this error:

```
Error: unsupported type: map
```

Any reason maps aren't supported? I am using a struct to JSON Schema library which has some mapped fields like this and its not in the actual struct.

sashabaranov:
@h0rv Good point, we are totally open for a PR with map support! :D cc @eiixy

h0rv:
Thanks the replies - I will try using the the string schema directly.

gspeicher:
> You can directly use `json.RawMessage(mySchema)`

Oh nice, I was not aware of that gem. This is much better than implementing your own as I suggested above. Thanks for calling this out! Perhaps it should go in the documentation...

h0rv:
Awesome - was able to get it working and add support for strict mode with instructor. Just added the feature in the latest release. 

Example: https://github.com/instructor-ai/instructor-go/blob/b63ca60f159b7c7cb9517aec12082b27a6cec0e5/examples/auto_ticketer/main.go#L106-L129

h0rv:
I'm going to close this as its not required with the last PR. Thanks for the help everyone!

PR #827, opened 2024-08-15
add strict parameter to function definition

jsingle:
**Describe the change**
This supports the "function calling" aspect of https://openai.com/index/introducing-structured-outputs-in-the-api/

Other PRs address the response_format part, but I haven't seen one for this yet?

https://openai.com/index/introducing-structured-outputs-in-the-api/
https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools

**Describe your solution**
This seems like a dead simple change, unless I'm missing something?

https://github.com/sashabaranov/go-openai/issues/814

### DIFF
--- a/api_integration_test.go
+++ b/api_integration_test.go
@@ -78,6 +78,20 @@ func TestAPI(t *testing.T) {
 	)
 	checks.NoError(t, err, "CreateChatCompletion (with name) returned error")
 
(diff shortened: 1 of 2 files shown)


jsingle:
`strict` is only supported in `tools`. So the required change is a bit larger

PR #838, opened 2024-09-01
Dynamic model

mathisen99:
- New Dynamic model that always points to the latest GPT4o model.

OpenAi documentation for this can be found here: https://platform.openai.com/docs/models/gpt-4o


sashabaranov (approved):
Thank you!

Issue #839, opened 2024-09-01
fix jsonschema test


shadowpigy:
The jsonschema tests are outdated. 

The main difference is that the `properties` field is now `omitempty`. 
Another problem is that the last test case with regard to array is incorrect. The following is not a valid json schema. The `properties` keyword should be nested within an object type, not at the root level of an array type.
```
{
   "type":"array",
   "items":{
      "type":"string",
      "properties":{
         
      }
   },
   "properties":{
      "name":{
         "type":"string",
         "properties":{}
      }
   }
}
```

PR #840, opened 2024-09-01
ordered jsonschema def

shadowpigy:
The old jsonschema package does not ensure that the order of fields is the same as go struct def , which is bad for taking advantage of chain of thought.

also fix Issue: #839


shadowpigy:
A string type Definition still has the `properties` field

PR #843, opened 2024-09-04
depricated model GPT3Ada changed to GPT3Babbage002

Arundas666:
**Describe the change**
Model GPT3Ada is deprecated, so changed to GPT3Babbage002 in examples,tests and in Readme.

**Describe your solution**
deprecated model is changed and used the GPT3Babbage002 model


sashabaranov:
Thank you for the PR!

sashabaranov (approved):

PR #847, opened 2024-09-09
feat: add ParallelToolCalls to RunRequest

kenshin54:
**Describe the change**
By default, `parallel_tool_calls` is set to true, which allows the llm to make multiple tool calls in one response. This can be opted out of with setting it to false, and ensures the llm only responds with one tool call at a time.

https://platform.openai.com/docs/api-reference/runs/createRun#runs-createrun-parallel_tool_calls

**Describe your solution**
This PR added the same change as #787 , an `any`  field `ParallelToolCalls`.

**Tests**
No new tests added.


sashabaranov (approved):
Thank you for the PR!

PR #851, opened 2024-09-13
feat: add MaxCompletionTokens and CompletionTokensDetails

kunerzzz:
**Describe the change**
add MaxCompletionTokens and CompletionTokensDetails for o1 models

[Provide a relevant API doc from https://platform.openai.com/docs/api-reference
](https://platform.openai.com/docs/api-reference/chat/create)


sashabaranov:
Thank you for the PR! We've merged similar stuff in #857

Issue #837, opened 2024-08-29
Validate fails incorrectly for integer values


tylergannon:
Your issue may already be reported!
Please search on the [issue tracker](https://github.com/sashabaranov/go-openai/issues) before creating one.

**Describe the bug**
Validator fails on schema type "integer".

**To Reproduce**
Make a schema that contains an integer data type and then validate an object using that schema.

**Expected behavior**
Validation should pass

**Additional context**

Golang unmarshals all numbers as float64.  The type check for `data.(int)` therefore fails.

What will succeed will be 
```
	case jsonschema.Integer:
		if num, ok := data.(float64); ok {
			if num == float64(int64(num)) {
				return true
			}
		}
		if _, ok := data.(int); ok {
			return true // Just in case other versions of Golang support this
		}

```

I don't have the time to create a PR at this point so I've just copied your `validate.go` into my own codebase FTTB.

PR #852, opened 2024-09-17
fix: jsonschema integer validation

WeiAnAn:
**Describe the change**
Fix the jsonschema validation failed failure on the integer field.

**Describe your solution**
Follow the #837 instruction. Because go unmarshalls all numbers as float64, we should check the value is an integer.

**Tests**
Add `jsonschema.Integer` cases, include a success and a failed case.

**Additional context**

Issue: #837

### DIFF
--- a/jsonschema/validate_test.go
+++ b/jsonschema/validate_test.go
@@ -86,14 +86,6 @@ func TestUnmarshal(t *testing.T) {
 		content []byte
 		v       any
 	}
-	var result1 struct {
-		String string  `json:"string"`
-		Number float64 `json:"number"`
-	}
-	var result2 struct {
-		String string  `json:"string"`
-		Number float64 `json:"number"`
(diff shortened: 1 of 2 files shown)


sashabaranov:
Thank you for the PR! Looks good, and here's a minimal example validating the float64 argument https://go.dev/play/p/VRMIyt9Zjqz

jsonschema/validate_test.go
```diff
@@ -94,6 +94,10 @@ func TestUnmarshal(t *testing.T) {
 		String string  `json:"string"`
 		Number float64 `json:"number"`
 	}
+	var result3 struct {
```
sashabaranov, line 97: Could we maybe inline all these result1,2,3 please?

sashabaranov (approved):
thank you!!

Issue #836, opened 2024-08-28
Modify assistant notice: invalid character '<' looking for beginning of value


WangQuanTang:
Your issue may already be reported!
Please search on the [issue tracker](https://github.com/sashabaranov/go-openai/issues) before creating one.

**Describe the bug**
When I do an update to assistant it says: "error, status code: 403, message: invalid character '<' looking for beginning of value"

**To Reproduce**
```
// UpdateAssistant update assistant
func (service *RemoteAssistantService) UpdateAssistant(ctx context.Context, updateAssistantReq UpdatedAssistantReq) error {
	tools, err := formatAssistantTools(updateAssistantReq.Tools)
	if err != nil {
		return err
	}

	toolResources, err := formatAssistantToolResources(updateAssistantReq.ToolResources)
	if err != nil {
		return err
	}

	assistantRequest := openai.AssistantRequest{
		Model:         updateAssistantReq.Model,
		Instructions:  &updateAssistantReq.Instructions,
		Name:          &updateAssistantReq.Name,
		Tools:         tools,
		ToolResources: &toolResources,
	}
	if updateAssistantReq.Instructions != "" {
		assistantRequest.Instructions = &updateAssistantReq.Instructions
	}

	_, err = service.OpenAiClient.ModifyAssistant(ctx, updateAssistantReq.AssistantAppId, assistantRequest)
	if err != nil {
		return err
	}

	return nil
}
```

**Expected behavior**
I tried to google some solutions. Some said it was because of the proxy, and some said it was a situation that could only be reproduced by chance. My previous project also used `go-openai`, `version v1.26.0`, and the same update operation was successful in the previous project. By printing the context, client and request parameters, they are almost the same. Therefore, I judged that there was no problem with my code, and it might be other abnormalities.
When I upgraded to v1.29.0, the issue still exists.

**Screenshots/Logs**
```
error, status code: 403, message: invalid character '<' looking for beginning of value
```

**Environment (please complete the following information):**
 - go-openai version: go-openai v1.27.0
 - Go version: version go1.22.5
 - OpenAI API version: [e.g. v2]
 - OS: [e.g. Ubuntu 20.04]

gspeicher:
[...]

Issue #777, opened 2024-06-25
Unclear error messages for failure errors (x < 200 and x >= 400), 413 in this case


arshamalh:
**Describe the bug**
Trying to transcribe an mp3 file which happens to be larger than expected, using the snippet below:
```go
config := openai.DefaultConfig("<token>")
client := openai.NewClientWithConfig(config)
req := openai.AudioRequest{
	Model:    openai.Whisper1,
	FilePath: "music.mp3",
}
resp, err := client.CreateTranscription(context.TODO(), req)
if err != nil {
	fmt.Printf("Transcription error: %v\n", err)
	return
}
fmt.Println(resp.Text)
```
the code returns this error:
```
error, status code: 413, message: invalid character '<' looking for beginning of value
```

OK, what is it all about? Is 413 an HTTP status code? where did '<' come from?

So I started to investigate, yep, 413 is an HTTP status code, but the error comes after it `message: invalid character '<' looking for beginning of value`, has nothing about the 413, and comes from json package which expects a json response.

it's a little bit confusing.
logging the error inside handleErrorResp method (`go-openai@v1.26.0/client.go:259`):
```html
<html>
<head><title>413 Request Entity Too Large</title></head>
<body>
<center><h1>413 Request Entity Too Large</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

**Expected behavior**
I expected handleErrorResp returns a clear error like `Request Entity Too Large` before it tries to unmarshal an HTML, leading to a wrong second error.

**Environment (please complete the following information):**
 - go-openai version: v1.26.0
 - Go version: v1.21.0
 - OpenAI API version: v1
 - OS: macOS

**Additional context**
I don't exactly remember but I think I had the same issue with rate limit related errors earlier.

arshamalh:
I'm willing to open a PR if the issue sounds reasonable.

arshamalh:
Changing the AudioRequest to:
```go
req := openai.AudioRequest{
	Model:    openai.Whisper1,
	FilePath: "hashhashhash",
}
```
the file is MP3, but I didn't include the extension, the error I get is still from json package:
```
error, status code: 500, message: invalid character 'e' looking for beginning of value
```

but I expect to get something at least like the one server is returning:
```
error decoding MP3 file, make sure you send a valid MP3 file.
```

PR #856, opened 2024-09-20
Optimize Client Error Return

eiixy:
fix: #836 #806 #777


j2gg0s:
Can we add a field to RequestError to log the body when the server returns a response that is neither JSON nor an ErrorResponse?

sashabaranov (approved):
Great job, thank you!

PR #857, opened 2024-09-20
add max_completions_tokens for o1 series models

chococola:
**Describe the change**
For new `o1 series models` attribute `max_tokens` not working. In documentation this field marks deprecated.
OpenAi introduced new attribute  `max_completion_tokens`.

```
To manage costs with the o1 series models, you can limit the total number of tokens the model generates (including both reasoning and completion tokens) by using the [max_completion_tokens](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_completion_tokens) parameter.

In previous models, the max_tokens parameter controlled both the number of tokens generated and the number of tokens visible to the user, which were always equal. However, with the o1 series, the total tokens generated can exceed the number of visible tokens due to the internal reasoning tokens.

Because some applications might rely on max_tokens matching the number of tokens received from the API, the o1 series introduces max_completion_tokens to explicitly control the total number of tokens generated by the model, including both reasoning and visible completion tokens. This explicit opt-in ensures no existing applications break when using the new models. The max_tokens parameter continues to function as before for all previous models.
```

https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens
https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_completion_tokens
https://platform.openai.com/docs/guides/reasoning/controlling-costs
https://community.openai.com/t/why-was-max-tokens-changed-to-max-completion-tokens/938077

**Describe your solution**
Update `ChatCompletionRequest` model

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -371,6 +371,10 @@ func (c *Client) CreateChatCompletion(
 		return
 	}
 
+	if err = validateRequestForO1Models(request); err != nil {
+		return
+	}
+
 	req, err := c.newRequest(
 		ctx,
 		http.MethodPost,
--- a/chat.go
+++ b/chat.go
@@ -200,18 +200,25 @@ type ChatCompletionResponseFormatJSONSchema struct {
 
 // ChatCompletionRequest represents a request structure for chat completion API.
 type ChatCompletionRequest struct {
-	Model            string                        `json:"model"`
-	Messages         []ChatCompletionMessage       `json:"messages"`
-	MaxTokens        int                           `json:"max_tokens,omitempty"`
-	Temperature      float32                       `json:"temperature,omitempty"`
-	TopP             float32                       `json:"top_p,omitempty"`
-	N                int                           `json:"n,omitempty"`
-	Stream           bool                          `json:"stream,omitempty"`
-	Stop             []string                      `json:"stop,omitempty"`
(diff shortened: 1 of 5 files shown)


chococola:
@sashabaranov please take a look to the PR 🙃

sashabaranov:
@chococola great catch, thank you for the PR!

Could we also panic when o1 is used and max_tokens≠0?

sashabaranov:
Or just return an error, panic might be too much :)

chococola:
> @chococola great catch, thank you for the PR!
> 
> Could we also panic when o1 is used and max_tokens≠0?

panic would be to much, and propably break backwards capabilities for oldest models. 

Maybe we should mark attribute as deprecated ? or leave the comment?


> return an error 

in all places ? and what about batch requests ?

sashabaranov:
@chococola let's cover the most frequent chat completion + streaming case, all other cases would eventually be covered by other PRs :D

chococola:
@sashabaranov  i add validation check for `o1 series models` and add some validation for fields which are limited in beta 
https://platform.openai.com/docs/guides/reasoning/beta-limitations

sashabaranov (approved):
Wow, o1 is a can of worms! Thank you so much for looking so deeply into this! 🙌🏻

PR #860, opened 2024-09-24
Fix max_completion_tokens

alexsacr:
The json tag is incorrect, and results in an error from the API when using the o1 model.  

I didn't modify the struct field name to maintain compatibility if anyone else had started using it, but it wouldn't work for them either.


l-winston (approved):
lgtm

sashabaranov (approved):
Thank you!

PR #861, opened 2024-09-24
fix: ParallelToolCalls should be added to RunRequest

kenshin54:
Sorry about that, my previous PR #847 didn't add this field into to the correct struct, it should be added to `RunRequest`, but not `Run`.

[Provide a relevant API doc from https://platform.openai.com/docs/api-reference](https://platform.openai.com/docs/api-reference/runs/createRun#runs-createrun-parallel_tool_calls)


sashabaranov (approved):
Thank you!

PR #863, opened 2024-09-26
feat: usage struct add CompletionTokensDetails

liushuangls:
referenced document: https://platform.openai.com/docs/api-reference/chat/object#chat/object-usage


sashabaranov (approved):
Thank you!

PR #864, opened 2024-09-26
add jailbreak filter result, add ContentFilterResults on output

juliomartinsdev:
Adding field for prompt_filter_results which is in the open content filters to assess the content of the user input. Current implementation in go-openai will only return content filter information on the model output using ChatCompletionResponse. This is for AzureOpenAI spec.

Adding field for content_filter_results which is in the open content filters to assess the content of the user input. Current implementation in go-openai will only return content filter information on the model output using ChatCompletionChoice. This is for AzureOpenAI spec.

Adding field for jailbreak which is in a content filter. Current implementation in go-openai will only return content filter information on the error output using InnerError.ContentFilterResults. This is for AzureOpenAI spec.

Provide OpenAI documentation link
Relevant API doc: https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/content-filter

Describe your solution
Change struct

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -344,19 +344,21 @@ type ChatCompletionChoice struct {
 	// function_call: The model decided to call a function
 	// content_filter: Omitted content due to a flag from our content filters
 	// null: API response still in progress or incomplete
-	FinishReason FinishReason `json:"finish_reason"`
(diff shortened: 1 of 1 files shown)


sashabaranov (approved):
Thank you!

PR #875, opened 2024-10-10
Adding new moderation model constants

Mhjacobs:
**Describe the change**
Please provide a clear and concise description of the changes you're proposing. Explain what problem it solves or what feature it adds.
This change enables go-openai to use the omni moderation models that are now available to us from openai
Provide a relevant API doc from https://platform.openai.com/docs/models/moderation

**Describe your solution**
These changes define the constants for the new moderation models as specified by openAI, and then puts them into the map of allowed models for the moderation endpoint.  I have tested these strings directly against the moderation api to confirm that these requests return a non-error response and specify the same model that is passed in. 
**Tests**
Briefly describe how you have tested these changes. If possible — please add integration tests.
Unit tests, with the addition of the new models.

### DIFF
--- a/moderation.go
+++ b/moderation.go
@@ -14,8 +14,10 @@ import (
 // If you use text-moderation-stable, we will provide advanced notice before updating the model.
 // Accuracy of text-moderation-stable may be slightly lower than for text-moderation-latest.
 const (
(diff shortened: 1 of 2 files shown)


sashabaranov (approved):
Thank you!

Issue #872, opened 2024-10-08
Implement new `store` and `metadata` params


akshaylb:
**Is your feature request related to a problem? Please describe.**
OpenAI recently introduced new parameters for distillation.

**Describe the solution you'd like**
Implement `store` and `metadata` fields for the chat completion request. And add a `.WithStore(bool)` and `.WithMetadata(map[string]string)` for the OpenAI Client.

unitythemaker:
I think this is added with: https://github.com/sashabaranov/go-openai/pull/870

Not sure about ``.WithStore(bool)`` and ``.WithMetadata(map[string]string)`` though.

unitythemaker:
Hmm, it seems like it's not added for Chat Completions yet.

rmarku:
metadata is object, so it should allow things like this:
```json
 "metadata": {
        "tags": ["testTag", "taskName:pass0", "company:meitar"]
    }
```
This is in particular used by LiteLLM Proxy.

PR #870, opened 2024-10-08
Completion API: add new params

isaacseymour:
**Describe the change**
- `store` in request: this param allows you to opt a completion request in to being stored, for use in distillations and evals.
- `usage.completion_tokens_details.audio_tokens` in response: tracking audio tokens generated
- `usage.prompt_tokens_details` in response: tracking tokens used in the prompt

**Provide OpenAI documentation link**
https://platform.openai.com/docs/api-reference/chat/create#chat-create-store
https://platform.openai.com/docs/api-reference/chat/object#chat/object-usage

**Describe your solution**
Currently it's not possible to opt a completion request in to being stored using go-openai, blocking usage of distillations and evals.

It's also not possible to track usage of audio tokens, nor prompt caching.

**Tests**
These are optional new params, so I've not added specific test cases.

PR #878, opened 2024-10-15
Completion API: add Store and Metadata parameters

smackcrane:
**Describe the change**
The `store` parameter recently introduced by OpenAI has been added to `CompletionRequest` already in #870 ; it is here added to `ChatCompletionRequest`, and the related `metadata` parameter added to both.

https://platform.openai.com/docs/api-reference/chat/create#chat-create-store
https://platform.openai.com/docs/api-reference/chat/create#chat-create-metadata

**Describe your solution**
Simply adds new fields to the API call objects.

**Tests**
No new tests added.

(Partly addresses) Issue: #872


sashabaranov (approved):
Thank you!

PR #881, opened 2024-10-16
Handling for non-json response

AyushSawant18588:
**Describe the change**
Since the original error response is included in the body field of openai.RequestError, even when the response is not in JSON format, it would be more appropriate to return openai.RequestError directly, rather than returning an errorString. This approach ensures that the original error details are preserved within the body field.

**Describe your solution**
Removed the condition that checked if the response was in a non-JSON format and returned an errorString with the original error response. Now, the original error response will be directly included in the body field of openai.RequestError.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -203,14 +203,16 @@ func TestHandleErrorResp(t *testing.T) {
 			name:        "413 Request Entity Too Large",
 			httpCode:    http.StatusRequestEntityTooLarge,
 			contentType: "text/html",
-			body: bytes.NewReader([]byte(`<html>
+			body: bytes.NewReader([]byte(`
+	<html>
 	<head><title>413 Request Entity Too Large</title></head>
 	<body>
 	<center><h1>413 Request Entity Too Large</h1></center>
 	<hr><center>nginx</center>
 	</body>
 	</html>`)),
-			expected: `error, status code: 413, status: , message: invalid character '<' looking for beginning of value, body: <html>
+			expected: `error, status code: 413, status: , message: invalid character '<' looking for beginning of value, body: 
+	<html>
 	<head><title>413 Request Entity Too Large</title></head>
 	<body>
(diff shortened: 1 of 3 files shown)


AyushSawant18588:
@sashabaranov Can you please look into this?

sashabaranov:
Thank you for the PR!

client_test.go
```diff
@@ -207,13 +207,7 @@ func TestHandleErrorResp(t *testing.T) {
...
 </body>
 </html>`)),
-			expected: `error, status code: 413, status: , body: <html>
-<head><title>413 Request Entity Too Large</title></head>
-<body>
-<center><h1>413 Request Entity Too Large</h1></center>
-<hr><center>nginx</center>
-</body>
-</html>`,
+			expected: `error, status code: 413, status: , message: invalid character '<' looking for beginning of value`,
```
sashabaranov, line 210: This expected error text is a bit misleading. Code 413 indeed points out to "Content Too Large" (generic HTTP error), but the text afterward applies to json de-serialization.

I agree that this approach is better in terms of preserving the code, but in this case we actually don't preserve data

AyushSawant18588: What if Error() for RequestError (https://github.com/sashabaranov/go-openai/blob/22cf414b4be89a38eb702b13c27be1de6bcef74d/error.go#L106C1-L108C2) is updated to return body also like:
fmt.Sprintf("error, status code: %d, status: %s, message: %s, body: %s", e.HTTPStatusCode, e.HTTPStatus, e.Err, e.Body)

Should I make this change?

AyushSawant18588: @sashabaranov I have updated this PR with the above mentioned change, now in error response the original error data is also present. Can you please review it?

johnugeorge:
Thanks @sashabaranov. Can you please review this again?

sashabaranov (approved):
Great work, thank you!

PR #844, opened 2024-09-04
Add refusal

qhenkart:
**Describe the change**
With structured outputs, the OpenAI API has added the "refusal" field where user generated prompt failures show up. This is clearly defined in the documentation: https://platform.openai.com/docs/guides/structured-outputs/refusals but not present in the SDK

**Provide OpenAI documentation link**
[Provide a relevant API doc from https://platform.openai.com/docs/api-reference](https://platform.openai.com/docs/guides/structured-outputs/refusals)

**Describe your solution**
I simply added the missing field

**Tests**
n/a

**Additional context**
n/a

PR #882, opened 2024-10-17
add chatcompletion stream delta refusal and logprobs

Yu0u:
**Describe the change**
Like #844, In ChatCompletionStream.delta the refusal and logprobs field is missing

[Provide a relevant API doc from https://platform.openai.com/docs/api-reference](https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-choices)

**Describe your solution**
Add Refusal and Logprobs Fields

**Tests**
n/a

**Additional context**
n/a

### DIFF
--- a/chat_stream.go
+++ b/chat_stream.go
@@ -32,11 +32,11 @@ type ChatCompletionTokenLogprobTopLogprob struct {
 }
 
 type ChatCompletionStreamChoice struct {
-	Index                int                                `json:"index"`
-	Delta                ChatCompletionStreamChoiceDelta    `json:"delta"`
-	Logprobs             ChatCompletionStreamChoiceLogprobs `json:"logprobs,omitempty"`
-	FinishReason         FinishReason                       `json:"finish_reason"`
-	ContentFilterResults ContentFilterResults               `json:"content_filter_results,omitempty"`
+	Index                int                                 `json:"index"`
+	Delta                ChatCompletionStreamChoiceDelta     `json:"delta"`
+	Logprobs             *ChatCompletionStreamChoiceLogprobs `json:"logprobs,omitempty"`
+	FinishReason         FinishReason                        `json:"finish_reason"`
+	ContentFilterResults ContentFilterResults                `json:"content_filter_results,omitempty"`
 }
 
 type PromptFilterResult struct {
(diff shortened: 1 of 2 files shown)


chat_stream.go
```diff
@@ -10,13 +10,33 @@ type ChatCompletionStreamChoiceDelta struct {
...
 }
 
 type ChatCompletionStreamChoice struct {
-	Index                int                             `json:"index"`
-	Delta                ChatCompletionStreamChoiceDelta `json:"delta"`
-	FinishReason         FinishReason                    `json:"finish_reason"`
-	ContentFilterResults ContentFilterResults            `json:"content_filter_results,omitempty"`
+	Index                int                                  `json:"index"`
+	Delta                ChatCompletionStreamChoiceDelta      `json:"delta"`
+	Logprobs             []ChatCompletionStreamChoiceLogprobs `json:"logprobs,omitempty"`
```
sashabaranov, line 37: Hey, thank you for the PR!

OpenAI doc mentions that this is an `object or null,` so I'm curious if we should use the slice here.

Also would appreciate an integration test, making sure all these details work as intended is pretty challenging these days!

image

chat_stream.go
```diff
@@ -10,13 +10,33 @@ type ChatCompletionStreamChoiceDelta struct {
...
 }
 
 type ChatCompletionStreamChoice struct {
-	Index                int                             `json:"index"`
-	Delta                ChatCompletionStreamChoiceDelta `json:"delta"`
-	FinishReason         FinishReason                    `json:"finish_reason"`
-	ContentFilterResults ContentFilterResults            `json:"content_filter_results,omitempty"`
+	Index                int                                `json:"index"`
+	Delta                ChatCompletionStreamChoiceDelta    `json:"delta"`
+	Logprobs             ChatCompletionStreamChoiceLogprobs `json:"logprobs,omitempty"`
```
sashabaranov, line 37: Sorry, one more quirk here: it's an `object or null`, so this should be a pointer probably
```suggestion
	Logprobs             *ChatCompletionStreamChoiceLogprobs `json:"logprobs,omitempty"`
```

Yu0u: Ok, I modified it here.

sashabaranov (approved):
Thank you!

PR #883, opened 2024-10-18
fix: Updated Assistent struct with latest fields based on OpenAI docs

ecoralic:
**Describe the change**
Added missing fields to Assistent struct.

https://platform.openai.com/docs/api-reference/assistants/object

**Describe your solution**
Updated Assistent struct based on OpenAI docs. Also synced order based on docs.

### DIFF
--- a/assistant.go
+++ b/assistant.go
@@ -14,17 +14,20 @@ const (
 )
 
 type Assistant struct {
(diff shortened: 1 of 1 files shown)


sashabaranov (approved):
Thank you!

PR #885, opened 2024-10-22
Updated checkPromptType function to handle prompt list in completions

AyushSawant18588:
**Describe the change**
The checkPromptType function has been updated to support validation of []interface{} slices, ensuring that each element in the slice is of type string. 
Currently if list of string is provided in prompt field ErrCompletionRequestPromptTypeNotSupported error is returned.
Example:
for this completions request-
```
{
	"model": "gemma-vllm",
	"prompt": ["Explain transformers in atelast 100 words or more","Hello whoare you"],
	"max_tokens": 150
}
```
This error will occur-
```
Completion error: the type of CompletionRequest.Prompt only supports string and []string
```
This is because arrays in JSON are decoded into Go as slices of interface{} ([]interface{}), not as slices of a specific type like []string. Therefore, when a JSON request containing a list of strings is unmarshalled, the resulting value of request.Prompt is a []interface{}. The checkPromptType function only checked for string and []string, so its returning false when request.Prompt is a []interface{} even if all elements within that slice are strings, because []interface{} is not the same type as []string.

This change allows better validation of prompt types, accommodating the scenario where the prompt is provided as a list of strings in an interface{} slice.

**Describe your solution**
The solution involves adding a new type check for []interface{} in the checkPromptType function. The code checks if prompt is of type []interface{} and, if so, iterates over the slice to ensure all elements are strings. If any element is not a string, the validation fails. This logic is integrated alongside the existing checks for string and []string types. This enhancement ensures that the function properly handles various input types for prompts and raises the ErrCompletionRequestPromptTypeNotSupported error if an unsupported type is provided.


johnugeorge:
@sashabaranov  Can you review? Currently, completions api is not working due to this bug

completion.go
```diff
@@ -161,7 +161,17 @@ func checkEndpointSupportsModel(endpoint, model string) bool {
...
+	if slice, isSlice := prompt.([]interface{}); isSlice {
+		isInterfaceSlice = true
+		for _, item := range slice {
+			if _, isStringItem := item.(string); !isStringItem {
+				isInterfaceSlice = false
+				break
+			}
+		}
+	}
+	return isString || isStringSlice || isInterfaceSlice
```
sashabaranov, line 174: @AyushSawant18588 thank you for the PR! Could we please use a less nested version of this with early returns? Something like this https://go.dev/play/p/cS9jrYnWrvN

AyushSawant18588: Yeah sure I will check and update the PR

AyushSawant18588: I have updated the PR @sashabaranov, can you please take a look?

sashabaranov (approved):

sashabaranov:
Thank you!

sashabaranov:
@AyushSawant18588 sure, I was waiting for #882 to be merged, but it seems it would take some time!

Note: you can always swap to your fork as well!

PR #890, opened 2024-11-03
Support Attachments in MessageRequest

kodernubie:
**Describe the change**
Add Attachments field in MessageRequest struct so we can attach file to thread message and can refer the file in chat content.

https://platform.openai.com/docs/api-reference/messages/createMessage

**Describe your solution**
Adding Attachments Field and tools type const in message.go

**Tests**
using unit test and example app

**Additional context**
-

Issue: -

### DIFF
--- a/messages.go
+++ b/messages.go
@@ -52,10 +52,11 @@ type ImageFile struct {
 }
 
 type MessageRequest struct {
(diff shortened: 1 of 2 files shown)


sashabaranov (approved):
Thank you!

PR #891, opened 2024-11-06
Added Vector Store File List properties that allow for pagination

MattDavisRV:
**Describe the change**

This change adds pagniation properties for listing Vector Store Files.  A vector store can handle up to 10,000 files at the time of this PR and a max of 100 files can be returned per request.

It's not specifically outlined in the documentation, however, the example response has these fields listed.  

https://platform.openai.com/docs/api-reference/vector-stores-files/listFiles

**Describe your solution**

Added the missing properties that are returned from the API request.


sashabaranov (approved):
Thank you for the PR

PR #900, opened 2024-11-14
Updated client_test to solve lint error

AyushSawant18588:
**Describe the change**
Latest version of golangci-lint is used in github workflows to check for lint errors. The [recent release](https://github.com/golangci/golangci-lint/releases/tag/v1.62.0) of golangci-lint introduced additional linting rules, which surfaced a new [lint error](https://github.com/sashabaranov/go-openai/actions/runs/11833389300/job/32971951797#step:5:25) in client_test.go file. Also linter execinquery is deprecated and linter gomnd is replaced by mnd [more info](https://github.com/golangci/golangci-lint/pull/5110)
This PR addresses and resolves that newly reported lint error and removes the deprecated linters.

**Describe your solution**
PR adds a type assertion to check if the recovered panic value (r) is a string before comparing it to the expected panic message (tt.wantPanic). If r is indeed a string, it verifies that the panic message matches the expected output.If r is not a string, the code now logs a message indicating that the panic value was of an unexpected type.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -169,7 +169,7 @@ linters:
     - errorlint # errorlint is a linter for that can be used to find code that will cause problems with the error wrapping scheme introduced in Go 1.13.
     # Removed execinquery (deprecated). execinquery is a linter about query string checker in Query function which reads your Go src files and warning it finds
     - exhaustive # check exhaustiveness of enum switch statements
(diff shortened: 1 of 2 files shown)


sashabaranov (approved):
Thank you for the PR!

PR #902, opened 2024-11-19
Allow streaming for O1 model family

handsomefox:
**Describe the change**
Don't return error if streaming is enable for o1 model family.

https://platform.openai.com/docs/api-reference/streaming

**Additional context**
https://x.com/OpenAIDevs/status/1858609150999359559


sashabaranov:
@handsomefox thank you for the PR! Please the latest changes from `master` here, so the CI will pass

PR #905, opened 2024-11-21
feat: add gpt-4o-2024-11-20 model

liushuangls:
add gpt-4o-2024-11-20 model


joglea (approved):

joglea (approved):

sashabaranov (approved):
Thank you for the PR! 

I wonder if we should refactor `disabledModelsForEndpoints`, so it wouldn't require a change when adding new models

Issue #898, opened 2024-11-13
Feature Request: Support extra_body Parameter for Advanced Model Features


wqj97:
Background
vLLM currently supports various model features through configuration parameters, but lacks support for passing additional model-specific parameters through extra_body, which is particularly important for features like structured output.
https://github.com/vllm-project/vllm/blob/v0.6.0/vllm/engine/arg_utils.py#L276

# Current OpenAI implementation
```python
completion = client.chat.completions.create(
    model="gpt-3.5-turbo",
    messages=[{"role": "user", "content": "Generate a user profile"}],
    extra_body={
        "guided_json": Test.schema_json,
        "guided_decoding_backend": "lm-format-enforcer"
    }
)
```

# Proposed implementation
```go
resp, err := integrations.LLMClient.Client.CreateChatCompletion(
		ctx,
		openai.ChatCompletionRequest{
			Model: "...",
			Messages: []openai.ChatCompletionMessage{
				...
			},
			ExtraBody: map[string]any{
				....
			},
		},
	)
```

blevz:
This is also necessary for extended reasoning support with anthropic via OpenAI sdk: https://docs.anthropic.com/en/api/openai-sdk#extended-thinking-support

SteveHere:
OpenRouter similarly needs this to choose & sort providers.


https://openrouter.ai/docs/community/open-ai-sdk

From the Python version of the OpenAI SDK docs:

> \# pass extra_body to access OpenRouter-only arguments.


The `extra_body` object is needed to pass the `provider` object:

https://openrouter.ai/docs/features/provider-routing

PR #906, opened 2024-11-23
Support for extra_body parameter for embeddings API

AyushSawant18588:
**Describe the change**
This change adds support for an ExtraBody parameter in the EmbeddingRequest struct, allowing developers to include additional, arbitrary key-value pairs in the API request payload. This ensures extensibility and flexibility when working with the OpenAI Embeddings API, especially for experimental features, future API updates, or custom parameters that are not explicitly defined in the current request structure. The addition of ExtraBody is particularly valuable for supporting LLM engines like vLLM, Nvidia NIM, etc, which may require additional fields beyond the standard parameters. [Refrence]( https://build.nvidia.com/nvidia/nv-embedqa-e5-v5?snippet_tab=Python)
There is an issue created that ExtraBody parameter support is required for vLLM engine also. https://github.com/sashabaranov/go-openai/issues/898
In this PR ExtraBody parameter support is added only for embedding API but it can be extended for other APIs as well like chat completions

OpenAI's official python client also supports this extra_body field which this go client could also support. [Refrence](https://github.com/openai/openai-python/blob/83091e96cf43f344d22799c22eea301aeae36d51/src/openai/resources/embeddings.py#L58C9-L58C19)

**Describe your solution**
The solution involves:
Introducing the ExtraBody field in the EmbeddingRequest struct. This is a map of string keys to arbitrary values (map[string]any) that allows developers to pass additional fields to the API dynamically.
Updating the CreateEmbeddings method to merge the ExtraBody values into the main request payload (body) before sending it to the API. This is done through the withExtraBody request option, which ensures the fields in ExtraBody are included without overwriting existing mandatory parameters.

**Tests**
Added unit tests and done sanity check of embeddings api with extra_body param for Nvidia NIM embedding model


AyushSawant18588:
@sashabaranov Can you please look into this?

embeddings.go
```diff
@@ -241,11 +252,29 @@ func (c *Client) CreateEmbeddings(
...
+	body["model"] = baseReq.Model
+	if baseReq.User != "" {
+		body["user"] = baseReq.User
+	}
+	if baseReq.EncodingFormat != "" {
+		body["encoding_format"] = baseReq.EncodingFormat
+	}
+	if baseReq.Dimensions > 0 { // Assuming 0 means the field is not set
+		body["dimensions"] = baseReq.Dimensions
+	}
```
sashabaranov, line 271: Thank you for the PR! Do I understand correctly that we'll need to extend logic here every time `EmbeddingRequest` gets updated?

sashabaranov: If so, could we maybe serialize `baseReq` to json and than back to `map[string]any`? That would allow us not to keep this logic updated

AyushSawant18588: yes will need to extend logic here also if EmbeddingRequest gets updated. But yeah I think to support extra_body param this approach is the way to do so in Golang.

AyushSawant18588: > If so, could we maybe serialize `baseReq` to json and then back to `map[string]any`? That would allow us not to keep this logic updated

Okay will look into it and update the PR

AyushSawant18588: @sashabaranov I have updated the PR and tested the changes, can you please look into the changes?

gavrissh:
@sashabaranov Would require your assistance in getting these changes in. Thank you

nagar-ajay:
@sashabaranov can you please check this? Thanks!

sashabaranov:
@AyushSawant18588 Thank you for updating the PR! I'm re-reading it for the fifth time, and there's something odd about it—and I can't pinpoint what it is exactly. It feels like we're adding a lot of counterintuitive logic for a tiny use case.

If the goal is to add `{"input_type": "query", "truncate": "NONE"}` to Embedding requests, maybe that's literally what we should do — just add two fields to `EmbeddingRequest`. Or are there more use cases I don't know of?

AyushSawant18588:
So the official OpenAI Python client offers an extra_body field for all inference APIs. For example, in the embedding API implementation, you can see this feature being used here: [Embeddings API](https://github.com/openai/openai-python/blob/d9c966dea77fa3493114865a7f785f3134f1cc1e/src/openai/resources/embeddings.py#L58). Their implementation and behaviour is same as done in this PR. 
The extra_body parameter allows users to add arbitrary JSON properties to the request body. This approach is particularly useful when working with diverse LLM engines that may introduce additional or custom fields in their APIs. Instead of updating the struct for every unique field, this mechanism provides flexibility to include extra fields as needed without modifying the core structure repeatedly.
Here are some examples where extra parameters may vary depending on the API and LLM engine:
- https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html?ref=blog.mozilla.ai#extra-parameters
- https://build.nvidia.com/nvidia/nv-embedqa-e5-v5?snippet_tab=Python

johnugeorge:
@sashabaranov 
1. We see that all embedding servers including vLLM, Nvidia NIM support "extra_body" in the embedding request.  
2. ExtraBody is supported in official [OpenAI Python clients](https://github.com/openai/openai-python/blob/83091e96cf43f344d22799c22eea301aeae36d51/src/openai/resources/embeddings.py#L58C9-L58C19).
3. `"input_type": "query", "truncate": "NONE"` cannot be added to the embedding request as it will break OpenAI compatible server implementations(it is not compatible with OpenAI API standard)

Can you please look into this?

lilien1010:
@sashabaranov Help!

Justa-Cai:
The Qwen3 LLM now supports switching between reasoning and non-reasoning modes, which can be controlled using this parameter. Would it be possible to merge this change?

embeddings.go
```diff
@@ -241,11 +253,29 @@ func (c *Client) CreateEmbeddings(
...
+
+	// Serialize baseReq to JSON
+	jsonData, err := json.Marshal(baseReq)
+	if err != nil {
+		return
+	}
+
+	// Deserialize JSON to map[string]any
+	var body map[string]any
+	_ = json.Unmarshal(jsonData, &body)
```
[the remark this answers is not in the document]
sashabaranov: @AyushSawant18588 just some tech debt :)

PR #919, opened 2025-01-02
files: Don't truncate filename to just the basename

jlewi:
**Describe the change**
When creating OpenAI Files, I want to be able to set the Filename to a file URI and not just the result of path.Base(filename).

The Assistants API will return the FileName in responses. If we truncate the URI we may not be able to locate the file. 
If a caller wants to just use the baseName then they can set this ins the [FileBytesRequest](https://github.com/jlewi/go-openai/blob/2a0ff5ac63e460cbe44cccd0d4199d51bf8682a4/files.go#L29) by setting `Name` as desired.

Currently [CreateFileBytes](https://github.com/sashabaranov/go-openai/blob/2a0ff5ac63e460cbe44cccd0d4199d51bf8682a4/files.go#L70) passes the user set Name to [CreateFormFileReader](https://github.com/sashabaranov/go-openai/blob/2a0ff5ac63e460cbe44cccd0d4199d51bf8682a4/internal/form_builder.go#L34) which truncates the name to the basename.

The PR would be a breaking change for anyone relying on the current behavior. If a non breaking change is preferred we could add options or a different method.

Provide a relevant API doc from https://platform.openai.com/docs/api-reference
https://platform.openai.com/docs/api-reference/files/object

**Describe your solution**
Describe how your changes address the problem or how they add the feature. This should include a brief description of your approach and any new libraries or dependencies you're using.

Remove the call to path.Base

**Tests**
Briefly describe how you have tested these changes. If possible — please add integration tests.

**Additional context**

Issue: #XXXX


jlewi:
I'm not sure why golang is failing with the following error

```
run golangci-lint
  Running [/home/runner/golangci-lint-1.63.2-linux-amd64/golangci-lint run --out-format=github-actions] in [] ...
  Error: Comment should end in a period (godot)
```

I suspect its not coming from this PR as there are no comments added in this PR. The output doesn't indicate the light number in question.

PR #926, opened 2025-01-24
Make "Content" field in "ChatCompletionMessage" omitempty

saileshd1402:
**Describe the change**
According to the OpenAI API spec, the content field of ChatCompletionMessage should be nullable. Right now when we try to marshal the struct with empty string, we get `"content": ""`, but ideally, the content key should exist if that field is empty

https://platform.openai.com/docs/api-reference/chat/object
Under: choices -> message -> content
image

**Describe your solution**
We can fix this my adding an omitempty json tag to the "content" field in the "ChatCompletionMessage" struct

**Tests**
Tested this by sending a completion request with empty content field to a Nvidia NIM Inference Server

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -93,7 +93,7 @@ type ChatMessagePart struct {
 
 type ChatCompletionMessage struct {
 	Role         string `json:"role"`
-	Content      string `json:"content"`
+	Content      string `json:"content,omitempty"`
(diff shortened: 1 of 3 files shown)


Datta0 (approved):
LGTM

sashabaranov (approved):
Thank you!

PR #930, opened 2025-01-31
Add support for O3-mini

rorymalcolm:
- Add support for the o3 mini set of models, including tests that match the constraints in OpenAI's API docs (https://platform.openai.com/docs/models#o3-mini).

### DIFF
--- a/chat_test.go
+++ b/chat_test.go
@@ -64,7 +64,7 @@ func TestO1ModelsChatCompletionsDeprecatedFields(t *testing.T) {
 				MaxTokens: 5,
 				Model:     openai.O1Preview,
 			},
-			expectedError: openai.ErrO3MaxTokensDeprecated,
+			expectedError: openai.ErrReasoningModelMaxTokensDeprecated,
 		},
 		{
 			name: "o1-mini_MaxTokens_deprecated",
@@ -72,7 +72,7 @@ func TestO1ModelsChatCompletionsDeprecatedFields(t *testing.T) {
 				MaxTokens: 5,
 				Model:     openai.O1Mini,
 			},
-			expectedError: openai.ErrO3MaxTokensDeprecated,
+			expectedError: openai.ErrReasoningModelMaxTokensDeprecated,
 		},
 	}
 
@@ -104,7 +104,7 @@ func TestO1ModelsChatCompletionsBetaLimitations(t *testing.T) {
 				LogProbs:            true,
 				Model:               openai.O1Preview,
 			},
-			expectedError: openai.ErrO3BetaLimitationsLogprobs,
+			expectedError: openai.ErrReasoningModelLimitationsLogprobs,
 		},
 		{
 			name: "message_type_unsupported",
@@ -155,7 +155,7 @@ func TestO1ModelsChatCompletionsBetaLimitations(t *testing.T) {
 				},
 				Temperature: float32(2),
 			},
-			expectedError: openai.ErrO3BetaLimitationsOther,
+			expectedError: openai.ErrReasoningModelLimitationsOther,
 		},
 		{
 			name: "set_top_unsupported",
@@ -173,7 +173,7 @@ func TestO1ModelsChatCompletionsBetaLimitations(t *testing.T) {
 				Temperature: float32(1),
 				TopP:        float32(0.1),
 			},
-			expectedError: openai.ErrO3BetaLimitationsOther,
+			expectedError: openai.ErrReasoningModelLimitationsOther,
 		},
 		{
 			name: "set_n_unsupported",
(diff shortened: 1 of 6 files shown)


1x-eng:
yes please!

completion.go
```diff
@@ -241,6 +256,37 @@ func validateRequestForO1Models(request ChatCompletionRequest) error {
...
+	}
+	if request.N > 0 && request.N != 1 {
+		return ErrO3BetaLimitationsOther
+	}
+	if request.PresencePenalty > 0 {
+		return ErrO3BetaLimitationsOther
+	}
+	if request.FrequencyPenalty > 0 {
+		return ErrO3BetaLimitationsOther
+	}
```
sashabaranov, line 285: I suggest we mark `ErrO1BetaLimitations*` as deprecated and introduce a single set of errors and a single validation function. Currently, it feels like a pure duplication

rorymalcolm: Thanks, will submit this patch later today (as in, in next few hours)

rorymalcolm: The only issue with this is some components (eg. Function calling) are supported in O3, but not O1

Going to keep those earlier errors (and tests) around - but will remove the rest in favour of O3

rorymalcolm: Done 👍

sash20m:
Thanks. I need it.

chat_test.go
```diff
@@ -64,15 +64,15 @@ func TestO1ModelsChatCompletionsDeprecatedFields(t *testing.T) {
 				MaxTokens: 5,
 				Model:     openai.O1Preview,
 			},
-			expectedError: openai.ErrO1MaxTokensDeprecated,
+			expectedError: openai.ErrO3MaxTokensDeprecated,
```
sashabaranov, line 67: So this is a bit weird — we're requesting an o1 model but are getting an o3 error instead. I suggest we name these errors more generic like use `ReasoningModel` instead of `O3` — this would future-proof us for o4, o5, etc.

completion.go
```diff
@@ -105,12 +115,20 @@ var O1SeriesModels = map[string]struct{}{
 	O1Preview20240912: {},
 }
 
+// O3SeriesModels List of new Series of OpenAI models.
+var O3SeriesModels = map[string]struct{}{
+	O3Mini:         {},
+	O3Mini20250131: {},
+}
```
sashabaranov, line 122: We'll need to bookkeep this map forever and ever, could we use the `strings.HasPrefix("o3-")` instead?

completion.go
```diff
@@ -192,52 +210,75 @@ var availableMessageRoleForO1Models = map[string]struct{}{
...
 
-	// Tools: tools, function calling, and response format parameters are not supported
-	for _, t := range request.Tools {
-		if _, found := unsupportedToolsForO1Models[t.Type]; found {
-			return ErrO1BetaLimitationsTools
-		}
+	return nil
+}
+
+func validateBasicParams(request ChatCompletionRequest) error {
```
sashabaranov, line 240: The method name is pretty generic, let's make it's name clearly indicate that it is used only for reasoning models. Maybe even put all the logic related to reasoning models into a separate file (struct?)

sashabaranov:
Thank you for the updates! I left a few minor comments, but it generally looks good 🙌🏻 

Ah, we also miss a test case in `chat_stream.go`!

chat_test.go
sash20m: From what I know they're intending to merge the models in the near future, the normal and reasoning capabilities will be under the same model. Most likely we'll specify to use reasoning with an option param in the API. 

Just fyi, if it changes anything about the implementation.

rorymalcolm:
> Thank you for the updates! I left a few minor comments, but it generally looks good 🙌🏻
> 
> Ah, we also miss a test case in `chat_stream.go`!

Now done

1x-eng:
@sashabaranov is this merge ready? i need `o3-mini` sooner than later, switching to OAI's official go sdk is a pain id rather avoid. plmk if this can go in soon-ish? thanks

reasoning_validator.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, m := range v.request.Messages {
+		if _, found := availableMessageRoleForO1Models[m.Role]; !found {
+			return ErrO1BetaLimitationsMessageTypes
+		}
+	}
+
+	for _, t := range v.request.Tools {
+		if _, found := unsupportedToolsForO1Models[t.Type]; found {
+			return ErrO1BetaLimitationsTools
+		}
```
sashabaranov, line 77: Wow, this implementation is so clean, thank you! ❤️ 

A couple of nitpicks before we merge:
- Let's move `ErrO1BetaLimitationsTools` and others into this file
- Let's move `unsupportedToolsForO1Models` to this file
- We're still missing a test here: [codecov link](https://app.codecov.io/gh/sashabaranov/go-openai/pull/925?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Alex+Baranov#03819010d8bb481405252d78cd1cd9d9-R84)

rorymalcolm: Will get these done today

rorymalcolm: Now done, good to merge, I think?

Issue #781, opened 2024-07-01
Inability to Parse API Response Data Due to Missing Space After "data:"


fruitbars:
**Describe the bug**
The bug occurs when the response data from the API does not include a space after the colon in the "data:" prefix. This causes the data to not be properly parsed or read by the client.

**To Reproduce**
1. Make a request to the OpenAI API endpoint that returns streaming data.
2. Ensure that the response includes a "data:" prefix without a space, like so: `data:{"code":0,"message":"Success","sid":"xxx","id":"xxx","created":1719810561,"choices":[{"delta":{"role":"assistant","content":"He"},"index":0}]}`.
3. Attempt to read or parse the response data.

**Expected behavior**
The client should be able to correctly parse and read the response data regardless of whether there is a space after the "data:" prefix.

**Screenshots/Logs**
Unfortunately, no screenshots are available, but the logs would show that the data is not being processed as expected due to the missing space.

**Environment**
- go-openai version: Not specified
- Go version: Not specified
- OpenAI API version: Not specified
- OS: Not specified

**Additional context**
This issue might be related to how the client library handles the parsing of streaming data responses. It may require adjustments in the parsing logic to accommodate responses with and without a space after the "data:" prefix.

stmcginnis:
This isn't impacting me, but I think this is a legitimate bug.

I had originally assumed the spec was `data: ` with a space, but the spec actually does call out that the space is not required:

```txt
 The following stream fires two identical events:

 data:test

 data: test

This is because the space after the colon is ignored if present.
```

PR #932, opened 2025-02-07
fix: Inability to Parse API Response Data Due to Missing Space After …

sunhailin-Leo:
### Fixed
* fix the problem in https://github.com/sashabaranov/go-openai/issues/781

### DIFF
--- a/stream_reader.go
+++ b/stream_reader.go
@@ -90,9 +90,6 @@ func (stream *streamReader[T]) handleReadError() ([]byte, error) {
 func (stream *streamReader[T]) parseLine(rawLine []byte) ([]byte, []byte) {
 	name, value, _ := bytes.Cut(rawLine, []byte(":"))
 	value = bytes.TrimSpace(value)
-	if len(value) > 0 && value[0] == ' ' {
-		value = value[1:]
-	}
 	return name, value
 }
 
@@ -109,7 +106,6 @@ func (stream *streamReader[T]) handleDataFlag(value []byte) ([]byte, error) {
 		if respErr != nil {
 			return nil, fmt.Errorf("error, %w", respErr.Error)
 		}
-		return nil, nil
 	}
 	return value, nil
 }
--- a/stream_reader.go
+++ b/stream_reader.go
@@ -11,8 +11,8 @@ import (
 )
 
 var (
-	headerData  = []byte("data: ")
-	errorPrefix = []byte(`data: {"error":`)
+	headerData  = []byte("data:")
+	errorPrefix = []byte(`{"error":`)
 )
 
 type streamable interface {
@@ -57,6 +57,7 @@ func (stream *streamReader[T]) processLines() ([]byte, error) {
 	var (
 		emptyMessagesCount uint
 		hasErrorPrefix     bool
+		noPrefixLine       []byte
 	)
 
 	for {
@@ -70,18 +71,19 @@ func (stream *streamReader[T]) processLines() ([]byte, error) {
 		}
 
 		noSpaceLine := bytes.TrimSpace(rawLine)
-		if bytes.HasPrefix(noSpaceLine, errorPrefix) || bytes.HasPrefix(noSpaceLine, bytes.TrimSpace(errorPrefix)) {
-			hasErrorPrefix = true
+		if bytes.HasPrefix(noSpaceLine, headerData) {
+			noPrefixLine = bytes.TrimPrefix(noSpaceLine, headerData)
+			if bytes.HasPrefix(noPrefixLine, []byte(" ")) {
+				noPrefixLine = bytes.TrimSpace(noPrefixLine)
+			}
+
+			if bytes.HasPrefix(noPrefixLine, errorPrefix) {
+				hasErrorPrefix = true
+			}
 		}
+
 		if !bytes.HasPrefix(noSpaceLine, headerData) || hasErrorPrefix {
-			if hasErrorPrefix {
-				noSpaceLine = bytes.TrimPrefix(noSpaceLine, headerData)
-			}
-			if bytes.HasPrefix(noSpaceLine, bytes.TrimSpace(headerData)) {
-				noSpaceLine = bytes.TrimPrefix(noSpaceLine, bytes.TrimSpace(headerData))
-				goto line
-			}
-			writeErr := stream.errAccumulator.Write(noSpaceLine)
+			writeErr := stream.errAccumulator.Write(noPrefixLine)
 			if writeErr != nil {
 				return nil, writeErr
 			}
@@ -93,8 +95,6 @@ func (stream *streamReader[T]) processLines() ([]byte, error) {
 			continue
 		}
 
-	line:
-		noPrefixLine := bytes.TrimPrefix(noSpaceLine, headerData)
 		if string(noPrefixLine) == "[DONE]" {
 			stream.isFinished = true
 			return nil, io.EOF
--- a/stream_reader.go
+++ b/stream_reader.go
@@ -11,8 +11,10 @@ import (
 )
 
 var (
-	headerData  = []byte("data:")
-	errorPrefix = []byte(`{"error":`)
+	dataFlag        = "data"
+	doneFlag        = []byte("[DONE]")
+	errorFlag       = []byte(`"error"`)
+	errorPrefixFlag = []byte(`{"error`)
 )
 
 type streamable interface {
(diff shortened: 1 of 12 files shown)


stream_reader.go
```diff
@@ -70,13 +70,17 @@ func (stream *streamReader[T]) processLines() ([]byte, error) {
...
 		noSpaceLine := bytes.TrimSpace(rawLine)
-		if bytes.HasPrefix(noSpaceLine, errorPrefix) {
+		if bytes.HasPrefix(noSpaceLine, errorPrefix) || bytes.HasPrefix(noSpaceLine, bytes.TrimSpace(errorPrefix)) {
 			hasErrorPrefix = true
 		}
 		if !bytes.HasPrefix(noSpaceLine, headerData) || hasErrorPrefix {
 			if hasErrorPrefix {
 				noSpaceLine = bytes.TrimPrefix(noSpaceLine, headerData)
 			}
+			if bytes.HasPrefix(noSpaceLine, bytes.TrimSpace(headerData)) {
```
sashabaranov, line 80: Thank you for the PR! Shouldn't we default `headerData` to not contain space in the end?

stream_reader_test.go
```diff
@@ -75,4 +75,16 @@ func TestStreamReaderRecvRaw(t *testing.T) {
 	if !bytes.Equal(rawLine, []byte("{\"key\": \"value\"}")) {
 		t.Fatalf("Did not return raw line: %v", string(rawLine))
 	}
+
+	// issue: https://github.com/sashabaranov/go-openai/issues/781
```
sashabaranov, line 79: No need for this comment, `git blame` will point to this PR/issue anyways

stream_reader.go
sunhailin-Leo: In various LLM inference frameworks, response stream data may begin with `data: ` or `data`. 
So, i think we should compatible different response prefix.

stream_reader_test.go
sunhailin-Leo: ok, got it~

stream_reader.go
sashabaranov: So it always starts with `data:`, right? This is what we should check for

sunhailin-Leo: In fact, there are two forms of response (`data:` and `data: `) in different LLM stream API.
At present, there seems to be no unified standard that the response result should be `data:` or `data: `

In `openai/openai-go`, it use `bytes.Cut(txt, []byte(":"))` to solve this problem. 
https://github.com/openai/openai-go/blob/main/packages/ssestream/ssestream.go#L80

sunhailin-Leo: Also in `openai/openai-python`, it use `str.partition` to solve this problem.
https://github.com/openai/openai-python/blob/main/src/openai/_streaming.py#L346-L349

sunhailin-Leo: All in all, I think this is due to the lack of mandatory standards in SSE. The SSE standard also mentions that if there are spaces, they should be replaced for event value processing.
https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation

sunhailin-Leo: Back to this pull request, to solve this issue now, may handle it in my way, or refactor the processing function according to `openai-go` way. 😄

sashabaranov: @sunhailin-Leo I think the solution is pretty straightforward:
1. Check if the starts with `data:`
2. If yes, trim `data:`
3. [optional] If the rest of the line starts with a space, trim it as well

sunhailin-Leo: So, you mean refactor it? I think `headerData` is easy to check, but `errorPrefix` will change the whole logic.

sunhailin-Leo:
@sashabaranov 
* Have any other problem or question in this pull-request?

sashabaranov:
@sunhailin-Leo Thank you for updates, the approach looks good, but the `processLines()` is really hard to read. Partly because it has `//nolint:gocognit` before it disabling cognitive complexity linter (cc @liushuangls)

I propose we remove `//nolint:gocognit` and try to simplify `processLines()`. I can't understand what this method is doing by reading the diff.

Here's an idea to simplify, but I'm not sure it's correct

```go
func trimAndCompare(s []byte, cutset string) ([]byte, bool) {
	originalLen = len(s)
	updated := bytes.Trim(s, cutset)
	lengthChanged := originalLen != len(updated)
	return updated, lengthChanged
}

func (stream *streamReader[T]) processLines() ([]byte, error) {
	var (
		emptyMessagesCount uint
		hasErrorPrefix     bool
		noPrefixLine       []byte
	)

	for {
		line, readErr := stream.reader.ReadBytes('\n')
		if readErr != nil || hasErrorPrefix {
			respErr := stream.unmarshalError()
			if respErr != nil {
				return nil, fmt.Errorf("error, %w", respErr.Error)
			}
			return nil, readErr
		}

		line = bytes.TrimSpace(rawLine)
		line, gotHeader = trimAndCompare(line, headerData)
		line = bytes.TrimSpace(line)
		if gotHeader {
			hasErrorPrefix = bytes.HasPrefix(line, errorPrefix)
		}
		if !gotHeader || hasErrorPrefix {
			writeErr := stream.errAccumulator.Write(line)
			if writeErr != nil {
				return nil, writeErr
			}
			emptyMessagesCount++
			if emptyMessagesCount > stream.emptyMessagesLimit {
				return nil, ErrTooManyEmptyStreamMessages
			}

			continue
		}

		if string(noPrefixLine) == "[DONE]" {
			stream.isFinished = true
			return nil, io.EOF
		}

		return noPrefixLine, nil
	}
}
```

sashabaranov:
Side idea: maybe we can re-factor this as a state machine here, with states clearly enumerated

sunhailin-Leo:
> Side idea: maybe we can re-factor this as a state machine here, with states clearly enumerated

If want to re-factor, I think we can refer to https://github.com/openai/openai-go/blob/main/packages/ssestream/ssestream.go#L80

sashabaranov:
@sunhailin-Leo that's a way to go as well!

sunhailin-Leo:
@sashabaranov
* I think this implementation may be better than before.

```go
var (
	//headerData      = []byte("data:")
	//errorPrefix     = []byte(`{"error":`)
	doneFlag        = []byte("[DONE]")
	dataFlag        = []byte("data")
	errorFlag       = []byte(`"error"`)
	errorPrefixFlag = []byte(`{"error`)
)


func (stream *streamReader[T]) processLines() ([]byte, error) {
	var (
		emptyMessagesCount uint
		hasErrorPrefix     bool
	)

	for {
		rawLine, readErr := stream.reader.ReadBytes('\n')
		if readErr != nil || (hasErrorPrefix && readErr == io.EOF) {
			respErr := stream.unmarshalError()
			if respErr != nil {
				return nil, fmt.Errorf("error, %w", respErr.Error)
			}
			return nil, readErr
		}

		// Split a string like "event: bar" into name="event" and value=" bar".
		name, value, _ := bytes.Cut(rawLine, []byte(":"))
		value = bytes.TrimSpace(value)

		// Consume an optional space after the colon if it exists.
		if len(value) > 0 && value[0] == ' ' {
			value = value[1:]
		}

		switch name {
		case dataFlag:
			if bytes.Equal(value, doneFlag) {
				stream.isFinished = true
				return nil, io.EOF
			}
			if bytes.HasPrefix(value, errorPrefixFlag) {
				if writeErr := stream.writeErrAccumulator(value); writeErr != nil {
					return nil, writeErr
				}
				respErr := stream.unmarshalError()
				if respErr != nil {
					return nil, fmt.Errorf("error, %w", respErr.Error)
				}
				continue
			}

			return value, nil
		default:
			if writeErr := stream.writeErrAccumulator(rawLine); writeErr != nil {
				return nil, writeErr
			}
			if bytes.Equal(name, errorFlag) {
				hasErrorPrefix = true
				continue
			}

			emptyMessagesCount++
			if emptyMessagesCount > stream.emptyMessagesLimit {
				return nil, ErrTooManyEmptyStreamMessages
			}

			continue
		}
	}
}
```

sunhailin-Leo:
@sashabaranov 
* There are some problems in `golangci.yml` ==> `linters-settings > varcheck`

sashabaranov:
@sunhailin-Leo Could you please rebase on the latest master for CI to pass?

sunhailin-Leo:
> There's also this PR: #941 @sunhailin-Leo do you think it makes sense or not?

I think PR is pretty interesting, and it looks like the way to go with minimal changes. 
But it's still not elegant enough, and the function complexity doesn't go down.

PR #933, opened 2025-02-09
ref: add image url support to messages

mazyaryousefinia:
Added image URL support in Message content struct


sashabaranov (approved):
Thank you!

PR #936, opened 2025-02-11
feat: add developer role

liushuangls:
docs: https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages


sashabaranov (approved):
Thank you!

PR #939, opened 2025-02-15
fix: remove validateO1Specific

liushuangls:
In the latest openai o1 release o1-2024-12-17, Function calling is supported, and I test that system roles can be passed

see: https://openai.com/index/o1-and-new-tools-for-developers/

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,8 +19,6 @@ jobs:
         go vet .
     - name: Run golangci-lint
       uses: golangci/golangci-lint-action@v6
-      with:
-        version: latest
     - name: Run tests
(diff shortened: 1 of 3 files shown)


rorymalcolm:
> In the latest openai o1 release o1-2024-12-17, Function calling is supported, and I test that system roles can be passed
> 
> see: https://openai.com/index/o1-and-new-tools-for-developers/

Thanks for this work - I added that function but didn't know this was true at the time 🙏

sashabaranov (approved):
Thank you!

kanwang:
Hey folks, what's the plan/timeline to release this? is there anything we can help? 
this has introduced regression when we tried to upgrade to 1.37.0. so seems worth pushing for a release sooner.

PR #944, opened 2025-02-25
[BREAKING_CHANGES] refactor: change metadata type from map[string]string to map[string]any

anilmisirlioglu:
**Describe the change**
We use Litellm and Langfuse and I realized that the `map[string]string` data type in the metadata field is insufficient to pass some trace fields (e.g. tags, trace_metadata) from the metadata. It is possible to make metadata type `map[string]any` for more complex data types. There is no statement in the OpenAI document stating that metadata must be of type `map[string]string`

[Litellm Langfuse Integration](https://docs.litellm.ai/docs/observability/langfuse_integration#set-custom-trace-id-trace-user-id-trace-metadata-trace-version-trace-release-and-tags)
[OpenAI Metadata](https://platform.openai.com/docs/api-reference/chat/create#chat-create-metadata) -> _Set of 16 key-value pairs that can be attached to an object. This can be useful for storing additional information about the object in a **structured** format, and querying for objects via API or the dashboard._

**Describe your solution**
I changed the type of the metadata field in `ChatCompletionRequest` and `ModifyMessage` to `map[string]any`

**Tests**
I ran the tests locally.

Issue: -

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -262,7 +262,7 @@ type ChatCompletionRequest struct {
 	// Controls effort on reasoning for reasoning models. It can be set to "low", "medium", or "high".
 	ReasoningEffort string `json:"reasoning_effort,omitempty"`
 	// Metadata to store with the completion.
-	Metadata map[string]string `json:"metadata,omitempty"`
+	Metadata map[string]any `json:"metadata,omitempty"`
 }
 
 type StreamOptions struct {
--- a/messages.go
+++ b/messages.go
@@ -162,7 +162,7 @@ func (c *Client) RetrieveMessage(
 func (c *Client) ModifyMessage(
 	ctx context.Context,
 	threadID, messageID string,
-	metadata map[string]string,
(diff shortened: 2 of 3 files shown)


lordtatty:
Hi @anilmisirlioglu, the OpenAI description for this field may have changed since you opened the PR.  Looking at it now it says:

```
Keys are strings with a maximum length of 64 characters. Values are strings with a maximum length of 512 characters.
```

https://platform.openai.com/docs/api-reference/chat/create#chat-create-metadata

image

Unless I've misunderstood which metadata field this is then it looks like the keys and values both need to be strings.  Are you saying that Litellm and Langfuse uses non-strings successfully for these fields?

anilmisirlioglu:
Hi, yes langfuse uses much more complex data types like the example below. It seems that the OpenAI specification has recently changed. When I opened the PR, there was no restriction that the value could be a string.

image

PR #945, opened 2025-02-26
Adapt different stream data prefix, with or without space

panzhongxian:
This is an alternative to another PR https://github.com/sashabaranov/go-openai/pull/932

Instead of compare header string, this PR use regex pattern to match and replace the header. It seems a more clear way to implement it.


sashabaranov (approved):
Thank you! Merging this one instead of #932 due to more comprehensible diff

PR #959, opened 2025-03-27
Add CompletionRequest.StreamOptions

stmcginnis:
**Describe the change**

The legacy completion API supports [a `stream_options` object when `stream` is set to true](https://platform.openai.com/docs/api-reference/completions/create#completions-create-stream_options). This adds a StreamOptions property to the CompletionRequest struct to support this setting.

https://platform.openai.com/docs/api-reference/completions/create#completions-create-stream_options

**Describe your solution**

This adds a `StreamOptions` property (marshaled as `stream_options`) to the `CompletionRequest` struct to expose the setting. This matches the `ChatCompletionRequest` property.


sashabaranov (approved):
Thank you!

PR #912, opened 2024-12-06
Add support for 'enum' tag to JSON schema generation

Dieg0Code:
This pull request includes enhancements to the `jsonschema` package, specifically adding support for enumerations in schema definitions and improving the test coverage for schema generation and comparison.

Enhancements to schema generation:

* [`jsonschema/json.go`](diffhunk://#diff-fe6cba4594f46686187f024c0dc219341ada04dade650bf69ac8589920df792eR134-R139): Added support for parsing `enum` tags in struct fields and including them in the generated schema. [[1]](diffhunk://#diff-fe6cba4594f46686187f024c0dc219341ada04dade650bf69ac8589920df792eR134-R139) [[2]](diffhunk://#diff-fe6cba4594f46686187f024c0dc219341ada04dade650bf69ac8589920df792eR148-R152)

Improvements to test coverage:

* [`jsonschema/json_test.go`](diffhunk://#diff-fc323dd0c6b702c830de6dc0be6c7bc961eb10c058424d1c268e702ec6dd1c35R11-R236): Added new tests to verify schema generation for types with `enum` tags and to compare struct-generated schemas with manually created schemas.
* [`jsonschema/json_test.go`](diffhunk://#diff-fc323dd0c6b702c830de6dc0be6c7bc961eb10c058424d1c268e702ec6dd1c35L20-R246): Updated existing tests to remove unnecessary `properties` fields from the expected JSON output. [[1]](diffhunk://#diff-fc323dd0c6b702c830de6dc0be6c7bc961eb10c058424d1c268e702ec6dd1c35L20-R246) [[2]](diffhunk://#diff-fc323dd0c6b702c830de6dc0be6c7bc961eb10c058424d1c268e702ec6dd1c35L38-R264) [[3]](diffhunk://#diff-fc323dd0c6b702c830de6dc0be6c7bc961eb10c058424d1c268e702ec6dd1c35L69-R297) [[4]](diffhunk://#diff-fc323dd0c6b702c830de6dc0be6c7bc961eb10c058424d1c268e702ec6dd1c35L117-R352) [[5]](diffhunk://#diff-fc323dd0c6b702c830de6dc0be6c7bc961eb10c058424d1c268e702ec6dd1c35L158-R396)

PR #962, opened 2025-04-03
add enum tag to jsonschema

jtarchie:
> A similar PR may already be submitted!
> Please search among the [Pull request](https://github.com/sashabaranov/go-openai/pulls) before creating one.

There is another issue, #912. It seems to have been abandoned.

**Describe the change**

When creating a JSON schema from a `struct`, a new tag `enum` can be used to see the `Enum` field.

For example,

```go
type Car struct {
  Color string `json:"color" enum:"red,green,blue"`
}
```

This will see the `Enum` on the field definition would be `Enum = []string{"red","green","blue"}`.

This is implementation is being [used](https://github.com/jtarchie/outrageous/blob/519dce0dc981b0bf37e41890dc49b6ec6859970e/assert/assert.go#L24).

Provide a relevant API doc from https://platform.openai.com/docs/api-reference

**Describe your solution**

See change above.

**Tests**

1. A heads up, the tests on Github Actions are not running these tests. I've added support to this PR to ensure it runs correctly.
1. I've added tests for the enum field in the `json_test.go`. This also required adding some other tests for struct to schema definition.
1. Because of `omitempty`, I've updated the tests to reflect this.

**Additional context**
n/a

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -22,6 +22,6 @@ jobs:
       with:
         version: v1.64.5
     - name: Run tests
-      run: go test -race -covermode=atomic -coverprofile=coverage.out -v .
+      run: go test -race -covermode=atomic -coverprofile=coverage.out -v ./...
     - name: Upload coverage reports to Codecov
       uses: codecov/codecov-action@v4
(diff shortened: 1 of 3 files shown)


jtarchie:
I'm not sure why code coverage went down. Maybe adding more tests from Github Action?

sashabaranov (approved):
Great stuff, thank you for the PR!

PR #963, opened 2025-04-04
Add support for cache control in ChatCompletionMessage

### LINKED ISSUE #897: Add support for cache control in ChatCompletionMessage
I am using `go-openai` to request [Braintrust AI proxy](https://www.braintrust.dev/docs/guides/proxy) which allows to access models from OpenAI, Anthropic, Google, AWS, Mistral, and third-party inference providers through a single, unified API (openai).

When requesting anthropic models with prompt caching I need to perform these kind of requests

```
curl https://api.anthropic.com/v1/messages \
  -H "content-type: application/json" \
  -H "x-api-key: $ANTHROPIC_API_KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "anthropic-beta: prompt-caching-2024-07-31" \
  -d '{
    "model": "claude-3-5-sonnet-20241022",
    "max_tokens": 1024,
    "system": [
      {
        "type": "text",
        "text": "You are an AI assistant tasked with analyzing literary works. Your goal is to provide insightful commentary on themes, characters, and writing style.\n"
      },
      {
        "type": "text",
        "text": "<the entire contents of Pride and Prejudice>",
        "cache_control": {"type": "ephemeral"}
      }
    ],
    "messages": [
      {
        "role": "user",
        "content": "Analyze the major themes in Pride and Prejudice."
      }
    ]
  }'
  ```
  
Would you consider adding `CacheControl` in `ChatCompletionMessage` for this kind of use case even if it is not part of openai per se?

trashhalo:
**Describe the change**
Fixes #897 many openai compatible apis use this key to let users control token caching for models that support it.

Not a openai feature, but common in openai compatible endpoints. https://docs.litellm.ai/docs/completion/prompt_caching#anthropic-example

**Describe your solution**
Add the extra keys

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -92,6 +92,16 @@ type ChatMessagePart struct {
 	ImageURL *ChatMessageImageURL `json:"image_url,omitempty"`
 }
 
+type CacheControlType string
+
+const (
+	CacheControlTypeEphemeral CacheControlType = "ephemeral"
+)
+
+type CacheControl struct {
+	Type CacheControlType `json:"type,omitempty"`
+}
+
 type ChatCompletionMessage struct {
 	Role         string `json:"role"`
 	Content      string `json:"content,omitempty"`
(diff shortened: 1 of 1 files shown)


trashhalo:
Closing because I realized I need way more LiteLLM specific keys like citations which is returned by perplexity seems out of scope fro this project.

MohitVachhani:
We can use semantic search here ?
Like we can cache the output of a prompt and store that prompt in qdrant or so, and then on next prompt, we can do a semantic search/comparison, if it is above certain threshold then we can use the same output again or so ?

PR #968, opened 2025-04-17
Add support for `o4-mini` and `o3`

rorymalcolm:
**Describe the change**
- This adds supports, and tests, for the o3 and o4-mini class of models

https://platform.openai.com/docs/models/o4-mini
https://platform.openai.com/docs/models/o3


sashabaranov (approved):
Thank you!

Issue #601, opened 2023-11-21
Issue while invoking Edit Image API


mohitkanwar:
**Describe the bug**
While integrating with the latest Edit Image API 
`(POST https://api.openai.com/v1/images/edits)` 
by using the following code :
```golang
req := openai.ImageEditRequest{Image: file, Mask: maskFile, Prompt: prompt, N: count, Size: size, ResponseFormat: responseFormat}

	response, err = srvc.client.CreateEditImage(context.Background(), req)
```
We are getting the following error response from openAI : 

```
error, status code: 500, message: The server had an error processing your request. Sorry about that! You can retry your request, or contact us through our help center at help.openai.com if you keep seeing this error. (Please include the request ID XXXXXXXXXXX in your email.)
```

This is a consistent response. 
However, if we hit the API using HTTP client directly, it works fine.

The examples do not contain the examples on using this API, and the tests are working fine with mock. 
So I am a bit stuck on using this client for Edit Images API

**Environment :**
 - go-openai version: v1.16.0
 - Go version: 1.21
 - OpenAI API version: v1

mohitkanwar:
You may test the integration test by adding the following code :


filePath := "/image.png" // Replace with your file path

	// Open the file
	file, err := os.Open(filePath)
	if err != nil {
		log.Println("Error opening file:", err)
		//fmt.Println("Error opening file:", err)
		return
	}
	defer file.Close()

	_, err = c.CreateEditImage(
		ctx, openai.ImageEditRequest{
			Image:  file,
			Mask:   nil,
			Size:   openai.CreateImageSize1024x1024,
			N:      1,
			Prompt: "Please add India's flag at the background",
		})
	log.Println("Error executing request file:", err)
	checks.NoError(t, err, "CreateEditImage returned error")

PR #979, opened 2025-04-28
Fix for Edit Image content type requirement

### LINKED ISSUE #967: CreateEditImage method upload file report Supported file formats are 'image/png'
Your issue may already be reported!
Please search on the [issue tracker](https://github.com/sashabaranov/go-openai/issues) before creating one.

**Describe the bug**
CreateEditImage method upload file report Supported file formats are 'image/png'

**To Reproduce**
~~~go
config := openai.DefaultConfig("token")
config.HTTPClient = &http.Client{}
c := openai.NewClientWithConfig(config)
f4, _ := os.Open("./4.png")
defer f4.Close()
ctx := context.Background()
reqUrl := openai.ImageEditRequest{
	Image:          f4,
	Prompt:         "添加一只红色的鸭子",
	Model:          openai.CreateImageModelDallE2,
	Size:           openai.CreateImageSize512x512,
	ResponseFormat: openai.CreateImageResponseFormatURL,
	N:              1,
}
respUrl, err := c.CreateEditImage(ctx, reqUrl)
if err != nil {
	fmt.Printf("Image creation error: %v\n", err)
	return
}
fmt.Println(respUrl.Data[0].URL)
~~~

**Expected behavior**
Print out the URL address

**Screenshots/Logs**
Invalid file 'image': unsupported mimetype ('application/octet-stream'). Supported file formats are 'image/png'.

**Environment (please complete the following information):**
 - go-openai version: [e.g. v1.38.1]
 - Go version: [e.g. 1.20]
 - OpenAI API version: [e.g. v1]
 - OS: [e.g. Windows 10]

**Additional context**

Static-Flow:
**Describe the change**
PR to fix #967 and #601 The OpenAI Edit Image endpoint `/v1/images/edits` only accepts specific image formats as seen in error responses like so:
`status: 400 Bad Request, message: Invalid file 'image': unsupported mimetype ('image/jpeg'). Supported file formats are 'image/png'`. Even though the API docs say other types are supported [https://platform.openai.com/docs/api-reference/images/createEdit](https://platform.openai.com/docs/api-reference/images/createEdit):
```
For gpt-image-1, each image should be a png, webp, or jpg file less than 25MB.
```

This is not true as of `4-28-25` I think because it is still stuck on DALLE mode which only allows PNGs.

**Describe your solution**
Pretty simple fix really, replaced the call to the default helper method which only does `application/octet-stream` with the code the helper method calls and set the content-type to `image/png`

**Tests**
I ran a modified version of the image example which calls the edit image endpoint instead using my forked version and it worked successfully.

### DIFF
--- a/internal/form_builder.go
+++ b/internal/form_builder.go
@@ -4,8 +4,10 @@ import (
 	"fmt"
 	"io"
 	"mime/multipart"
+	"net/textproto"
 	"os"
 	"path"
+	"strings"
 )
 
 type FormBuilder interface {
@@ -34,12 +36,23 @@ func (fb *DefaultFormBuilder) CreateFormFileReader(fieldname string, r io.Reader
 	return fb.createFormFile(fieldname, r, path.Base(filename))
 }
 
+
+var quoteEscaper = strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
+
+func escapeQuotes(s string) string {
+	return quoteEscaper.Replace(s)
+}
+
 func (fb *DefaultFormBuilder) createFormFile(fieldname string, r io.Reader, filename string) error {
 	if filename == "" {
 		return fmt.Errorf("filename cannot be empty")
 	}
-
-	fieldWriter, err := fb.writer.CreateFormFile(fieldname, filename)
+	h := make(textproto.MIMEHeader)
+	h.Set("Content-Disposition",
+		fmt.Sprintf(`form-data; name="%s"; filename="%s"`,
+			escapeQuotes(fieldname), escapeQuotes(filename)))
+	h.Set("Content-Type", "image/png")
+	fieldWriter, err := fb.writer.CreatePart(h)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


Static-Flow:
@sashabaranov I'm not sure how to fix this codecov failing CI check. It says there's a 13% drop in testing coverage but nothing I changed materially impacted tests. It seems to think there's more changes than what my PR made, maybe due to `BASE commit is 91 commits behind HEAD on master`? Please advise, the edit image api's are currently broken without this fix.

internal/form_builder.go
```diff
@@ -34,12 +36,22 @@ func (fb *DefaultFormBuilder) CreateFormFileReader(fieldname string, r io.Reader
 	return fb.createFormFile(fieldname, r, path.Base(filename))
 }
 
+var quoteEscaper = strings.NewReplacer("\\", "\\\\", `"`, "\\\"")
```
sashabaranov, line 39: Might be better to use backtics here!

sashabaranov (commented):
Thank you for the PR! Could I kindly ask for an integration test here? Without them it's really hard for me to verify multitude of coming changes

sashabaranov:
@Static-Flow on the coverage note, could you try rebasing your branch on the latest master branch?

PR #980, opened 2025-04-30
add ChatTemplateKwargs to ChatCompletionRequest

justa-cai:
**Describe the change**
add chat_template_kwargs param support 

[vllm qwen3 thinking modes](https://qwen.readthedocs.io/en/latest/deployment/vllm.html#thinking-non-thinking-modes
)

`curl http://localhost:8000/v1/chat/completions -H "Content-Type: application/json" -d '{
  "model": "Qwen/Qwen3-8B",
  "messages": [
    {"role": "user", "content": "Give me a short introduction to large language models."}
  ],
  "temperature": 0.7,
  "top_p": 0.8,
  "top_k": 20,
  "max_tokens": 8192,
  "presence_penalty": 1.5,
  "chat_template_kwargs": {"enable_thinking": false}
}'`

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -275,6 +275,8 @@ type ChatCompletionRequest struct {
 	Metadata map[string]string `json:"metadata,omitempty"`
 	// Configuration for a predicted output.
 	Prediction *Prediction `json:"prediction,omitempty"`
+	// ExtraBody provides a way to add non-standard parameters to the request body. such as think mode for qwen3
+	ChatTemplateKwargs map[string]interface{} `json:"chat_template_kwargs,omitempty"`
 }
 
 type StreamOptions struct {


chat.go
```diff
@@ -275,6 +275,8 @@ type ChatCompletionRequest struct {
 	Metadata map[string]string `json:"metadata,omitempty"`
 	// Configuration for a predicted output.
 	Prediction *Prediction `json:"prediction,omitempty"`
+	// ExtraBody provides a way to add non-standard parameters to the request body. such as think mode for qwen3
+	ChatTemplateKwargs map[string]interface{} `json:"chat_template_kwargs,omitempty"`
```
sashabaranov, line 279:
```suggestion
	ChatTemplateKwargs map[string]any `json:"chat_template_kwargs,omitempty"`
```

torrischen:
@sashabaranov 
Hi, I wonder if the pr will be merged soon?
Our project is using this repo and currently we are using qwen3, we urgently need the feature to support switching between thinking and non-thinking

sashabaranov:
@justa-cai @torrischen there are still a few pending comments and I'd also add that we should explicitly name the variable so it would be obvious that the parameter is not related to OpenAI first-party API. I'd be happy to merge after all this is fixed 🫶

justa-cai:
#ChatTemplateKwargs
Additional kwargs to pass to the template renderer. Will be accessible by the chat template.
[VLLM](https://github.com/vllm-project/vllm/blob/05a4324f8e3932c25554791ff248e3e0200eef92/vllm/entrypoints/openai/protocol.py#L320)

justa-cai:
@sashabaranov help meger

sashabaranov (approved):
Thank you for update!

justa-cai:
> `enable_thinking` received by qwen3 model should be at the same level as `messages` parameter. Here, we need to take out `enable_thinking` passed by `chat_template_kwargs` parameter and put it at the same level as `messages` parameter.

It works fine on the Qwen3-30B-A3B model using vLLM.

Issue #1002, opened 2025-05-16
ChatCompletionResponse.Created r string type


dragonTalon:
I found that the format created returned by some models is of type String
So i think created  must support compatibility between int64 and string formats 
```
{
    "id": "",
    "created": "2025/05/16 16:29:57",
    "model": null,
    "object": "chat.completion",
    "system_fingerprint": null,
    "choices": [
        {
            "finish_reason": "stop",
            "index": 0,
            "message": {
                "content": "....",
                "role": "assistant",
                "tool_calls": null,
                "function_call": null
            }
        }
    ],
    "usage": {
  
    },
    "service_tier": null,
    "code": 0,
    "error_message": ""
}
```

PR #1003, opened 2025-05-16
update ChatCompletionResponse.Created filed ,support  return Created type is string

dragonTalon:
ChatCompletionResponse.Created filed support compatibility between int64 and string formats. and created is null

```
Created             interface           `json:"created,omitempty"`

```

Issue: #1002

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -407,7 +407,6 @@ func (ft *FlexibleTime) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &timeStr); err != nil {
 		return fmt.Errorf("created type need number or string")
 	}
-
 	return ft.parseWithLayouts([]string{
 		"2006/01/02 15:04:05",
 		time.RFC3339,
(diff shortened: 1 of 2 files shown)


sashabaranov:
@dragonTalon thank you for the PR! OpenAI docs specify that the type of this object is integer here https://platform.openai.com/docs/api-reference/chat/object#chat/object-created

Are you using a non-OpenAI api by any chance?

stmcginnis:
OpenAI also publishes [an official schema now](https://github.com/openai/openai-openapi/blob/498c71ddf6f1c45b983f972ccabca795da211a3e/openapi.yaml#L19722-L19725). This should remain an integer value, not a string.

stmcginnis:
> we believe that discrepancies in the 'created' field should not hinder the usage of go-openai

The right answer here is that whatever software you are using to host that model should be fixed to return the correct data type. It should not be up to this library to deviate from the official schema to work around a bug in one implementation.

Issue #1012, opened 2025-06-03
CreateEditImage with gpt-image-1 always returns missing_file_mimetype


Icety:
**Describe the bug**
When trying to use the gpt-image-1 model with the imageEdit feature to edit a local png, I always get the error:
`missing_file_mimetype`. I tried streaming, fetching as http and passing the body, but I always get the same error.

**To Reproduce**
```
file, _ := os.Open("./example.png")
		defer file.Close()

		resp, err = openai.NewClient(key).CreateEditImage(
			context.Background(),
			openai.ImageEditRequest{
				Image:          file,
				Model:          openai.CreateImageModelGptImage1,
				Prompt:         prompt,
				Quality:        openai.CreateImageQualityLow,
				N:              1,
				Size:           openai.CreateImageSize1024x1024,
				ResponseFormat: openai.CreateImageOutputFormatPNG,
			},
		)
```

**Expected behavior**
I expect an image to be created based on the provided example image.

**Screenshots/Logs**
Error code: missing_file_mimetype
Error message: Invalid file 'image': failed to determine mimetype. Ensure you are providing the file's mimetype as part of the 'Content-Type' header.

**Environment (please complete the following information):**
 - go-openai version: 1.40.1
 - Go version: 1.22.4
 - OpenAI API version: v1
 - OS: Mac Sequoia

Issue #1010, opened 2025-05-30
V1.40.1 Regression


cthackers:
When adding a file as a io.Reader to the transcription request, the 1.40.0 `CreateFormFileReader` used to call `createFormFile` which did the right thing, but now the in 1.40.1 was changed to create the form part differently and misses the Content-Type header so it's not accepted as a valid file. Can you please revert or fix this.

Thank you

cthackers:
I know there is a reader there and you can't read from it, the file name is just "." so won't help, but you can do something like this

```go
// CreateFormFileReader creates a multipart form part from an io.Reader with detected Content-Type.
func (fb *DefaultFormBuilder) CreateFormFileReader(fieldname string, r io.Reader, filename string) error {
	// Buffer to capture the first 512 bytes
	var buf bytes.Buffer
	tee := io.TeeReader(r, &buf)

	// Read up to 512 bytes for Content-Type detection
	data := make([]byte, 512)
	n, err := tee.Read(data)
	if err != nil && err != io.EOF {
		return fmt.Errorf("failed to read from reader for Content-Type detection: %w", err)
	}

	// Detect Content-Type
	contentType := http.DetectContentType(data[:n])

	// Create a new reader that includes the read bytes plus the rest
	combinedReader := io.MultiReader(&buf, r)

	// Set up MIME headers
	h := make(textproto.MIMEHeader)
	h.Set(
		"Content-Disposition",
		fmt.Sprintf(
			`form-data; name="%s"; filename="%s"`,
			escapeQuotes(fieldname),
			escapeQuotes(filepath.Base(filename)),
		),
	)
	h.Set("Content-Type", contentType)

	// Create the multipart part
	fieldWriter, err := fb.writer.CreatePart(h)
	if err != nil {
		return fmt.Errorf("failed to create form part: %w", err)
	}

	// Copy the full content to the form part
	_, err = io.Copy(fieldWriter, combinedReader)
	if err != nil {
		return fmt.Errorf("failed to copy reader to form part: %w", err)
	}

	return nil
}

PR #1017, opened 2025-06-10
fix image api missing filename bug

Axb12:
Fixed the bug that the filename was missing due to the last PR.(https://github.com/sashabaranov/go-openai/pull/994)
I've added a file struct to override what was going on before the last PR.(https://github.com/sashabaranov/go-openai/pull/994)
I'm so sorry because of my negligence that caused this issue.
Here are two related issues I've seen:(https://github.com/sashabaranov/go-openai/issues/1012), (https://github.com/sashabaranov/go-openai/issues/1010)

### DIFF
--- a/internal/form_builder_test.go
+++ b/internal/form_builder_test.go
@@ -1,6 +1,8 @@
 package openai //nolint:testpackage // testing private field
 
 import (
+	"io"
+
 	"github.com/sashabaranov/go-openai/internal/test/checks"
 
 	"bytes"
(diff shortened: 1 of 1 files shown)


internal/form_builder.go
```diff
@@ -39,9 +39,18 @@ func escapeQuotes(s string) string {
...
-// The filename in parameters can be an empty string.
-// The filename in Content-Disposition is required, But it can be an empty string.
+// The filename in Content-Disposition is required.
 func (fb *DefaultFormBuilder) CreateFormFileReader(fieldname string, r io.Reader, filename string) error {
+	if filename == "" {
+		if f, ok := r.(interface{ Name() string }); ok {
+			filename = f.Name()
+		}
+	}
+	var contentType string
```
Axb12, line 49: I'm not sure if the default Content-Type is set to application/octet-stream here

sashabaranov (approved):
thank you for the fix!

Issue #1021, opened 2025-06-17
stream_options.include_usage: false still returns usage in streamed chunks


Hritik003:
**Describe the bug**
When using the streaming API with `stream: true` and `stream_options.include_usage: false`, the streamed responses still include the `"usage"` field in each data chunk.
This is inconsistent with the documented or expected behavior — setting `include_usage: false` should suppress the `"usage"` field in the streamed response chunks.

---

**To Reproduce**
Use the `/v1/completions` endpoint with the following payload (or equivalent in `go-openai`):

```json
{
  "model": "gemma-2",
  "stream": true,
  "stream_options": {
    "include_usage": false
  },
  "prompt": "Write something interesting."
}
```

Observe that each `data:` chunk still includes:

```json
"usage": {
  "prompt_tokens": 0,
  "completion_tokens": 0,
  "total_tokens": 0,
  "prompt_tokens_details": null,
  "completion_tokens_details": null
}
```

Also tested with `stream_options.include_usage: true`, and both cases show `"usage"` in stream chunks.

---

**Expected behavior**
If `stream_options.include_usage` is set to `false`, the streamed data chunks should omit the `"usage"` field entirely or return it as `null`. Including usage despite explicitly disabling it violates the expected contract of the API.

---

**Screenshots/Logs**
Example log:

```json
data: {"id":"cmpl-a16ec15821d444e8a22372900e99cc7e","object":"text_completion","created":1750151236,"model":"gemma-2","choices":[{"text":"writing","index":0,"finish_reason":""}],"usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}
```

---

**Environment (please complete the following information):**

* `go-openai` version: `v1.40.1`
* Go version: `1.21`
* OpenAI API version: `v1`
* OS: macOS

---

**Additional context**
The bug is confusing for consumers trying to reduce response payloads or suppress token accounting per stream chunk. This appears to be a server-side issue or a client-side mishandling of response parsing.

---

PR #1022, opened 2025-06-17
Fix for removing usage in every stream chunk response.

Hritik003:
### **Describe the change**

This PR fixes an issue where the `usage` field is always present in streamed completion responses, even when `stream_options.include_usage` is explicitly set to `false`.

By changing the `Usage` field in `CompletionResponse` to a pointer and tagging it with `omitempty`, we ensure that if `usage` is not included in the actual API response, it will be omitted from the parsed Go struct as well.

This improves correctness and reduces confusion when users want to opt out of usage metadata in streaming mode.

---

### **Provide OpenAI documentation link**

API Reference: [https://platform.openai.com/docs/api-reference/completions/create](https://platform.openai.com/docs/api-reference/completions/create)

---

### **Describe your solution**

Modified the `CompletionResponse` struct to update the `Usage` field:

```go
Usage *Usage `json:"usage,omitempty"`
```

This allows Go’s `encoding/json` to omit the field when it's not present in the streamed response. This makes the client-side response cleaner and honors the behavior expected when `include_usage` is set to `false`.

No new libraries or dependencies were added.

---

### **Tests**

* Manually tested by calling `/v1/completions` with both `stream_options.include_usage: true` and `false`
* Verified that:

  * When `include_usage: true`, the field is present.
  * When `include_usage: false`, the field is omitted in streamed chunks.
* Updated integration tests to assert presence/absence of `Usage` field conditionally based on stream option.

---

### **Additional context**

This PR improves alignment with OpenAI’s API behavior and helps consumers avoid parsing unnecessary fields.

Issue: #1021

### DIFF
--- a/completion.go
+++ b/completion.go
@@ -242,7 +242,7 @@ type CompletionResponse struct {
 	Created int64              `json:"created"`
 	Model   string             `json:"model"`
 	Choices []CompletionChoice `json:"choices"`
-	Usage   *Usage             `json:"usage"`
+	Usage   *Usage             `json:"usage,omitempty"`
 
 	httpHeader
 }
--- a/completion.go
+++ b/completion.go
@@ -242,7 +242,7 @@ type CompletionResponse struct {
 	Created int64              `json:"created"`
 	Model   string             `json:"model"`
 	Choices []CompletionChoice `json:"choices"`
-	Usage   Usage              `json:"usage"`
+	Usage   *Usage             `json:"usage"`
 
 	httpHeader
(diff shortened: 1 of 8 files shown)


Hritik003:
@sashabaranov can you please look into this?

aishwaryaraimule21 (approved):

AyushSawant18588 (approved):

Hritik003:
@sashabaranov A reminder to look into this?

Issue #1027, opened 2025-07-02
Cannot bind response_format.json_schema.schema into json.Marshaler field for structured outputs in chat/completions


AyushSawant18588:
I’m building a service-to-service integration that receives Chat Completion requests and binds them directly into the Go client’s ChatCompletionRequest struct using Gin:
```
var req openai.ChatCompletionRequest
if err := c.ShouldBindJSON(&req); err != nil {
    // handle binding error
}
```

The struct for the JSON Schema portion is: https://github.com/sashabaranov/go-openai/blob/c125ae2ad7b239355161c7f4260a3743ec0c182b/chat.go#L220

Because json.Marshaler only implements MarshalJSON() and not UnmarshalJSON(), any nested object under "schema" fails to bind.

**Reproduction**
Send an HTTP POST with a body like:
```
{
  "model": "llama3-1b",
  "messages": [
    { "role": "system", "content": "You are a helpful math tutor." },
    { "role": "user",   "content": "solve 8x + 31 = 2" }
  ],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name":   "math_response",
      "strict": true,
      "schema": {
        "type": "object",
        "properties": {
          "steps": {
            "type": "array",
            "items": {
              "type": "object",
              "properties": {
                "explanation": { "type": "string" },
                "output":      { "type": "string" }
              },
              "required": ["explanation","output"],
              "additionalProperties": false
            }
          },
          "final_answer": { "type": "string" }
        },
        "required": ["steps","final_answer"],
        "additionalProperties": false
      }
    }
  }
}

```

**Actual error:**
```
json: cannot unmarshal object into Go struct field 
ChatCompletionResponseFormatJSONSchema.response_format.json_schema.schema of type json.Marshaler
```

**Expected Behavior:**
The request should bind successfully into openai.ChatCompletionRequest without modifying the client library.
I can include a full JSON Schema object under "schema" and have it round-trip through Gin’s ShouldBindJSON into the existing json.Marshaler field.

**Question**
1. Is there a valid JSON payload I can send right now—without changing the client’s struct definition—that will satisfy ShouldBindJSON and allow my handler to bind into ChatCompletionResponseFormatJSONSchema.Schema? If so, please provide an example.

2. Else, Can we update the ChatCompletionResponseFormatJSONSchema struct (in configuration or code) so that this bind is successful?

Thank you!

PR #1028, opened 2025-07-07
fix: properly unmarshal JSON schema in ChatCompletionResponseFormatJSONSchema.schema

eiixy:
When handling the response format for ChatCompletionResponseFormatJSONSchema, the field
response_format.json_schema.schema fails to unmarshal properly

Issue: #1027

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -214,11 +214,15 @@ type ChatCompletionResponseFormat struct {
 	JSONSchema *ChatCompletionResponseFormatJSONSchema `json:"json_schema,omitempty"`
 }
 
+type JSONSchema interface {
+	json.Marshaler
+	json.Unmarshaler
+}
 type ChatCompletionResponseFormatJSONSchema struct {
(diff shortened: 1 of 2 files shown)


AyushSawant18588:
@eiixy thanks for opening a PR for this. I tested these change but I am still seeing the following error :
```
json: cannot unmarshal object into Go struct field ChatCompletionResponseFormatJSONSchema.response_format.json_schema.schema of type openai.JSONSchema
```

I think issue stems from the Schema field in ChatCompletionResponseFormatJSONSchema being defined as an interface JSONSchema.
Go's JSON unmarshaler cannot instantiate interface types directly, it needs to know what concrete type to create. When it encounters the schema field, it doesn't know whether to create a jsonschema.Definition or another type that implements JSONSchema.

Are you also seeing this error or am I missing something?

sashabaranov (approved):
Thank you for the PR!

PR #1033, opened 2025-07-10
Removed root $ref from GenerateSchemaForType

eiixy:
#1030

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+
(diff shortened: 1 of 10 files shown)


sashabaranov:
@eiixy please lmk if copilot suggestions are legit, otherwise lgtm!

eiixy:
> @eiixy please lmk if copilot suggestions are legit, otherwise lgtm!

The suggestion about field order isn’t an issue — we convert the structToMap before comparing via reflect.DeepEqual, so the order doesn’t affect the result.

sashabaranov (approved):

PR #1034, opened 2025-07-10
Add GuidedChoice to ChatCompletionRequest

AyushSawant18588:
**Describe the change**
Adds guided_choice param support to chat completions

[Structured output vLLM support](https://docs.vllm.ai/en/latest/features/structured_outputs.html#online-serving-openai-api)

```
curl --request post \
  --url http://localhost:8000/v1/chat/completions \
  --header 'Content-Type: application/json' \
  --data '{
    "model": "llama3-3b",
    "messages": [
        {
            "role": "user",
            "content": "Classify this sentiment: vLLM is great!"
        }
    ],
    "guided_choice": [
        "negative",
        "positive"
    ]
}'
```

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -248,7 +248,7 @@ func (r *ChatCompletionResponseFormatJSONSchema) UnmarshalJSON(data []byte) erro
 	return nil
 }
 
-// NonOpenAIExtensions contains non-standard OpenAI API extensions
+// NonOpenAIExtensions contains non-standard OpenAI API extensions.
 type NonOpenAIExtensions struct {
 	// GuidedChoice restricts output to a set of predefined choices.
 	GuidedChoice []string `json:"guided_choice,omitempty"`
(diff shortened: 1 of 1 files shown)


AyushSawant18588:
@sashabaranov most of the fields we use are on top level and since this field is widely used for the use case of structured outputs can we merge this PR? 
I will raise a separate PR to include extra body in chat completions request

sashabaranov:
@AyushSawant18588 we can make it `.NonOpenAIExtensions.GuidedChoice` to preserve type-safety

AyushSawant18588:
@sashabaranov do these changes look good?

sashabaranov:
@AyushSawant18588 thank you for the update, this seems like a pretty elegant solution!

Here are a few things to consider:
1. Copilot point about `,inline` not needed above
2. The `NonOpenAIExtensions` is a very generic name that could apply to pretty much anything in the codebase, in that context it should be `ChatCompletionRequestNonOpenAIExtensions` it seems

AyushSawant18588:
@sashabaranov Can you check the changes? I renamed struct to ChatCompletionRequestExtensions cause name was getting too big but let me know if it has to be changed

sashabaranov:
Thank you!!

PR #1048, opened 2025-07-17
Support Google VertexAI Gemini-specific parameters

nemo1105:
**Describe the change**
Support Google VertexAI Gemini-specific parameters. 

https://cloud.google.com/vertex-ai/generative-ai/docs/migrate/openai/overview#gemini-specific_parameters

**Describe your solution**
Just add two fields to `ChatCompletionRequest` struct

**Tests**
I did not add any tests.

**Additional context**
nothing

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -311,8 +311,8 @@ type ChatCompletionRequest struct {
 	ServiceTier ServiceTier `json:"service_tier,omitempty"`
 	// Support Google VertexAI Gemini-specific parameters
 	// https://cloud.google.com/vertex-ai/generative-ai/docs/migrate/openai/overview#extra_body_features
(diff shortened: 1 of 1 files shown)


sashabaranov:
@wclssdn thank you for the PR!

I suggest we go with typed approach similar to https://github.com/sashabaranov/go-openai/pull/1034 instead of `map[string]any`.

wclssdn:
> @wclssdn thank you for the PR!
> 
> I suggest we go with typed approach similar to #1034 instead of `map[string]any`.

Currently, the vast majority of model providers support OpenAI's API standards, while also offering some personalized capabilities. How to better support these capabilities is a relatively challenging issue. My suggestion is to use generics, with each manufacturer having a generic support. However, this workload is too heavy for me~ I hope someone can implement this~

PR #1064, opened 2025-08-11
Add Verbosity parameter to Chat Completion Request

azdaev:
## Add Verbosity parameter to Chat Completion Request

Verbosity - Constrains the verbosity of the model's response. Lower values will result in more concise responses, while higher values will result in more verbose responses. Currently supported values are low, medium, and high.

OpenAI Docs - https://platform.openai.com/docs/api-reference/chat/create#chat_create-verbosity

## Testing
✅ All existing tests pass

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -320,6 +320,10 @@ type ChatCompletionRequest struct {
 	ChatTemplateKwargs map[string]any `json:"chat_template_kwargs,omitempty"`
 	// Specifies the latency tier to use for processing the request.
 	ServiceTier ServiceTier `json:"service_tier,omitempty"`
+	// Verbosity determines how many output tokens are generated.
+	// Lowering the number of tokens reduces overall latency.
+	// It can be set to "low", "medium", or "high".
+	Verbosity string `json:"verbosity,omitempty"`
 	// Embedded struct for non-OpenAI extensions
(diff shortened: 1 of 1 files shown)


sashabaranov:
@azdaev got it, thanks! Could we please add a note about that in the comment for this field?

azdaev:
@sashabaranov no problem! done! feel free to edit the noe if it doesn't satisfy you

aleiva17:
> Hey @azdaev, thank you for the PR! I don't see the `verbosity` in the OpenAI docs link you've provided, and if I search for it I get to this page: https://platform.openai.com/docs/guides/latest-model#verbosity
> 
> This page says that it's supported in the `client.responses.create` API, which we haven't implemented in this library yet.
> 
> Please lmk if this parameter is supported in the chat completions API and is not documented.
> 
> Also, is this option only available for GPT-5?

It's not being displayed in that page, but if you install the official openai go sdk (v2), it's available as part of the completions API:

```
client := openai.NewClient()

resp, err := client.Chat.Completions.New(ctx,
	openai.ChatCompletionNewParams{
		Model: openai.ChatModelGPT5,
		// ...
		Verbosity: openai.ChatCompletionNewParamsVerbosityHigh,
	},
)
```

PR #1066, opened 2025-08-12
feat: add safety_identifier params

liushuangls:
support safety_identifier params
docs: https://platform.openai.com/docs/api-reference/chat/create#chat_create-safety_identifier


sashabaranov (approved):
Thank you!

Issue #1025, opened 2025-06-24
add extra_body


Tongsir-China:
model like doubao has a new model seed-1.6. 
it has a paramemter named "extra_body"
https://www.volcengine.com/docs/82379/1449737
![Image](https://github.com/user-attachments/assets/f6ff357c-85e9-447b-a657-fcfe2263e2af)

SteveHere:
#906 addresses this issue, but is still pending review from @sashabaranov

tumingke0:
Please approve the review as soon as possible, I urgently need to use it.

I321065:
this feature is vary important, please handle it asap

PR #1069, opened 2025-08-21
Support extra_body parameters

ArtdragonXoX:
**Describe the change**
add extra_body param support in ChatCompletionRequestExtensions

https://ai.google.dev/gemini-api/docs/openai#thinking

```
curl "https://generativelanguage.googleapis.com/v1beta/openai/chat/completions" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer GEMINI_API_KEY" \
-d '{
    "model": "gemini-2.5-flash",
      "messages": [{"role": "user", "content": "Explain to me how AI works"}],
      "extra_body": {
        "google": {
           "thinking_config": {
             "include_thoughts": true
           }
        }
      }
    }'
```

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -257,6 +257,11 @@ type ChatCompletionRequestExtensions struct {
 	// ensuring predictable and consistent outputs in scenarios where specific
 	// choices are required.
 	GuidedChoice []string `json:"guided_choice,omitempty"`
+	// ExtraBody provides configuration options for the generation process in Gemini API.
+	// Additional configuration parameters to control model behavior. Will be passed directly to the Gemini API.
+	// Such as thinking mode for Gemini. "extra_body": {"google": {"thinking_config": {"include_thoughts": true}}}
+	// https://ai.google.dev/gemini-api/docs/openai
+	ExtraBody map[string]any `json:"extra_body,omitempty"`
 }
 
 // ChatCompletionRequest represents a request structure for chat completion API.
--- a/chat.go
+++ b/chat.go
@@ -487,6 +487,7 @@ func (c *Client) CreateChatCompletion(
 		http.MethodPost,
 		c.fullURL(urlSuffix, withModel(request.Model)),
 		withBody(request),
+		withExtraBody(request.ExtraBody),
 	)
 	if err != nil {
 		return
--- a/chat.go
+++ b/chat.go
@@ -257,11 +257,6 @@ type ChatCompletionRequestExtensions struct {
 	// ensuring predictable and consistent outputs in scenarios where specific
 	// choices are required.
 	GuidedChoice []string `json:"guided_choice,omitempty"`
-	// ExtraBody provides configuration options for the generation process in Gemini API.
-	// Additional configuration parameters to control model behavior. Will be passed directly to the Gemini API.
-	// Such as thinking mode for Gemini. "extra_body": {"google": {"thinking_config": {"include_thoughts": true}}}
-	// https://ai.google.dev/gemini-api/docs/openai
-	ExtraBody map[string]any `json:"extra_body,omitempty"`
 }
 
 // ChatCompletionRequest represents a request structure for chat completion API.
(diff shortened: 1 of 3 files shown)


zhouyuchuan00:
anyone can review？I also need this pr

ArtdragonXoX:
I moved the ExtraBody field from ChatCompletionRequestExtensions into the ChatCompletionRequest struct.

ArtdragonXoX:
> This should be similar to #1025, but the extra_body should be added to the root node of the JSON like #906

Thanks for the tip, I have embedded extra_body into the root node of the JSON.

ArtdragonXoX:
@sashabaranov Can you check this pr please?

chat.go
```diff
@@ -477,11 +482,28 @@ func (c *Client) CreateChatCompletion(
...
+	var body map[string]any
+	_ = json.Unmarshal(jsonData, &body)
+
 	req, err := c.newRequest(
 		ctx,
 		http.MethodPost,
 		c.fullURL(urlSuffix, withModel(request.Model)),
-		withBody(request),
+		withBody(body),           // Main request body.
+		withExtraBody(extraBody), // Merge ExtraBody fields.
```
sashabaranov, line 506: I don't think you are using this correctly: `withExtraBody` merges `extraBody` into request body. However, in the example you've mentioned it goes as a separate field:

```
curl "https://generativelanguage.googleapis.com/v1beta/openai/chat/completions" \
-H "Content-Type: application/json" \
-H "Authorization: Bearer GEMINI_API_KEY" \
-d '{
    "model": "gemini-2.5-flash",
      "messages": [{"role": "user", "content": "Explain to me how AI works"}],
      "extra_body": {
        "google": {
           "thinking_config": {
             "include_thoughts": true
           }
        }
      }
    }'
```

ArtdragonXoX: Thank you for your inspection. I have seen it working correctly during testing and use. For example:
```
_, err := client.CreateChatCompletion(context.Background(), openai.ChatCompletionRequest{
	Model: "doubao-seed-1-6-250615",
		{
			Content: "Hello!",
		},
	},
	ExtraBody: map[string]any{
		"thinking":map[string]any {
			"type": "disabled",
		},
	},
})
```
After the request is processed as the body, it will change to:
```
{
    "model": "doubao-seed-1-6-250615",
    "messages": [{"role": "user", "content": "Hello!"}],
    "thinking": {
        "type": "disabled"
    }
}
```
For the gemini example, I'm sorry that I didn't give a clear example that caused your misunderstanding. In Gemini's official API, they have additional support for extra_body fields. Let me explain the example call in the documentation:
```
from openai import OpenAI

client = OpenAI(
    api_key="GEMINI_API_KEY",
    base_url="https://generativelanguage.googleapis.com/v1beta/openai/"
)

response = client.chat.completions.create(
    model="gemini-2.5-flash",
    messages=[{"role": "user", "content": "Explain to me how AI works"}],
    extra_body={
      'extra_body': {
        "google": {
          "thinking_config": {
            "thinking_budget": 800,
            "include_thoughts": True
          }
        }
      }
    }
)
```
They use OpenAI's official python library OpenAI. In OpenAI extra_body will be promoted to root. The final body is:
```
{
    "messages": [
        {
            "role": "user",
            "content": "Explain to me how AI works"
        }
    ],
    "model": "gemini-2.5-flash",
    "extra_body": {
        "google": {
            "thinking_config": {
                "thinking_budget": 800,
                "include_thoughts": true
            }
        }
    }
}
```
So Gemini's example is actually to ensure that extra_body is still under extra_body after being extracted to root, so there is an extra layer of extra_body.
Like the example in https://github.com/sashabaranov/go-openai/issues/1025, the requirement is to extract the contents of extra_body to the root.
I put a sample code here, which can print out the processed body, you can test it:
```
import httpx
from openai import OpenAI


def log_request(request):
    print("\n===== HTTP Request Details ======")
    print(f"Method: {request.method}")
    print(f"URL: {request.url}")
    print("Headers:")
    for key, value in request.headers.items():
        print(f"  {key}: {value}")
    print("Body:")
    print(request.content.decode())
    print("================================\n")


client = OpenAI(
    api_key="GEMINI_API_KEY",
    base_url="https://generativelanguage.googleapis.com/v1beta/openai/",
    http_client=httpx.Client(event_hooks={"request": [log_request]}),
)

response = client.chat.completions.create(
    model="gemini-2.5-flash",
    messages=[{"role": "user", "content": "Explain to me how AI works"}],
    extra_body={
        'extra_body': {
          "google": {
              "thinking_config": {
                  "thinking_budget": 800,
                  "include_thoughts": True
              }
          }
        }
    }
)

print(response)
```

xenv:
@sashabaranov This MR looks solid. What's the timeline for merging it?

surxme:
need this too

xuges:
I need this PR too, why not merge?

liushuangls:
@sashabaranov This PR looks very useful. Given that it also benefits other AI platforms that are compatible with the OpenAI API, can we continue pushing to get it merged?

sashabaranov:
there are still standing issues from codex review, I'm happy to merge once they'll be fixed

PR #1095, opened 2025-12-25
feat: Add MIMEType field to ChatMessageImageURL

kangkangkk:
1. Added MIMEType field to ChatMessageImageURL struct
2. allowing users to specify the MIME type of image URLs in multi-content messages.

3. relevant API doc 
https://platform.openai.com/docs/api-reference/uploads/create

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -77,8 +77,9 @@ const (
 )
 
 type ChatMessageImageURL struct {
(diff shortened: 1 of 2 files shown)


kangkangkk:
hello @sashabaranov , this is ready for review whenever you have a moment.
Let me know if you need any additional context.

PR #1106, opened 2026-04-27
fix: detect Content-Type from file extension in multipart uploads

blackwell-systems:
**Describe the change**

`CreateFormFileReader` only sets the `Content-Type` multipart header when the reader implements a `ContentType() string` interface. A plain `*os.File` does not implement this, so the header is omitted. The `gpt-image-1` model requires `Content-Type` and rejects requests with `missing_file_mimetype` when it is absent.

https://platform.openai.com/docs/api-reference/images/createEdit

**Describe your solution**

When the reader does not implement `ContentType()`, fall back to `mime.TypeByExtension(filepath.Ext(filename))` to detect the content type from the file extension. This covers common image formats (.png, .jpg, .webp, .gif) and any other extension registered in the system's MIME database.

**Tests**

Added `TestCreateFormFileReaderDetectsMimeFromExtension` verifying that a plain reader with a `.png` filename gets `Content-Type: image/png` in the multipart header.

All tests pass: `go test ./...`

Issue: #1012

### DIFF
--- a/internal/form_builder.go
+++ b/internal/form_builder.go
@@ -3,6 +3,7 @@ package openai
 import (
 	"fmt"
 	"io"
+	"mime"
 	"mime/multipart"
 	"net/textproto"
 	"os"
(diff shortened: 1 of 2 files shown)


sashabaranov:
@blackwell-systems sorry this was merged with another PR, thank you for your contribution though!

PR #1109, opened 2026-06-04
Document OpenAI-compatible base URL configuration

cerebrixos:
This keeps the default OpenAI endpoint unchanged while documenting how to configure an OpenAI-compatible endpoint through OPENAI_BASE_URL.\n\nWhy this may be useful:\n- local inference servers\n- private model gateways\n- self-hosted routers\n- observability and enterprise control planes\n\nThe example uses the existing ClientConfig.BaseURL field and does not add a provider-specific integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -440,6 +440,23 @@ See also: https://pkg.go.dev/github.com/sashabaranov/go-openai#ClientConfig
 </details>
 
 <details>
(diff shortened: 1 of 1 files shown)


sashabaranov:
Thank you for the PR, the current README has a similar-ish clause

PR #1111, opened 2026-07-23
fix: require non-empty model for chat completions

sonnemusk:
## What

`CreateChatCompletion` and `CreateChatCompletionStream` return `ErrChatCompletionRequestEmptyModel` when `Model` is empty.

## Why

The API rejects empty model names. Failing early saves a round trip and gives a clear, typed error.

## Test

- `TestCreateChatCompletionEmptyModel`


sonnemusk:
Closing this PR as low-signal noise after a quality pass on open contributions. Sorry for the distraction — focusing on fewer, higher-impact fixes going forward.

PR #1116, opened 2026-08-14
feat: add ChunkingStrategy to AudioRequest for diarization models

### LINKED ISSUE #1101: `AudioRequest` missing `chunking_strategy` field
**Is your feature request related to a problem? Please describe.**

I'm using `gpt-4o-transcribe-diarize` for speaker diarization. It works fine for short clips, but anything bigger fails with:

```
chunking_strategy is required for diarization models: bad request
```

Looking at the [transcription API docs](https://platform.openai.com/docs/api-reference/audio/createTranscription), `chunking_strategy` is required for diarization on longer audio. But `AudioRequest` doesn't have this field, so there's no way to set it through the library.

**Describe the solution you'd like**

Add `ChunkingStrategy` field to `AudioRequest` struct.

Also, `AudioResponseFormatDiarizedJSON` for the `diarized_json` format would be useful.

**Describe alternatives you've considered**

Making direct HTTP requests to work around it, which defeats the purpose of using the library.

serhiizghama:
Closes #1101. Transcription with a diarization model (`gpt-4o-transcribe-diarize`) fails on anything longer than a short clip with `chunking_strategy is required for diarization models: bad request`, and there was no way to set that parameter through `AudioRequest`.

Added a `ChunkingStrategy` field. It takes the string `"auto"` for automatic chunking, or a `TranscriptionChunkingStrategy` value when you want explicit server_vad control — the latter gets JSON-encoded into the multipart form, a plain string is sent as-is. When unset nothing is written, so existing requests are unchanged.

While adding the form-writing I pulled the timestamp-granularities loop out into its own helper alongside the new one; the multipart builder was getting long.

Tests cover both the `"auto"` and server_vad-object cases (asserting the exact form value) plus the unset case, and the field is added to the existing failing-form-builder coverage.

### DIFF
--- a/audio.go
+++ b/audio.go
@@ -3,6 +3,7 @@ package openai
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
@@ -38,6 +39,16 @@ const (
 	TranscriptionTimestampGranularitySegment TranscriptionTimestampGranularity = "segment"
 )
 
+// TranscriptionChunkingStrategy configures server-side VAD chunking for transcription.
+// Assign it to AudioRequest.ChunkingStrategy when you need explicit control; pass the
+// string "auto" instead to let the server pick the boundaries.
+type TranscriptionChunkingStrategy struct {
+	Type              string  `json:"type"` // "server_vad"
+	PrefixPaddingMs   int     `json:"prefix_padding_ms,omitempty"`
+	SilenceDurationMs int     `json:"silence_duration_ms,omitempty"`
+	Threshold         float64 `json:"threshold,omitempty"`
+}
+
 // AudioRequest represents a request structure for audio API.
 type AudioRequest struct {
 	Model string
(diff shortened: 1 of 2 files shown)


sashabaranov (approved):
@serhiizghama Thank you for the PR! Looks good to me. Quick question: should we default to chunking strategy `auto` by default?

serhiizghama:
Thanks! I'd lean toward leaving it unset by default. Right now nil means the field is omitted and the server applies its own default, which matches how the other optional `AudioRequest` fields behave (`Language`, `TimestampGranularities`, etc.). Defaulting to `auto` in the client would send `chunking_strategy` on every transcription call — including `whisper-1`/`gpt-4o-transcribe` where it isn't needed — and would quietly change behavior for existing callers on upgrade.

Since it's really only required for the diarization models, keeping it opt-in (with the doc comment pointing that out) felt like the least-surprising option. Happy to switch to an `auto` default if you'd rather have it, though.

PR #1117, opened 2026-08-18
fix: preserve json_schema verbatim when unmarshaling response_format

serhiizghama:
Unmarshaling a `response_format` with a `json_schema` currently decodes the schema into `jsonschema.Definition`, which only models a subset of JSON Schema. Any keyword it has no field for — `anyOf`, `title`, `$schema`, `oneOf`, `allOf`, `format`, and so on — gets silently dropped, so the schema comes back corrupted. A nullable field written as `{"anyOf":[{"type":"string"},{"type":"null"}]}` round-trips down to `{}`, and titles disappear entirely.

Kept the raw schema instead. `Schema` is typed `json.Marshaler` and `json.RawMessage` implements it, so storing the verbatim bytes preserves the schema exactly without changing the field type. It still checks the schema is a JSON object, so the existing rejection of non-object schemas (a bare number, an array) is unchanged. Added a round-trip test where a schema with `anyOf`/`title` survives intact.

### DIFF
--- a/chat.go
+++ b/chat.go
@@ -5,8 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
-
-	"github.com/sashabaranov/go-openai/jsonschema"
 )
 
 // Chat message role defined by the OpenAI API.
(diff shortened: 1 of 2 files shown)


sashabaranov (approved):
Great work, thanks!

PR #1118, opened 2026-08-26
fix: don't append api-version query param for Anthropic clients

serhiizghama:
Building a client with `DefaultAnthropicConfig` sends every request to `https://api.anthropic.com/v1/chat/completions?api-version=2023-06-01`. That `APIVersion` value is only meant for the `anthropic-version` header, which `setCommonHeaders` already sets — but `fullURL` also appends it as an `api-version` query param whenever it's non-empty, which is an Azure convention that Anthropic got swept into once it started reusing the `APIVersion` field.

Added `&& c.config.APIType != APITypeAnthropic` to the guard so Azure/AzureAD/CloudflareAzure keep the query param they need and Anthropic clients get a clean URL. Added a `fullURL` test for the Anthropic case; existing tests still pass.

### DIFF
--- a/api_internal_test.go
+++ b/api_internal_test.go
@@ -36,6 +36,15 @@ func TestOpenAIFullURL(t *testing.T) {
 	}
 }
 
+func TestAnthropicFullURL(t *testing.T) {
+	cli := NewClientWithConfig(DefaultAnthropicConfig("dummy", ""))
+	// APIVersion is set for the anthropic-version header; it must not be
+	// appended as an api-version query param the way Azure requires.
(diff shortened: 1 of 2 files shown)


serhiizghama:
The red Sanity check here isn't from this diff — golangci-lint 2.12.2 couldn't typecheck the Go 1.27.0 stdlib that was in the runner toolcache on Aug 26 (`could not import math/rand/v2: method must have no type parameters`). The nightly master runs have been green since Aug 27, so I re-pushed the same commit to pick up a fresh run. Diff is unchanged.

Issue #1070, opened 2025-09-04
Add support for reasoning field from gpt-oss models while maintaining backward compatibility


LubyRuffy:
Hello `go-openai` maintainers,

First and foremost, thank you for your excellent work on this essential library for the Go community.

I am writing to report a compatibility issue when using the library with gpt-oss models, particularly those served through popular platforms like Ollama and LM Studio.

## **Problem Description**

The library's `ChatCompletionStreamChoiceDelta` struct is designed to parse reasoning steps from a stream using the JSON tag `reasoning_content`.

While this works for certain models, the emerging standard for gpt-oss models, as seen in community implementations, is to use the field name `reasoning`. This discrepancy prevents the library from correctly unmarshaling the reasoning data from these models, leaving the `ReasoningContent` field empty.

## **Evidence**

**1. Current Library Implementation (`chat_stream.go`):**
The struct expects the `reasoning_content` field.
```go
type ChatCompletionStreamChoiceDelta struct {
    // ... other fields
	ReasoningContent string `json:"reasoning_content,omitempty"`
}
```

**2. Actual gpt-oss JSON Output (Example from Ollama):**
The model stream provides a `reasoning` field instead.
```json
"delta": {
    "role": "assistant",
    "content": "",
    "reasoning": "This is a reasoning step."
}```
*(Full JSON chunk examples from both Ollama and LM Studio can be provided if needed.)*

## **Importance of Backward Compatibility**

A key consideration is that simply renaming the JSON tag from `reasoning_content` to `reasoning` would likely be a **breaking change** for users who rely on the current implementation for other models (e.g., DeepSeek). Any potential solution should ideally preserve this existing functionality.

## **Request**

To improve compatibility with the growing gpt-oss ecosystem, would it be possible to enhance the library to support the `reasoning` field in addition to the existing `reasoning_content`?

We believe that accommodating both formats would make the library more robust and versatile for all users. We defer to your expertise on the best architectural approach to implement this.

Thank you for your time and for considering this request. I am happy to provide any further information or assist with testing.

dgrieser:
Another attempt:
https://github.com/sashabaranov/go-openai/pull/1119

PR #1119, opened 2026-08-28
Add reasoning alias handling in ChatCompletionMessage unmarshalling

dgrieser:
**Describe the change**

`ChatCompletionMessage.ReasoningContent` and `ChatCompletionStreamChoiceDelta.ReasoningContent` are only populated when the server sends the JSON key `reasoning_content`. A number of OpenAI-compatible backends (OpenRouter, vLLM's OpenAI server, LiteLLM, Ollama and others) emit the same data under the key `reasoning` instead. Against those servers the reasoning text is silently dropped — `ReasoningContent` stays empty and there is no other field the text lands in, so callers lose it entirely unless they decode the raw JSON themselves.

This change makes both types accept `reasoning` as an alias for `reasoning_content` when unmarshalling. Behaviour is unchanged for servers that already send `reasoning_content`, and nothing changes on the request side.

Chat Completions object: https://platform.openai.com/docs/api-reference/chat/object

To be upfront: neither `reasoning_content` nor `reasoning` appears in the official Chat Completions reference. `reasoning_content` is already supported by this library (`chat.go`, `chat_stream.go`) for DeepSeek-style compatible servers; `reasoning` is the other de facto spelling used by the same class of servers. OpenAI's own reasoning output is exposed through the Responses API reasoning item (https://platform.openai.com/docs/api-reference/responses/object), which this PR does not touch.

**Describe your solution**

- `chat.go`: both anonymous decode structs in `ChatCompletionMessage.UnmarshalJSON` (single-content and multi-content variants) gain `Reasoning string \`json:"reasoning,omitempty"\``. After decoding, `ReasoningContent` is filled from `Reasoning` only when `ReasoningContent` is empty, so an explicit `reasoning_content` always wins. Since the temporary struct no longer matches `ChatCompletionMessage` field-for-field, the `ChatCompletionMessage(msg)` conversions are replaced by explicit field assignments.
- `chat_stream.go`: `ChatCompletionStreamChoiceDelta` had no custom unmarshaller, so it gets one with the same shape and the same precedence rule.
- The public API is unchanged: `Reasoning` exists only in the internal decode structs, so there is no new exported field and no change to marshalling — requests and responses still serialize `reasoning_content` only.
- No new dependencies; `encoding/json` in `chat_stream.go` is the only added import.

**Tests**

- `TestChatCompletionMessageUnmarshalReasoningAlias` (`chat_test.go`): unmarshals `{"role":"assistant","content":"final","reasoning":"hidden chain"}` and asserts `ReasoningContent == "hidden chain"` with `Content` untouched.
- `TestChatCompletionStreamChoiceDeltaUnmarshalReasoningAlias` (`chat_stream_test.go`): the same for the stream delta.
- `compareChatStreamResponseChoices` now also compares `Delta.ReasoningContent`. Previously the existing streaming tests would have passed even if reasoning content were dropped.
- `go test ./...` and `go vet ./...` pass locally.
- No integration test against a live endpoint: reproducing this needs a non-OpenAI compatible server. If you would prefer end-to-end coverage, I can add a streaming test that serves a `reasoning`-keyed SSE payload from the existing `internal/test` server.

**Additional context**

Maintenance note: both unmarshallers now assign fields explicitly, so any future field added to `ChatCompletionMessage` or `ChatCompletionStreamChoiceDelta` has to be added to the decode struct and the assignment too. This can be avoided with a local `type alias X` plus embedding so the decode struct stays in sync automatically — happy to rework it that way if you prefer.

Issue: #1070


sashabaranov:
Hey, thank you for the PR! Can we simplify this? Something along the lines of
```go
func (m *ChatCompletionMessage) UnmarshalJSON(bs []byte) error {
	type message ChatCompletionMessage

	msg := struct {
		message
		Reasoning string `json:"reasoning,omitempty"`
	}{}

	if err := json.Unmarshal(bs, &msg); err == nil {
		if msg.ReasoningContent == "" {
			msg.ReasoningContent = msg.Reasoning
		}
		*m = ChatCompletionMessage(msg.message)
		return nil
	}

	multiMsg := struct {
		message
		MultiContent []ChatMessagePart `json:"content"`
		Reasoning    string            `json:"reasoning,omitempty"`
	}{}

	if err := json.Unmarshal(bs, &multiMsg); err != nil {
		return err
	}
	if multiMsg.ReasoningContent == "" {
		multiMsg.ReasoningContent = multiMsg.Reasoning
	}

	multiMsg.message.MultiContent = multiMsg.MultiContent
	*m = ChatCompletionMessage(multiMsg.message)
	return nil
}
```

and 
```go
func (d *ChatCompletionStreamChoiceDelta) UnmarshalJSON(bs []byte) error {
	type delta ChatCompletionStreamChoiceDelta

	aux := struct {
		delta
		Reasoning string `json:"reasoning,omitempty"`
	}{}

	if err := json.Unmarshal(bs, &aux); err != nil {
		return err
	}
	if aux.ReasoningContent == "" {
		aux.ReasoningContent = aux.Reasoning
	}

	*d = ChatCompletionStreamChoiceDelta(aux.delta)
	return nil
}
```

dgrieser:
> Hey, thank you for the PR! Can we simplify this? Something along the lines of
> 
> [...]

Good idea! Thanks.

Applied your suggestion with https://github.com/sashabaranov/go-openai/pull/1119/changes/51044f3878224a01c8d3c8dc8b5c899061b50952

Dave

PR #1120, opened 2026-08-31
[BREAKING_CHANGES] fix: return response headers from streaming endpoints on HTTP error

### LINKED ISSUE #867: Retry-After headers when 'CreateChatCompletionStream' fails due rate limit hits
I believe this could be a duplicate of #452, but the discussion there diverged from the actual point.

# Problem

When using the `CreateChatCompletionStream`, and hitting rate limits errors (Code: 429), the returned error nor the stream doesn't contain the http headers. I need to retrieve 'Retry-After' headers to pause the job for the specified time.

> Those headers are standard and provide the amount of time (seconds/time-stamp on 'Retry-After', milliseconds on 'Retry-After-Ms').

If a 429 is obtained, what I do with `CreateChatCompletion` is to parse those headers and pause to go routine for the specified amount of time.

```go
	resp, err := client.client.CreateChatCompletion(ctx, *openaiReq)
	if err != nil {
		... check status code ...
		parseAndPause(resp.Header()) // even with error, we still get `resp`
	}
```

But for `CreateChatCompletionStream` the above code could not work as:

```go
	stream, err := client.client.CreateChatCompletionStream(ctx, *openaiReq)
	if err != nil {
		... check status code ...
		parseAndPause(stream.Header()) // stream.header could be nil, thus this panics
	}
```

# Proposed solution

I believe here:

https://github.com/sashabaranov/go-openai/blob/7c145ebb4be68610bc3bb5377b754944307d44fd/client.go#L166C1-L172C3

```go
	resp, err := client.config.HTTPClient.Do(req) //nolint:bodyclose // body is closed in stream.Close()
	if err != nil {
		return new(streamReader[T]), err
	}
	if isFailureStatusCode(resp) {
		return new(streamReader[T]), client.handleErrorResp(resp)
	}
```

Those returns should add the response into the `streamReader` struct.

fra-shipper:
Fixes #867

## Root cause

`sendRequestStream` returned a zero-value `streamReader[T]` (no `response`, no headers) whenever the API responded with a failure status code:

```go
if isFailureStatusCode(resp) {
	return new(streamReader[T]), client.handleErrorResp(resp)
}
```

`CreateChatCompletionStream`, `CreateCompletionStream`, and `CreateResponseStream` then discarded that value entirely on error (early `return` / `return nil, err`), so callers got a nil stream with no way to read `Retry-After` or the `x-ratelimit-*` headers to implement backoff. This is the same class of bug that #751 fixed for the non-streaming path (`sendRequest` already calls `v.SetHeader(res.Header)` before checking `isFailureStatusCode`); it was just never applied to the streaming path.

## Fix

**[BREAKING_CHANGES]**: on an HTTP-status error, `CreateChatCompletionStream` / `CreateCompletionStream` / `CreateResponseStream` now return a non-nil stream instead of `nil`, so its headers stay reachable.

- `sendRequestStream` now distinguishes two failure modes:
  - Transport failure (`HTTPClient.Do` itself errors — connection refused, DNS failure, canceled context): still returns `nil, err`. There is no response to attach, so the caller keeps getting a nil stream exactly as before.
  - HTTP-status failure (a response came back, e.g. 429): now returns a `streamReader[T]` with `response` and `httpHeader` populated from the failed response, instead of a bare zero value. Attaching `response` (not just the header) also lets `stream.Close()` release the connection instead of leaking it, since `handleErrorResp` drains the body via `io.ReadAll` but never closes it.
- `CreateChatCompletionStream`, `CreateCompletionStream`, and `CreateResponseStream` now hand back that stream alongside the error (only wrapping it when non-nil) instead of dropping it, so `stream.Header()` / `stream.GetRateLimitHeaders()` work on the HTTP-status-error path.
- Because this HTTP-status-error stream is never wired to a real response body, `streamReader.RecvRaw()` now returns `io.EOF` when its internal `reader` is nil, instead of the code that used to be unreachable on that path (a nil-pointer panic inside `processLines`). Its only intended use on that path is reading headers, then `Close()`.
- Added doc comments to the three exported constructors stating the new contract: non-nil stream (headers reachable, `Recv()` returns `io.EOF`) on an HTTP-status error, nil stream on a transport failure.

## Testing

- `TestCreateChatCompletionStreamRateLimitErrorHeaders`, `TestCreateCompletionStreamRateLimitErrorHeaders` (new): mock a 429 with `Retry-After` and rate-limit headers, assert the returned stream is non-nil, its headers are reachable, and `Recv()` returns `io.EOF` rather than panicking.
- `TestCreateResponseStreamHTTPError` (updated): previously asserted `stream == nil` on HTTP error — the exact behavior being fixed here. Now asserts the stream is non-nil, `Retry-After` is readable, and `Recv()` returns `io.EOF`.
- `TestCreateChatCompletionStreamTransportError`, `TestCreateCompletionStreamTransportError`, `TestCreateResponseStreamTransportError` (new): regression coverage for the transport-failure sub-case, via a new `setupUnreachableClient()` helper that points the client at a closed local port so `HTTPClient.Do` fails with "connection refused" (no network dependency, no cost). Asserts the stream is still nil in that case.
- Verified RED against the pre-fix source with the new tests in place: the three headers tests failed with a nil-pointer panic inside `processLines` (e.g. `chat_stream_test.go:397: ... nil pointer dereference`), and the three transport-error tests failed their explicit "expected a nil stream" assertions.
- `go build ./...`, `go vet ./...`, and `gofmt -l` on all 9 touched files are clean.
- `go test ./...` and `go test ./... -race`: all packages `ok`, no failures, no races.

Not run: `api_integration_test.go` (guarded by `//go:build integration`, requires a live `OPENAI_API_KEY` and real network calls) and `goimports`/`golangci-lint` (not installed locally; installing them would require a network fetch outside the scope of this change). `gofmt` + `go vet` were used as the closest available local equivalents.


sashabaranov:
@fra-shipper thank you for the PR! could you please update it to the latest `master` branch, I believe there are couple of lint issues here

